Issue #123, opened 2021-11-16
Don't log secrets


dylan-smith:
The verbose logs generated today probably contain the PAT's in some cases. We should deliberately mask out any secret values before writing logs to disk/console.

bryantson:
👍

PR #191, opened 2022-01-12
don't log secrets

dylan-smith:
#123 

The EnvironmentVariableProviders now register their secrets with OctoLogger, and OctoLogger will mask them out of logs and/or console output.

I also made some changes to OctoLogger to make it more testable, and added a unit test for the secrets behavior.

- [x] Issue linked


src/Octoshift/OctoLogger.cs
```diff
@@ -15,29 +16,55 @@ internal static class LogLevel
     public class OctoLogger
     {
         public bool Verbose { get; set; }
-        private readonly DateTime _logStartTime;
+        private readonly List<string> _secrets = new List<string>();
```
ArinGhazarian, line 19: I would use a `HashSet<string>` in order to not to add duplicate data.

ArinGhazarian (commented):
Overall looks good, just a minor comment to use a `HashSet` instead of a `List`.

src/Octoshift/OctoLogger.cs
dylan-smith: done

Issue #114, opened 2021-11-16
Unit Tests


dylan-smith:
## TODO

- [x] #159
- [x] #163
- [x] #164
- [x] #177
- [x] #173

PR #196, opened 2022-01-13
AdoApi improvements

dylan-smith:
Updating AdoApi to use the anonymous object approach instead of giant JSON strings (like was done in GithubApi already). This will make the code cleaner, and also the unit tests will be more robust because they aren't tied to very specific string/format implementation.

#114 

- [x] Issue linked

### DIFF
--- a/src/Octoshift/AdoApi.cs
+++ b/src/Octoshift/AdoApi.cs
@@ -4,6 +4,7 @@ using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
 using Newtonsoft.Json.Linq;
+using OctoshiftCLI.Extensions;
 
 namespace OctoshiftCLI
 {
@@ -28,7 +29,6 @@ namespace OctoshiftCLI
                 return userId;
             }
 
-            // TODO: Throw an exception instead
             Console.WriteLine("Unexpected response when retrieving User ID");
             Console.WriteLine(response);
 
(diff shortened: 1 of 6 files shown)


src/Octoshift/AdoApi.cs
```diff
@@ -284,139 +286,127 @@ public virtual async Task ShareServiceConnection(string adoOrg, string adoTeamPr
...
+                        name = adoTeamProject
+                    }
+                }
+            };
 
             await _client.PatchAsync(url, payload);
         }
 
-        public virtual async Task<AdoPipeline> GetPipeline(string org, string teamProject, int pipelineId)
+        public virtual async Task<(string defaultBranch, string clean, string checkoutSubmodules)> GetPipeline(string org, string teamProject, int pipelineId)
```
ArinGhazarian, line 305: Since behind the scenes a value tuple is actually a `struct` all properties should be pascal cased. Specially when we expose a value tuple. The only exception is when we use the properties in a method where they are actually local variables.

```suggestion
        public virtual async Task<(string DefaultBranch, string Clean, string CheckoutSubmodules)> GetPipeline(string org, string teamProject, int pipelineId)
```

src/Octoshift/AdoApi.cs
```diff
@@ -284,139 +286,127 @@ public virtual async Task ShareServiceConnection(string adoOrg, string adoTeamPr
...
+            clean = clean == null ? "null" : clean.ToLower();
+
+            var checkoutSubmodules = (string)data["repository"]["checkoutSubmodules"];
+            checkoutSubmodules = checkoutSubmodules == null ? "null" : checkoutSubmodules.ToLower();
+
+            return (defaultBranch, clean, checkoutSubmodules);
         }
 
-        public virtual async Task ChangePipelineRepo(AdoPipeline pipeline, string githubOrg, string githubRepo, string serviceConnectionId)
+        //public virtual async Task ChangePipelineRepo(AdoPipeline pipeline, string githubOrg, string githubRepo, string connectedServiceId)
```
ArinGhazarian, line 328: Was this intended?

src/Octoshift/AdoApi.cs
```diff
@@ -498,41 +489,39 @@ public virtual async Task LockRepo(string org, string teamProjectId, string repo
...
-    }
+    //public class AdoPipeline
+    //{
+    //    public int Id { get; set; }
+    //    public string Org { get; set; }
+    //    public string TeamProject { get; set; }
+    //    public string DefaultBranch { get; set; }
+    //    public string Clean { get; set; }
+    //    public string CheckoutSubmodules { get; set; }
+    //}
```
ArinGhazarian, line 526: This probably wasn't intended also.

src/ado2gh/Commands/RewirePipelineCommand.cs
```diff
@@ -72,8 +72,8 @@ public async Task Invoke(string adoOrg, string adoTeamProject, string adoPipelin
             var ado = _lazyAdoApi.Value;
 
             var adoPipelineId = await ado.GetPipelineId(adoOrg, adoTeamProject, adoPipeline);
-            var pipelineDetails = await ado.GetPipeline(adoOrg, adoTeamProject, adoPipelineId);
-            await ado.ChangePipelineRepo(pipelineDetails, githubOrg, githubRepo, serviceConnectionId);
+            var (defaultBranch, clean, checkoutSubmodules) = await ado.GetPipeline(adoOrg, adoTeamProject, adoPipelineId);
```
ArinGhazarian, line 75: here it's ok to use camel cased properties in a value tuple because as I mentioned there are going to be uses as local variables.

PR #200, opened 2022-01-14
Add support for source and target github pats

### LINKED ISSUE #192: Multi PAT support for GH to GH migrations
For GH to GH migrations we need to support a source PAT in addition to the target PAT. By default if it is not provided, it should fall back to the target PAT.

## Todo
- [ ] Leverage `GH_SOURCE_PAT` for the source repo, if not provided fall back to the existing `GH_PAT`.

ArinGhazarian:
Closes #192 

This PR added the support for having two separate Github PATs for GHEC migrations for source and target. 

- [x] Issue linked

### GEI migration
A single repo migration succeeds using different PATs for source and target. This test did a migration within one org.
image

### Updated descriptions for `ado2gh` that include required env variables
image

### Updated descriptions for `gei` that include required env variables
image

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -4,4 +4,3 @@
 - Released gei.exe that adds support for Github -> Github migrations (GHEC only for now). In the future this will be exposed as an extension to the Github CLI.
 - Automatically remove secrets from log files and console output (previously the verbose logs would contain your PAT's)
 - Added --ssh option to generate-script and migrate-repo commands (in both ado2gh and gei). This forces the migration to use an older version of the API's that uses SSH to push the repos into GitHub. The newer API's use HTTPS instead. However some customers have been running into problems with some repos that work fine using the older SSH API's. In the future this option will be deprecated once the issues with the HTTPS-based API's are resolved.
-- gei.exe can now optionally use GH_SOURCE_PAT for source.
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -4,3 +4,4 @@
 - Released gei.exe that adds support for Github -> Github migrations (GHEC only for now). In the future this will be exposed as an extension to the Github CLI.
 - Automatically remove secrets from log files and console output (previously the verbose logs would contain your PAT's)
 - Added --ssh option to generate-script and migrate-repo commands (in both ado2gh and gei). This forces the migration to use an older version of the API's that uses SSH to push the repos into GitHub. The newer API's use HTTPS instead. However some customers have been running into problems with some repos that work fine using the older SSH API's. In the future this option will be deprecated once the issues with the HTTPS-based API's are resolved.
+- gei.exe can now optionally use GH_SOURCE_PAT for source.
--- a/src/Octoshift/GithubClient.cs
+++ b/src/Octoshift/GithubClient.cs
@@ -16,10 +16,13 @@ namespace OctoshiftCLI
             _log = log;
             _httpClient = httpClient;
 
-            _httpClient.DefaultRequestHeaders.Add("Accept", "application/vnd.github.v3+json");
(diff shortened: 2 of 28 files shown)


src/Octoshift/GithubClient.cs
```diff
@@ -10,10 +11,15 @@ public class GithubClient
         private readonly HttpClient _httpClient;
         private readonly OctoLogger _log;
 
-        public GithubClient(OctoLogger log, HttpClient httpClient)
+        public GithubClient(OctoLogger log, HttpClient httpClient, string personalAccessToken)
         {
             _log = log;
             _httpClient = httpClient;
+
+            _httpClient.DefaultRequestHeaders.Add("Accept", "application/vnd.github.v3+json");
```
dylan-smith, line 19: You might consider wrapping this in an if _httpClient != null block that way you can pass null into the mocks and it won't break

src/OctoshiftCLI.Tests/GithubApiTests.cs
```diff
@@ -107,7 +107,7 @@ public async Task RemoveTeamMember_Calls_The_Right_Endpoint()
 
             var url = $"https://api.github.com/orgs/{org}/teams/{teamName}/memberships/{member}";
 
-            var githubClinetMock = new Mock<GithubClient>(null, null);
+            var githubClinetMock = new Mock<GithubClient>(null, new HttpClient(), "pat");
```
dylan-smith:
```suggestion
            var githubClientMock = new Mock<GithubClient>(null, new HttpClient(), "pat");
```

RELEASENOTES.md
```diff
@@ -3,4 +3,5 @@
...
 - Automatically remove secrets from log files and console output (previously the verbose logs would contain your PAT's)
-- Added --ssh option to generate-script and migrate-repo commands (in both ado2gh and gei). This forces the migration to use an older version of the API's that uses SSH to push the repos into GitHub. The newer API's use HTTPS instead. However some customers have been running into problems with some repos that work fine using the older SSH API's. In the future this option will be deprecated once the issues with the HTTPS-based API's are resolved.
\ No newline at end of file
+- Added --ssh option to generate-script and migrate-repo commands (in both ado2gh and gei). This forces the migration to use an older version of the API's that uses SSH to push the repos into GitHub. The newer API's use HTTPS instead. However some customers have been running into problems with some repos that work fine using the older SSH API's. In the future this option will be deprecated once the issues with the HTTPS-based API's are resolved.
+- gei.exe can now optionally use GH_SOURCE_PAT for source.
```
dylan-smith, line 7: I don't think we need this bullet in the release notes because this is the first release with gei so addin the source pat is not a "change" it's just the way it is the first time the customers see it.

I WOULD however maybe add a blurb to the Description field in the two gei commands (and possibly all other commands) noting which env vars that specific command expects to be set.

ArinGhazarian: addressed in d9a01643f5ca0392832eb66d288995af18a3bfff

src/OctoshiftCLI.Tests/gei.Commands/EnvironmentVariableProviderTests.cs
```diff
@@ -0,0 +1,87 @@
...
+        public EnvironmentVariableProviderTests()
+        {
+            _environmentVariableProvider = new EnvironmentVariableProvider(new Mock<OctoLogger>().Object);
+        }
+
+        [Fact]
+        public void SourceGithubPersonalAccessToken_Should_Return_Github_Source_Pat()
+        {
+            // Arrange
+            Environment.SetEnvironmentVariable("GH_SOURCE_PAT", SOURCE_GH_PAT);
```
ArinGhazarian, line 25: NOTE: These tests are not going to run in parallel because by default they are in the same xUnit collection and this is going to be the only place where we set the env variables. So it should be ok to set them. For the rest of the tests we always mock the `EnvironmentVariableProvider`

dylan-smith (approved):

PR #206, opened 2022-01-19
e2e testing

dylan-smith:
Implements #3 

Creates 4x e2e test cases
Updated CI to run them on all 3 OS's in parallel
Setup testing environments that the tests will use
Each test resets the environment state, creates some testing data, then runs generate-script + runs the script
At the end each test inspects the test environment to assert that the final state is correct

- [x] Issue linked


src/Octoshift/AdoApi.cs
```diff
@@ -262,8 +264,25 @@ public virtual async Task<string> GetTeamProjectId(string org, string teamProjec
...
+                    // The repo may be disabled, can still get the ID by getting it from the repo list
+                    url = $"https://dev.azure.com/{org}/{teamProject}/_apis/git/repositories?api-version=4.1";
+
+                    var response = await _client.GetWithPagingAsync(url);
+
+                    return (string)response.Single(x => ((string)x["name"]) == repo)["id"];
+                }
+
+                throw;
+            }
```
ArinGhazarian, line 285: if we're only going to check for `NotFound` to remove indentation we can use `when`:
```suggestion
            catch (HttpRequestException ex) when (ex.StatusCode == HttpStatusCode.NotFound)
            {
                // The repo may be disabled, can still get the ID by getting it from the repo list
                url = $"https://dev.azure.com/{org}/{teamProject}/_apis/git/repositories?api-version=4.1";

                var response = await _client.GetWithPagingAsync(url);

                return (string)response.Single(x => ((string)x["name"]) == repo)["id"];
            }
```

src/OctoshiftCLI.IntegrationTests/TestHelper.cs
```diff
@@ -0,0 +1,564 @@
...
+
+            _output.WriteLine($"Found {teamProjects.Count()} Team Projects");
+
+            foreach (var teamProject in teamProjects.Where(x => x != "service-connection-project-do-not-delete"))
+            {
+                _output.WriteLine($"Deleting Team Project: {adoOrg}\\{teamProject}...");
+                var teamProjectId = await _adoApi.GetTeamProjectId(adoOrg, teamProject);
+                var operationId = await DeleteTeamProject(adoOrg, teamProjectId);
+
+                while (await GetOperationStatus(adoOrg, operationId) is "notSet" or "queued" or "inProgress")
```
ArinGhazarian, line 50: nit: Def. not a biggie but I'd create constants for these strings (`NOT_SET`, `QUEUED`, `IN_PROGRESS`) since they are being used in other methods too.

src/OctoshiftCLI.IntegrationTests/TestHelper.cs
```diff
@@ -0,0 +1,564 @@
...
+                    else
+                    {
+                        startInfo.EnvironmentVariables.Add("ADO_PAT", token.Value);
+                    }
+                }
+            }
+
+            _output.WriteLine($"Running command: {startInfo.FileName} {startInfo.Arguments}");
+            var p = Process.Start(startInfo);
+            p.WaitForExit();
```
ArinGhazarian, line 420: We can asynchronously wait for exit: [p.WaitForExitAsync()](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.waitforexitasync?view=net-6.0#system-diagnostics-process-waitforexitasync(system-threading-cancellationtoken)). That way the method will be `async` and non-blocking.

ArinGhazarian (commented):
Looks great, just a few minor comments.

src/Octoshift/AdoApi.cs
dylan-smith: first time I've seen that syntax, I like!

src/OctoshiftCLI.IntegrationTests/TestHelper.cs
dylan-smith: done

src/OctoshiftCLI.IntegrationTests/TestHelper.cs
dylan-smith: done

PR #219, opened 2022-01-30
add path to generate migration archive

oakeyc:
resolves https://github.com/github/octoshift/issues/3588

Adds migration path for generating github archive.

We add a new command for gei `generate-archive` which using the migration on github to create 2 migration archives, 1 for metadata and 1 for git data. We use the IHttpClientFactory dependency injection pattern for determining which http to use. For GHES (github enterprise server) we need an http client which can ignore client ssl errors, which can be specified with `--no-ssl-verify`

- [x] Issue linked


src/Octoshift/GithubClient.cs
```diff
@@ -58,7 +61,7 @@ public virtual async IAsyncEnumerable<JToken> GetAllAsync(string url)
         public virtual async Task<string> DeleteAsync(string url) => (await SendAsync(HttpMethod.Delete, url)).Content;
 
         private async Task<(string Content, KeyValuePair<string, IEnumerable<string>>[] ResponseHeaders)> SendAsync(
-            HttpMethod httpMethod, string url, object body = null)
+            HttpMethod httpMethod, string url, object body = null, bool shouldEnsureSuccessStatusCode = true)
```
ArinGhazarian, line 64: why was this flag added? We def. want to raise exceptions in case on non success responses. In case we want to treat them differently (i.e. 4xx vs 5xx) then we can raise custom exceptions based on the status code.

src/Octoshift/GithubApi.cs
```diff
@@ -304,5 +304,34 @@ public virtual async Task<bool> RevokeMigratorRole(string org, string actor, str
...
+            var url = $"{apiUrl}/orgs/{org}/migrations";
+
+            var payload = new { repositories };
+
+            var response = await _client.PostAsync(url, payload);
+            var data = JObject.Parse(response);
+            return (int)data["id"];
+        }
+
+        public virtual async Task<string> MigrationStatus(string apiUrl, string org, int migrationId)
```
ArinGhazarian, line 319: Two things here:
1.  By convention method names in `C#` usually have the verb in them to show that they are performing some action. So here a better name would start with `Get` i.e. `GetMigrationStatus`
2. We already have a `GetMigrationState` method which may easily be confused with this one. I am guessing this is returning the migration status for the archive generation? if so then it's better to name it accordingly, i.e. `GetArchiveGenerationStatus`.

src/OctoshiftCLI.Tests/Integration/GenerateArchiveTests.cs
```diff
@@ -0,0 +1,28 @@
...
+
+            await OctoshiftCLI.GithubEnterpriseImporter.Program.Main(parameters);
+            return "";
+        }
+        #endregion
+
+        [Fact]
+        public async void Should_Fail_To_Generate_With_Invalid_Parameters()
+        {
+            var response = await GenerateArchive("invalid-parameters", "--unsupport-parameter");
```
ArinGhazarian, line 25: I don't think there should be an INT test for this command, it can easily be converted to a unit test. We tend to write scenario based INT tests but this is just testing the `generate-archive` command.

src/OctoshiftCLI.Tests/Integration/GenerateArchiveTests.cs
```diff
@@ -0,0 +1,28 @@
+using System.IO;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace OctoshiftCLI.Tests.Integration
+{
+    public class GenerateArchiveTests
+    {
+        #region Script generation and verification helpers
```
ArinGhazarian, line 9: Using `#region` is discouraged nowadays because ppl usually hide their smelly code in region blocks instead of paying attention to the code smell and try to address it 😄

src/gei/GithubApiFactory.cs
```diff
@@ -23,6 +22,16 @@ GithubApi ISourceGithubApiFactory.Create()
...
+        GithubApi ISourceGithubApiFactory.CreateDefaultClient()
+        {
+            var githubPat = _environmentVariableProvider.SourceGithubPersonalAccessToken();
+
+            #pragma warning disable CA2000 // We don't want to dispose the handler until the client is disposed
+            var client = new HttpClientFactory().CreateClient("OctoShift");
+            var githubClient = new GithubClient(_octoLogger, client, githubPat);
+            return new GithubApi(githubClient);
+        }
+
```
ArinGhazarian, line 34: What is the purpose of this factory method? it seems like it is exactly creating an Api corresponding the the source PAT. 
Also we use DI, so there is no need to have `var client = new HttpClientFactory().CreateClient("OctoShift");` b/c it defies the reason why we use DI in the first place!

src/gei/HttpClientFactory.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        // todo add the name
+        public HttpClient CreateClient(string name)
+        {
+            #pragma warning disable CA2000 // We don't want to dispose the handler until the client is disposed
+            var handler = new HttpClientHandler
+            {
+                AllowAutoRedirect = false,
+                CheckCertificateRevocationList = true
+            };
```
ArinGhazarian, line 15: We shouldn't directly create MessageHandlers, that's why we use `HttpClientFactory` in the first place.

ArinGhazarian (changes requested):

src/Octoshift/GithubClient.cs
```diff
@@ -29,6 +30,8 @@ public GithubClient(OctoLogger log, HttpClient httpClient, string personalAccess
             }
         }
 
+        public virtual async Task<string> GetHeadersAsync(string url) => (await SendAsync(HttpMethod.Get, url, shouldEnsureSuccessStatusCode: false)).Content;
```
ArinGhazarian, line 33: This method is very confusing. It suggests that it is just going to get the headers of a response but it's not! it's just a `GET` that ignores the success status check. I can see that in the API the use case is just to get (more likely to check) the archive's url, based on [this](https://docs.github.com/en/rest/reference/migrations#download-an-organization-migration-archive).
So just to check if the url exists or not you can just make a request with the existing `GetAsync` and if case no errors just return `true`.

src/OctoshiftCLI.Tests/GithubClientTests.cs
```diff
@@ -542,6 +542,23 @@ public async Task DeleteAsync_Logs_The_Url()
...
+
+            var expectedLogMessage = $"RESPONSE ({HttpStatusCode.OK}): {EXPECTED_RESPONSE_CONTENT}";
+
+            // Act
+            await githubClient.GetHeadersAsync("http://example.com");
+
+            // Assert
+            _loggerMock.Verify(m =>
+                m.LogVerbose(It.Is<string>(actualLogMessage => actualLogMessage == expectedLogMessage)));
+        }
```
ArinGhazarian, line 560: This test is actually not testing if the request is successful it is just verifying that we do log the response!

src/gei/Commands/GenerateArchiveCommand.cs
```diff
@@ -0,0 +1,96 @@
...
+
+                if (archiveStatus == "exported")
+                {
+                    isFinished = true;
+                }
+                else if (archiveStatus == "failed")
+                {
+                    _log.LogError($"Archive generation failed with id: {migrationId}");
+                    return;
+                }
```
ArinGhazarian, line 86: We can return a strongly typed `enum` instead of strings from `MigrationStatus`

src/Octoshift/GithubApi.cs
```diff
@@ -304,5 +304,34 @@ public virtual async Task<bool> RevokeMigratorRole(string org, string actor, str
...
+        {
+            var url = $"{apiUrl}/orgs/{org}/migrations/{migrationId}";
+
+            var response = await _client.GetAsync(url);
+            var data = JObject.Parse(response);
+
+            return (string)data["state"];
+        }
+
+        public virtual async Task<string> MigrationArchiveURL(string apiUrl, string org, int migrationId)
```
ArinGhazarian, line 329: `GetMigrationArchiveUrl`

src/Octoshift/GithubClient.cs
oakeyc: We talked about it, and the api returns 302 on purpose: https://docs.github.com/en/rest/reference/migrations#download-an-organization-migration-archive

src/gei/Commands/GenerateArchiveCommand.cs
```diff
@@ -0,0 +1,96 @@
...
+        public async Task Invoke(string githubURL, string githubSourceOrg, string githubSourceRepo, bool verbose = false)
+        {
+            _log.Verbose = verbose;
+
+            _log.LogInformation("Generating Migration Archive...");
+            _log.LogInformation($"GITHUB SOURCE ORG: {githubSourceOrg}");
+
+            if (string.IsNullOrWhiteSpace(githubURL))
+            {
+                _log.LogInformation("GithubURL not provided, defaulting to https://api.github.com");
```
petermeglis, line 58:
```suggestion
                _log.LogInformation("GitHubURL not provided, defaulting to https://api.github.com");
```

ArinGhazarian: We did you `Github` everywhere so if we change this we need to change it everywhere. It looks like a tech-dept to me.

src/Octoshift/GithubApi.cs
oakeyc: Updated the name, and we bass in the apiUrl because for ghes and ghae we'll need it since it's not the default 👍

src/Octoshift/GithubClient.cs
oakeyc: Updated the naming

src/gei/Commands/GenerateArchiveCommand.cs
oakeyc: Updated!

src/gei/Commands/GenerateArchiveCommand.cs
```diff
@@ -0,0 +1,98 @@
...
+        {
+            _log = log;
+            _sourceGithubApiFactory = sourceGithubApiFactory;
+            _environmentVariableProvider = environmentVariableProvider;
+
+            Description = "Invokes the GitHub Migration API's to generate a migration archive";
+            Description += Environment.NewLine;
+            Description += "Note: Expects GH_PAT and GH_SOURCE_PAT env variables to be set. GH_SOURCE_PAT is optional, if not set GH_PAT will be used instead. This authenticates to source GHES API";
+
+            var githubURL = new Option<string>("--github-url")
```
petermeglis, line 26: I think @roferg mentioned changing this to `--ghes-url`, does that still stand?

src/gei/Commands/GenerateArchiveCommand.cs
```diff
@@ -0,0 +1,98 @@
...
+
+namespace OctoshiftCLI.GithubEnterpriseImporter.Commands
+{
+    public class GenerateArchiveCommand : Command
+    {
+        private readonly OctoLogger _log;
+        private readonly ISourceGithubApiFactory _sourceGithubApiFactory;
+        private readonly EnvironmentVariableProvider _environmentVariableProvider;
+
+        public GenerateArchiveCommand(OctoLogger log, ISourceGithubApiFactory sourceGithubApiFactory, EnvironmentVariableProvider environmentVariableProvider) : base("generate-archive")
```
petermeglis, line 16: Are we planning on having the archive generation be a separate step? Or have this whole GHES migration (archive generation, dowload/upload to azure blob, then kick off octoshift) be in one command? If the latter, should we update the name to something like `MigrateGHESRepoCommand`?

Also, if the latter, we'll have lots of overlap with the `MigrateRepoCommand`, maybe we can consolidate these in a followup PR

petermeglis (commented):
Not sure if you're in the "taking feedback" stage or still just in the "get it functionally working" stage, but dropped some thoughts regardless

src/gei/Commands/GenerateArchiveCommand.cs
```diff
@@ -0,0 +1,99 @@
...
+            if (string.IsNullOrWhiteSpace(githubURL))
+            {
+                _log.LogInformation("GitHub URL not provided, defaulting to https://api.github.com");
+                githubURL = "https://api.github.com";
+            }
+
+            var repositories = new string[] { githubSourceRepo };
+
+            var githubApi = _sourceGithubApiFactory.CreateClientNoSSL();
+            var migrationId = await githubApi.StartArchiveGeneration(githubURL, githubSourceOrg, repositories);
```
petermeglis, line 67: And one more callout that we talked about, we'll want to generate an archive for both the git archive and metadata archives

src/gei/Commands/GenerateArchiveCommand.cs
roferg: Yes, we should make this `--ghes-url`

src/OctoshiftCLI.Tests/Integration/GenerateArchiveTests.cs
dylan-smith: Yeah, I'm about to delete all the existing integration test stuff in another PR and commit a complete rewrite of the integration tests. For this PR unit tests should suffice.

src/gei/Commands/GenerateArchiveCommand.cs
dylan-smith: yeah I realized that about a month into coding this, that I capitilized github wrong, I've just been sticking with "Github" across the code base to keep it consistent (possibly using "GitHub" for user displayed strings).

src/gei/Commands/GenerateArchiveCommand.cs
oakeyc: I'm feeling MigrateGHESRepoCommand :+1:

oakeyc: We talked online, and decided to address in follow up

src/gei/GithubApiFactory.cs
oakeyc: We have this factory so we can manipulate the client for specific kinds of http client connections that we might want. For example, for ghes, we will want to not follow redirects and allow non-ssl signed client connections

src/gei/HttpClientFactory.cs
oakeyc: I think we came to the conclusion to have it work with different kinds of connections, we will keep the factory for now.

src/gei/Commands/GenerateArchiveCommand.cs
oakeyc: Good call out, done

oakeyc:
Screen Shot 2022-02-02 at 5 05 31 PM

src/gei/Commands/GenerateArchiveCommand.cs
```diff
@@ -0,0 +1,150 @@
...
+
+                await Task.Delay(_delayInMS);
+            }
+
+            foreach (var id in ids)
+            {
+                var urlLocation = await githubApi.GetArchiveMigrationUrl(ghesUrl, githubSourceOrg, id);
+                _log.LogInformation($"Archive dowload url: {urlLocation}");
+            }
+        }
```
dylan-smith, line 148: I took a crack at refactoring this a bit:

```suggestion
            var gitArchiveUrl = await WaitForArchiveGeneration(gitDataArchiveId);
            _log.LogInformation($"Archive(git) download url: {gitArchiveUrl}");

            var metadataArchiveUrl = await WaitForArchiveGeneration(metadataArchiveId);
            _log.LogInformation($"Archive(metadata) download url: {metadataArchiveUrl}");
        }

        public async Task<string> WaitForArchiveGeneration(GithubApi githubApi, string ghesUrl, string githubSourceOrg, int archiveId)
        {
            var timeOut = DateTime.Now.AddHours(_timeoutInHours);

            while (DateTime.Now < timeOut)
            {
                var archiveStatus = await githubApi.GetArchiveMigrationStatus(ghesUrl, githubSourceOrg, archiveId);
                _log.LogInformation($"Waiting for archive with id {archiveId} generation to finish. Current status: {archiveStatus}");

                if (archiveStatus == ArchiveMigrationStatus.Exported)
                {
                    return await githubApi.GetArchiveMigrationUrl(ghesUrl, githubSourceOrg, archiveId);
                }

                if (archiveStatus == ArchiveMigrationStatus.Failed)
                {
                    throw new OctoshiftCliException($"Archive generation failed with id: {archiveId}");
                }

                await Task.Delay(_delayInMS);
            }

            throw new TimeoutException($"Archive generation timed out after {_timeoutInHours} hours");
        }
```

src/Octoshift/GithubApi.cs
```diff
@@ -304,5 +305,34 @@ public virtual async Task<bool> RevokeMigratorRole(string org, string actor, str
                 return false;
             }
         }
+
+        public virtual async Task<int> StartArchiveGeneration(string apiUrl, string org, object options)
```
dylan-smith, line 309: I would split this into 2 separate functions so we can abstract away the api payload syntax from our command classes. E.g.

```
public virtual async Task<int> StartMetadataArchiveGeneration(string apiUrl, string org, string repo)
{
    var url = $"{apiUrl}/orgs/{org}/migrations";
    var payload = new {
        repositories = new [] { repo },
        exclude_git_data = true,
        exclude_releases = true,
        exclude_owner_projects = true
    };

    var response = await _client.PostAsync(url,  payload);
    ...
}

**Review by dylan-smith — DISMISSED** [review]

**Comment by dylan-smith** [comment]
I left a whole whack of comments/suggestions, we can pair on them tomorrow if you want.

File: src/gei/Commands/GenerateArchiveCommand.cs [hunk 17]
```diff
@@ -0,0 +1,150 @@
...
+                        archiveIdsWithFinished[id] = true;
+                        currFinished++;
+                    }
+                    else if (archiveStatus == GithubEnums.ArchiveMigrationStatus.Failed)
+                    {
+                        _log.LogError($"Archive generation failed with id: {id}");
+                        return;
+                    }
+
+                }
```
ArinGhazarian, line 138: This `foreach` logic can easily be converted to a method. It seems like all you care is whether the status is `Exported` or `Failed`. So the method can just return the enum. Then in your while loop instead of the `foreach` you can just call the method twice (once per id) and decide whether to continue or not instead of incrementing `currFinished` variable. Something like this:

```c#
var gitDataArchiveMigrationStatus = ArchiveMigrationStatus.Pending;
var metadataArchiveMigrationStatus = ArchiveMigrationStatus.Pending;
while (gitDataArchiveMigrationStatus != ArchiveMigrationStatus.Exported &&
       metadataArchiveMigrationStatus != ArchiveMigrationStatus.Exported &&  
       DateTime.Now < timeout)
{
  gitDataArchiveMigrationStatus = gitDataArchiveMigrationStatus || QueryArchiveMigrationStatus(gitDataArchiveId);
  if (gitDataArchiveMigrationStatus == ArchiveMigrationStatus.Failed)
  {
    return;
  }

  metadataArchiveMigrationStatus = metadataArchiveMigrationStatus || QueryArchiveMigrationStatus(metadataArchiveId);
  if (metadataArchiveMigrationStatus == ArchiveMigrationStatus.Failed)
  {
    return;
  }
}
```

IMO this is way more readable than using dictionary, etc. `C#` supports local functions too (functions inside functions), so you can also create one for checking the overall status, i.e. `ShouldContinue(...)` and use it in the while loop's predicate 😃

ArinGhazarian (dismissed):

src/Octoshift/GithubApi.cs
```diff
@@ -304,5 +305,34 @@ public virtual async Task<bool> RevokeMigratorRole(string org, string actor, str
...
+        {
+            var url = $"{apiUrl}/orgs/{org}/migrations";
+
+            var response = await _client.PostAsync(url, options);
```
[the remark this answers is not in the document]
ArinGhazarian: I like this better than creating a strongly typed object, b/c other that the repo the rest is just bunch of flags which can be set right in the Api method 😃

src/Octoshift/GithubClient.cs
```diff
@@ -82,7 +88,14 @@ public virtual async IAsyncEnumerable<JToken> GetAllAsync(string url)
             var content = await response.Content.ReadAsStringAsync();
             _log.LogVerbose($"RESPONSE ({response.StatusCode}): {content}");
 
-            response.EnsureSuccessStatusCode();
+            if (shouldEnsureSuccessStatusCode)
```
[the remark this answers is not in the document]
ArinGhazarian: @dylan-smith I added a comment about refactoring the `GithubApi`, but even in that case we are going to need this flag b/c success is a range rather than just OK, i.e. 2XX

src/Octoshift/GithubApi.cs
```diff
@@ -304,5 +305,34 @@ public virtual async Task<bool> RevokeMigratorRole(string org, string actor, str
...
+            var data = JObject.Parse(response);
+            return (int)data["id"];
+        }
+
+        public virtual async Task<GithubEnums.ArchiveMigrationStatus> GetArchiveMigrationStatus(string apiUrl, string org, int migrationId)
```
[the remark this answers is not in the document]
ArinGhazarian: I also added a comment about the `enum`. But I am OK with whatever that is easier to implement and use.

src/Octoshift/GithubApi.cs
ArinGhazarian: I really like the suggestion 👍

ArinGhazarian:
I also left a bunch of comments but happy to pair with you 😃

docs/GettingStarted.md
```diff
@@ -0,0 +1,41 @@
...
+
+Run tests with
+```bash
+dotnet test src/OctoshiftCLI.Tests/OctoshiftCLI.Tests.csproj
+```
+
+Format your files locally with:
+```bash
+dotnet format src/OctoshiftCLI.sln
+```
```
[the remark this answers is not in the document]
slenguyen: Two cents: I think it is potentially valuable for engineers who aren't experienced with .NET (which is a lot of the Octoshift team).

src/Octoshift/GithubClient.cs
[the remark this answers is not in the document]
dylan-smith: either way suggested I think works fine, in my suggestion I think it does still work, because by comparing it against OK (which is the default assigned to that arg) you are basically checking if the caller did NOT pass in an explicit status value, and if not calling EnsureSuccess().

src/gei/Commands/GenerateArchiveCommand.cs
```diff
@@ -0,0 +1,150 @@
...
+
+        public GenerateArchiveCommand(OctoLogger log, ISourceGithubApiFactory sourceGithubApiFactory, EnvironmentVariableProvider environmentVariableProvider) : base("generate-archive")
+        {
+            _log = log;
+            _sourceGithubApiFactory = sourceGithubApiFactory;
+            _environmentVariableProvider = environmentVariableProvider;
+
+            Description = "Invokes the GitHub Migration API's to generate a migration archive";
+            Description += Environment.NewLine;
+            Description += "Note: Expects GH_PAT and GH_SOURCE_PAT env variables to be set. GH_SOURCE_PAT is optional, if not set GH_PAT will be used instead. This authenticates to the source GHES API.";
```
roferg, line 26: This should include that it expects `--ghes-url` as well

src/gei/Commands/GenerateArchiveCommand.cs
```diff
@@ -0,0 +1,150 @@
...
+        public async Task Invoke(string ghesUrl, string githubSourceOrg, string sourceRepo, bool noSslVerify = false, bool verbose = false)
+        {
+            _log.Verbose = verbose;
+
+            _log.LogInformation("Generating Migration Archives...");
+            _log.LogInformation($"GHES SOURCE ORG: {githubSourceOrg}");
+
+            if (string.IsNullOrWhiteSpace(ghesUrl))
+            {
+                _log.LogInformation("ghesUrl not provided, defaulting to https://api.github.com");
```
roferg, line 67: Nit: we should standardize this reference to be `--ghes-url`. Just to make it more clear.

src/gei/Commands/GenerateArchiveCommand.cs
```diff
@@ -0,0 +1,150 @@
...
+                ghesUrl = "https://api.github.com";
+            }
+
+            if (noSslVerify)
+            {
+                _log.LogInformation("No SSL verification enabled");
```
roferg, line 73: Could this just be `SSL verification disabled`

roferg (approved):
Left a few comments regarding the user facing errors and logs.

src/gei/Commands/GenerateArchiveCommand.cs
dylan-smith: --ghes-url is treated the same as any other arg, it will show up in the help in the list of args. We call out the PAT's separately there because they aren't args, but env vars.

docs/GettingStarted.md
```diff
@@ -0,0 +1,41 @@
+# Getting Started Contributing
```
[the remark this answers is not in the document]
oakeyc: sure, added in 5b6bcb

src/Octoshift/GithubApi.cs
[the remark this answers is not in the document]
oakeyc: Sure, seems like an ok change to me

src/Octoshift/GithubClient.cs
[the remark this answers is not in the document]
oakeyc: We want `shouldEnsureSuccessStatusCode` because the API returns a `302` found instead of a 200. Which is the correct response for what we're trying to do

src/Octoshift/GithubEnums.cs
```diff
@@ -0,0 +1,39 @@
...
+{
+    public static class GithubEnums
+    {
+        public enum ArchiveMigrationStatus
+        {
+            Pending,
+            Exporting,
+            Exported,
+            Failed
+        }
```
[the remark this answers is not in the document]
oakeyc: I like it! I'll add it

src/gei/GithubApiFactory.cs
```diff
@@ -5,28 +5,34 @@ namespace OctoshiftCLI.GithubEnterpriseImporter
...
+            var githubClient = new GithubClient(_octoLogger, _clientFactory.CreateClient("Default"), githubPat);
+            return new GithubApi(githubClient);
+        }
+
+        GithubApi ISourceGithubApiFactory.CreateClientNoSSL()
+        {
+            var githubPat = _environmentVariableProvider.SourceGithubPersonalAccessToken();
+            var githubClient = new GithubClient(_octoLogger, _clientFactory.CreateClient("NoSSL"), githubPat);
             return new GithubApi(githubClient);
         }
```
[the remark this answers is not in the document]
oakeyc: Good call out, will add

src/Octoshift/GithubApi.cs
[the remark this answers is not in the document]
oakeyc: Adding the string repo as suggested in another comment

src/gei/Commands/GenerateArchiveCommand.cs
```diff
@@ -0,0 +1,150 @@
...
+
+            var githubApi = noSslVerify ? _sourceGithubApiFactory.CreateClientNoSSL() : _sourceGithubApiFactory.Create();
+
+            var repositories = new string[] { sourceRepo };
+
+            var gitDataOptions = new
+            {
+                repositories,
+                exclude_metadata = true
+            };
```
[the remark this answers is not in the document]
oakeyc: Passing in a string repo as described in another comment

src/gei/Commands/GenerateArchiveCommand.cs
```diff
@@ -0,0 +1,150 @@
...
+
+            _log.LogInformation($"Archive generation of git data started with id: {gitDataArchiveId}");
+
+            var metadataOptions = new
+            {
+                repositories,
+                exclude_git_data = true,
+                exclude_releases = true,
+                exclude_owner_projects = true
+            };
```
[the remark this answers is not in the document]
oakeyc: Passing in a string repo as described in another comment

src/gei/Commands/GenerateArchiveCommand.cs
```diff
@@ -0,0 +1,150 @@
...
+        {
+            _log = log;
+            _sourceGithubApiFactory = sourceGithubApiFactory;
+            _environmentVariableProvider = environmentVariableProvider;
+
+            Description = "Invokes the GitHub Migration API's to generate a migration archive";
+            Description += Environment.NewLine;
+            Description += "Note: Expects GH_PAT and GH_SOURCE_PAT env variables to be set. GH_SOURCE_PAT is optional, if not set GH_PAT will be used instead. This authenticates to the source GHES API.";
+
+            var ghesUrl = new Option<string>("--ghes-url")
```
petermeglis, line 28: IMO (as you saw in our 🍐 session) it's not obvious that this is the api url and not just the regular ghes url. Could we add a description here to clarify that this is the `...api.<ghes_url>` and/or update the name to something like `ghesApiUrl`?

src/gei/Commands/GenerateArchiveCommand.cs
```diff
@@ -0,0 +1,150 @@
...
+namespace OctoshiftCLI.GithubEnterpriseImporter.Commands
+{
+    public class GenerateArchiveCommand : Command
+    {
+        private readonly OctoLogger _log;
+        private readonly ISourceGithubApiFactory _sourceGithubApiFactory;
+        private readonly EnvironmentVariableProvider _environmentVariableProvider;
+
+        private const int _timeoutInHours = 10;
+        private const int _delayInMS = 10000; // 10 seconds
```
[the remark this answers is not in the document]
oakeyc: Sounds good, will update

src/gei/Commands/GenerateArchiveCommand.cs
```diff
@@ -0,0 +1,150 @@
...
+                {
+                    if (pair.Value)
+                    {
+                        continue;
+                    }
+                    var id = pair.Key;
+                    var archiveStatus = await githubApi.GetArchiveMigrationStatus(ghesUrl, githubSourceOrg, id);
+                    var stringStatus = GithubEnums.ArchiveMigrationStatusToString(archiveStatus);
+
+                    _log.LogInformation($"Waiting for archive with id {id} generation to finish. Current status: {stringStatus}");
```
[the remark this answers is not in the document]
oakeyc: Yup, this line will be edited after the pr feedback commits go up

src/gei/Commands/GenerateArchiveCommand.cs
oakeyc: Sounds good, will add

src/gei/Commands/GenerateArchiveCommand.cs
oakeyc: Yes, it can, and I will change it to the wording you provided 👍

src/gei/Commands/GenerateArchiveCommand.cs
oakeyc: Yup, I'll update the wording, and let me know what you think

src/Octoshift/GithubClient.cs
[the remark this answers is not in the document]
dylan-smith: right, but what I'm saying is if you just pass the status arg (and eliminate the shouldEnsureSuccessStatusCode arg) you can just check if status is set, and if so skip the EnsureSuccess() call and check if it equals 302 (or whatever status is set to)

src/gei/Commands/GenerateArchiveCommand.cs
oakeyc: Sounds good, I can adopt this refactor

src/gei/Commands/GenerateArchiveCommand.cs
oakeyc: Accepting the other comment refactor 👍

docs/GettingStarted.md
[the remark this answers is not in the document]
oakeyc: I agree with @slenguyen as someone who is trying to contribute but hasn't learned all the dotnet things, it's nice to have an approachable place to start.

docs/GettingStarted.md
```diff
@@ -0,0 +1,41 @@
...
+
+Familiarize yourself with [.NET core](https://docs.microsoft.com/en-us/dotnet/core/introduction#net-core-and-net-5), which is a cross-platform framework for .NET (also read "dotnet").
+
+## Basic Getting Started
+
+Try developing this repo using GitHub codespaces, so all the dependencies are installed for you!
```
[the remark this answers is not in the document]
oakeyc: I disagree, and it reads as a suggestion, which it is! And we _do_ suggest people develop with codespaces as it is our main supported path as github engineers. That doesn't mean that other ways are not acceptable. But I'm planning on keeping this line

src/gei/Commands/GenerateArchiveCommand.cs
dylan-smith: you can add a description to the arg, right after where the IsRequired flag is

RELEASENOTES.md
```diff
@@ -1,4 +1,6 @@
+
 - Adds the ability to migrate ADO repos using the `gh gei` CLI. This overlaps with some of the capabilities of ado2gh, but the `gh gei` will not include all the extra ADO migration capabilities like re-wiring pipelines and boards integration.
     - `gh gei generate-script` now has an `--ado-source-org` option
     - `gh gei migrate-repo` now has `--ado-source-org` and `--ado-team-project` options
 - Added `grant-migrator-role` and `revoke-migrator-role` commands to `gh gei`
+- Add gei command path for generating a migration archive
```
dylan-smith, line 6: We should probably elaborate a bit more here. Mention what the new command is, what it does, and that it's specifically for GHES.

src/Octoshift/GithubClient.cs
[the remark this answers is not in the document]
oakeyc: I see, thanks for explaining. It looks like there's no null type for HttpStatusCodes, so it would be checking against the default value set, probably `OK`

src/Octoshift/GithubApi.cs
```diff
@@ -304,5 +305,34 @@ public virtual async Task<bool> RevokeMigratorRole(string org, string actor, str
...
+            var state = (string)data["state"];
+
+            return GithubEnums.StringToArchiveMigrationStatus(state);
+        }
+
+        public virtual async Task<string> GetArchiveMigrationUrl(string apiUrl, string org, int migrationId)
+        {
+            var url = $"{apiUrl}/orgs/{org}/migrations/{migrationId}/archive";
+
+            var response = await _client.GetNonSuccessAsync(url, HttpStatusCode.Found);
```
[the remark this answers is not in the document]
oakeyc: I think this wouldn't be ideal because we actually want the response since the status code is what we're expecting. There's been some refactoring the client side, but I see this as a stylistic difference

src/Octoshift/GithubClient.cs
```diff
@@ -82,7 +88,14 @@ public virtual async IAsyncEnumerable<JToken> GetAllAsync(string url)
...
+            {
+                response.EnsureSuccessStatusCode();
+            }
+            else if (response.StatusCode != status)
+            {
+                throw new HttpRequestException($"Expected status code {status} but got {response.StatusCode}");
+            }
```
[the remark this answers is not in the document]
oakeyc: As said above, I think it fits, and it seems like a stylistic choice :+1:

src/gei/Program.cs
```diff
@@ -23,7 +24,20 @@ public static async Task Main(string[] args)
                 .AddSingleton<GithubApiFactory>()
                 .AddTransient<ITargetGithubApiFactory>(sp => sp.GetRequiredService<GithubApiFactory>())
                 .AddTransient<ISourceGithubApiFactory>(sp => sp.GetRequiredService<GithubApiFactory>())
-                .AddHttpClient();
+                .AddHttpClient("NoSSL")
+                .ConfigurePrimaryHttpMessageHandler(x => new HttpClientHandler()
```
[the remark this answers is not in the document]
oakeyc: Sounds good, I can update that

src/Octoshift/GithubClient.cs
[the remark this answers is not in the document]
ArinGhazarian: I still don't think comparing against 200 would be a good idea since for example `201` also indicates success!.

src/gei/Program.cs
```diff
@@ -23,7 +24,20 @@ public static async Task Main(string[] args)
...
+                    CheckCertificateRevocationList = false,
+                    ServerCertificateCustomValidationCallback = delegate { return true; }
+                });
+
+            serviceCollection
+                .AddHttpClient("Default")
+                .ConfigurePrimaryHttpMessageHandler(x => new HttpClientHandler()
+                {
+                    AllowAutoRedirect = false
+                });
```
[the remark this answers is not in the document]
oakeyc: Yes! Since they might use the ghes client with ssl. Cool, I didn't know about `Services` I'll add that 👍

src/Octoshift/GithubClient.cs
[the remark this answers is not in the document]
oakeyc: As written (this line is out of date), the if the default status hasn't been changed (`OK` 200) then it uses `ensuresuccessstatuscode` which checks 200-299 (it does not check value content). If the default status has changed, it validates against that :+1:

RELEASENOTES.md
oakeyc: Sounds good, I update the wording a little. Let me know what you think

src/Octoshift/GithubApi.cs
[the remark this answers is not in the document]
ArinGhazarian: It is not a stylistic difference that I am suggesting, with what I suggested you would both get the response content and the status code with `SendAsync`. The reason I am against using `GetNonSuccessAsync` is as I said before, it is there to accommodate a bad api design on the GitHub side and it's actually a business rule that we want to apply here. All I am saying it to simply use `SendAsync` if we also want to get the status code back.
To further explain why this is a bad design, imagine a scenario where the same thing happens with POST or PUT, what would you do in those cases? Would you just go ahead and add another method to the client class?
Besides what i am suggesting is a common practice that you can for example see in another popular .NET http  client called [Restsharp](https://restsharp.dev/)

src/Octoshift/GithubClient.cs
```diff
@@ -29,6 +31,8 @@ public GithubClient(OctoLogger log, HttpClient httpClient, string personalAccess
             }
         }
 
+        public virtual async Task<string> GetNonSuccessAsync(string url, HttpStatusCode status) => (await SendAsync(HttpMethod.Get, url, status: status)).Content;
```
ArinGhazarian, line 34: Again I still stand by my original comment and think that this is a bad design in a lower level client. the client should be agnostic and do just the basic stuff. If some API returns 3XX as a success code then we should just accommodate it in our business (ie. `GithubApi`)

src/Octoshift/GithubClient.cs
```diff
@@ -82,7 +86,14 @@ public virtual async IAsyncEnumerable<JToken> GetAllAsync(string url)
             var content = await response.Content.ReadAsStringAsync();
             _log.LogVerbose($"RESPONSE ({response.StatusCode}): {content}");
 
-            response.EnsureSuccessStatusCode();
+            if (status == HttpStatusCode.OK)
+            {
+                response.EnsureSuccessStatusCode();
+            }
```
ArinGhazarian, line 92: This code is semantically wrong. It will never raise any error at all. I still think the passing a flag like before to not raise an error in case of non success status is a better choice. And I have seen it in other popular projects too, it's a common practice.

ArinGhazarian (commented):
I pretty much like everything here except the `GeNonSuccessAsync` in the `GithubClient`. I think it's a bad API design. I can explain more if you want to 😀

RELEASENOTES.md
```diff
@@ -1,4 +1,6 @@
+
 - Adds the ability to migrate ADO repos using the `gh gei` CLI. This overlaps with some of the capabilities of ado2gh, but the `gh gei` will not include all the extra ADO migration capabilities like re-wiring pipelines and boards integration.
     - `gh gei generate-script` now has an `--ado-source-org` option
     - `gh gei migrate-repo` now has `--ado-source-org` and `--ado-team-project` options
 - Added `grant-migrator-role` and `revoke-migrator-role` commands to `gh gei`
+- Add gei command path for generating a migration archive `gh gei generate-archive` which uses the migration api on that instance to generate two archives of data, the metadata for a repository and the git data for that repository
```
dylan-smith, line 6:
```suggestion
- Add gei command path for generating a migration archive `gh gei generate-archive` which uses the migration api on that instance to generate two archives of data, the metadata for a repository and the git data for that repository (this will primarily be useful for migrations from GHES)
```

src/Octoshift/GithubClient.cs
ArinGhazarian: Oh wait sorry, I got confused this code is OK, as is. I confused it with `response.status == HttpStatusCode.OK`

dylan-smith: I don't think it's semantically wrong, if you call SendAsync and don't pass in the status param, then EnsureSuccessStatusCode will be run. If you do pass in a status (presumably something other than OK), then it won't check for success in the Client.

src/gei/Commands/GenerateArchiveCommand.cs
```diff
@@ -0,0 +1,110 @@
...
+
+            Description = "Invokes the GitHub Migration API's to generate a migration archive";
+            Description += Environment.NewLine;
+            Description += Environment.NewLine;
+            Description += "Note: Expects GH_PAT and GH_SOURCE_PAT env variables to be set. GH_SOURCE_PAT is optional, if not set GH_PAT will be used instead. This authenticates to the source GHES API. For GHES, we expect that --ghes-api-url is passed in as the api endpoint for the hostname of your GHES instance. For example: https://api.myghes.com";
+
+            var ghesApiUrl = new Option<string>("--ghes-api-url")
+            {
+                IsRequired = false
+            };
```
dylan-smith, line 31:
```suggestion
            Description += "Note: Expects GH_PAT and GH_SOURCE_PAT env variables to be set. GH_SOURCE_PAT is optional, if not set GH_PAT will be used instead. This authenticates to the source GHES API.";

            var ghesApiUrl = new Option<string>("--ghes-api-url")
            {
                IsRequired = false,
                Description = "The api endpoint for the hostname of your GHES instance. For example: https://api.myghes.com"
            };
```

src/gei/Commands/GenerateArchiveCommand.cs
```diff
@@ -0,0 +1,110 @@
...
+
+            _log.LogInformation("Generating Migration Archives...");
+            _log.LogInformation($"GHES SOURCE ORG: {githubSourceOrg}");
+            _log.LogInformation($"GHES SOURCE REPO: {sourceRepo}");
+
+            if (string.IsNullOrWhiteSpace(ghesApiUrl))
+            {
+                _log.LogInformation("--ghes-api-url not provided, defaulting to https://api.github.com");
+                ghesApiUrl = "https://api.github.com";
+            }
```
dylan-smith, line 70: I would also log ghesApiUrl (when provided)

```suggestion
            }
            else
            {
                _log.LogInformation($"GHES API URL: {ghesApiUrl}");
            }
```

src/gei/Program.cs
```diff
@@ -24,7 +25,19 @@ public static async Task Main(string[] args)
...
                 .AddTransient<ISourceGithubApiFactory>(sp => sp.GetRequiredService<GithubApiFactory>())
-                .AddHttpClient();
+                .AddHttpClient("NoSSL")
+                .ConfigurePrimaryHttpMessageHandler(() => new HttpClientHandler()
+                {
+                    AllowAutoRedirect = false,
+                    CheckCertificateRevocationList = false,
+                    ServerCertificateCustomValidationCallback = delegate { return true; }
+                })
+                .Services
```
dylan-smith, line 35: I'm surprised this even compiles without parens after it. 😮

dylan-smith (changes requested):

src/Octoshift/GithubClient.cs
ArinGhazarian: @dylan-smith yeah it was my bad, the code is ok.

src/gei/Program.cs
ArinGhazarian: why? it's legit code, it's just a property 😃

oakeyc: Turns out to be a property!

dylan-smith (approved):

RELEASENOTES.md
oakeyc: updated

src/gei/Commands/GenerateArchiveCommand.cs
oakeyc: updated

src/gei/Commands/GenerateArchiveCommand.cs
oakeyc: updated

PR #261, opened 2022-03-10
Pass tokens to startMigration

### LINKED ISSUE #257: Add `accessToken` and `githubPat` to `StartMigration`
### Description
Based on [this issue](https://github.com/github/octoshift/issues/3354), Octoshift added the `access_token` and `github_pat` fields to [the `startRepositoryMigration` mutation](https://github.com/github/octoshift/blob/main/proto/migrations/v1/start_migration_api.proto). In order to be in accordance with the BE changes, the CLI needs to pass those fields to the `startRepositoryMigration` API call. 
Note that those fields are going to get removed from the `createMigrationSource` mutation in the future but right now CLI still needs to pass them for backward compatibility. 

## Todo
- [ ] Add the `accessToken` and `githubPat` fields to `createRepositoryMigration` mutation API call (`GithubApi#StartMigration()`).

ArinGhazarian:
Closes #257 

- [x] Issue linked

### DIFF
--- a/src/Octoshift/GithubApi.cs
+++ b/src/Octoshift/GithubApi.cs
@@ -158,12 +158,47 @@ namespace OctoshiftCLI
             return (string)data["data"]["createMigrationSource"]["migrationSource"]["id"];
         }
 
-        public virtual async Task<string> StartMigration(string migrationSourceId, string adoRepoUrl, string orgId, string repo, string gitArchiveUrl = "", string metadataArchiveUrl = "")
+        public virtual async Task<string> StartMigration(string migrationSourceId, string adoRepoUrl, string orgId, string repo, string sourceToken, string targetToken, string gitArchiveUrl = "", string metadataArchiveUrl = "")
         {
             var url = $"{_apiUrl}/graphql";
 
-            var query = "mutation startRepositoryMigration($sourceId: ID!, $ownerId: ID!, $sourceRepositoryUrl: URI!, $repositoryName: String!, $continueOnError: Boolean!, $gitArchiveUrl: String!, $metadataArchiveUrl: String!)";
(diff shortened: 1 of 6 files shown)


src/Octoshift/GithubApi.cs
```diff
@@ -158,12 +158,47 @@ public virtual async Task<string> CreateGhecMigrationSource(string orgId, string
...
+                    input: { 
+                        sourceId: $sourceId,
+                        ownerId: $ownerId,
+                        sourceRepositoryUrl: $sourceRepositoryUrl,
+                        repositoryName: $repositoryName,
+                        continueOnError: $continueOnError,
+                        gitArchiveUrl: $gitArchiveUrl,
+                        metadataArchiveUrl: $metadataArchiveUrl,
+                        accessToken: $accessToken,
+                        githubPat: $githubPat
```
ArinGhazarian, line 187: I am following the same pattern as `Create[Ado/Ghec]MigrationSource()` by not requiring the `githubPat` field which is also not required currently in the BE. Later on as part of making the token fields required we should also make this required too.

src/Octoshift/GithubApi.cs
```diff
@@ -158,12 +158,47 @@ public virtual async Task<string> CreateGhecMigrationSource(string orgId, string
...
+                mutation startRepositoryMigration(
+                    $sourceId: ID!,
+                    $ownerId: ID!,
+                    $sourceRepositoryUrl: URI!,
+                    $repositoryName: String!,
+                    $continueOnError: Boolean!,
+                    $gitArchiveUrl: String!,
+                    $metadataArchiveUrl: String!,
+                    $accessToken: String!,
+                    $githubPat: String)";
```
ArinGhazarian, line 175: I am adding extra tabs for readability but if ppl feel like it's just better not to add those extra tabs I can update the code. Those extra tabs are going to make the testing more tricky to verify the payload since we need to actually match the tabs.

dylan-smith (approved):

PR #267, opened 2022-03-15
Idempotency for create team command - code complete

teja-nanduri:
- [ ] Issue linked

### DIFF
--- a/src/Octoshift/GithubApi.cs
+++ b/src/Octoshift/GithubApi.cs
@@ -68,6 +68,15 @@ namespace OctoshiftCLI
             return (string)data["id"];
         }
 
+        public virtual async Task<List<string>> GetTeams(string org)
+        {
+            var url = $"{_apiUrl}/orgs/{org}/teams";
+
+            return await _client.GetAllAsync(url)
+                .Select(t => (string)t["name"])
+                .ToListAsync();
+        }
+
         public virtual async Task<IEnumerable<string>> GetTeamMembers(string org, string teamName)
         {
             var url = $"{_apiUrl}/orgs/{org}/teams/{teamName}/members?per_page=100";
--- a/src/OctoshiftCLI.Tests/GithubApiTests.cs
+++ b/src/OctoshiftCLI.Tests/GithubApiTests.cs
@@ -139,6 +139,69 @@ namespace OctoshiftCLI.Tests
         }
 
         [Fact]
+        public async Task GetTeams_Returns_All_Teams()
+        {
+            // Arrange
+            const string org = "ORG";
+
+            var url = $"https://api.github.com/orgs/{org}/teams";
+
+            const string team1 = "TEAM_1";
+            const string team2 = "TEAM_2";
+            var responsePage1 = $@"
+            [
+                {{
+                    ""name"": ""{team1}"",
+                    ""id"": 1
+                }},
+                {{
+                    ""name"": ""{team2}"", 
+                    ""id"": 2
+                }}
+            ]";
+
+            const string team3 = "TEAM_3";
+            const string team4 = "TEAM_4";
+            var responsePage2 = $@"
+            [
+                {{
+                    ""name"": ""{team3}"",
+                    ""id"": 3
+                }},
+                {{
+                    ""name"": ""{team4}"", 
+                    ""id"": 4
+                }}
+            ]";
+
+            async IAsyncEnumerable<JToken> GetAllPages()
+            {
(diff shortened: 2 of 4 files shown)


src/OctoshiftCLI.Tests/GithubApiTests.cs
```diff
@@ -138,6 +138,69 @@ public async Task CreateTeam_Returns_Created_Team_Id()
...
+                    ""name"": ""{team3}"",
+                    ""id"": 3
+                }},
+                {{
+                    ""name"": ""{team4}"", 
+                    ""id"": 4
+                }}
+            ]";
+
+            async IAsyncEnumerable<JToken> GetAllPages()
```
dylan-smith, line 177: This may be a question for @ArinGhazarian but I'm expecting/hoping there is a way to write simpler tests without having to mock out paging/JSON in situations like this.

I don't think this test needs to test our paging implementation (we have separate unit tests for that), I would hope there is a way to simplify this test so it has the mock just return a simple list of strings (team names) instead of pages of JSON.

src/Octoshift/GithubApi.cs
```diff
@@ -68,6 +68,15 @@ public virtual async Task<string> CreateTeam(string org, string teamName)
             return (string)data["id"];
         }
 
+        public virtual async Task<List<string>> GetTeams(string org)
```
dylan-smith, line 71:
```suggestion
        public virtual async Task<IEnumerable<string>> GetTeams(string org)
```

dylan-smith (changes requested):

dylan-smith:
What happens if the team already exists but is linked to a different Idp Group?

dylan-smith:
We should also add a bullet to the release notes (should have done it in the autolink PR also, we can cover both in this PR).  Something like: 

- Updates most commands to be idempotent. They will check if there is anything to do, and if not they will print a message to that effect and complete successfully. E.g. create-team will check if the team already exists and if so exit as success (compared to previously where it would crash). The following commands have been updated:
  - configure-autolink
  - create-team

src/OctoshiftCLI.Tests/GithubApiTests.cs
ArinGhazarian: Oh yeah, we most certainly can right simpler tests now. At the time I was writing my test for `GetRepos` I wasn't using `System.Linq.Async` library (to keep things simple), but now that we're using it we can simply convert any `IEnumerable` To `IAsyncEnumerable` with `ToAsyncEnumerable()` extension method like this:

```c#
var asyncEnumerable = new[] { item1, item2 }.ToAsyncEnumerable();
```

so that entire local method can go away and replaced by a `ToAsyncEnumerable` call on any `IEnumerable` 😃

src/ado2gh/Commands/CreateTeamCommand.cs
```diff
@@ -55,9 +55,16 @@ public async Task Invoke(string githubOrg, string teamName, string idpGroup, boo
 
             var githubApi = _githubApiFactory.Create();
 
-            await githubApi.CreateTeam(githubOrg, teamName);
-
-            _log.LogSuccess("Successfully created team");
+            var teams = await githubApi.GetTeams(githubOrg);
+            if (teams.Contains(teamName))
+            {
+                _log.LogSuccess($"Team '{teamName}' already exists. New team will not be created");
```
ArinGhazarian, line 61: nit: don't we need a full stop at the end of the second sentence?

teja-nanduri: Looks like we are not using full stops anywhere else. I am modifying it to `-`

dylan-smith (approved):

src/OctoshiftCLI.Tests/TestExtensionMethods.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using Newtonsoft.Json.Linq;
+
+namespace OctoshiftCLI.Tests
+{
+    public static class TestExtensionMethods
+    {
+        public static IAsyncEnumerable<JToken> ToAsyncJTokenEnumerable<T>(this IEnumerable<T> list) => list.Select(x => JToken.FromObject(x)).ToAsyncEnumerable();
```
ArinGhazarian, line 9: 🎉

ArinGhazarian (approved):

PR #269, opened 2022-03-16
Deprecate `--ssh` flag

### LINKED ISSUE #265: Deprecate the ssh support
### Description
With the parallel migration support changes, because of the async nature of queueing a migration and then asynchronously waiting for it to finish with `wait-for-migraiton` command, the synchronous retry logic in `migrate-repo` command will no longer work (since it just queues up a migration and return). This will specifically cause a migration to fail if it encounters the `ECDSA SSH` error (with synchronous one it would just retry the command).
One way to fix this is to use `https` for repo migrations instead of `ssh`. We introduced the `--ssh` flag as an alternative to `https` to work around some `https` related errors, but now that those errors have been fixed in the BE we can just simply get rid of the `--ssh`.

## Todo
- [x] Remove all `ssh` logic from the code base including the `ECDSA` retry logic from `migrate-repo`.
- [x] Make `--ssh` flag hidden in `migrate-repo` and if provided, log a warning and ignore it.

ArinGhazarian:
Closes #265 

Because of the latest BE changes and the fact that the problems we had with `https` are now resolved, we can deprecate the `--ssh` flag and use `https` permanently. For backward compatibility, we are not going to remove it completely just yet so passing it will do noting (keep using `https`) and log a warning to notify the user. 

- [x] Issue linked


dylan-smith:
should add a bullet to release notes for this

ArinGhazarian:
> should add a bullet to release notes for this

Yup :) while I was adding it I noticed that it'd be a good idea to also test for `option.IsHidden` in the command tests. Adding the ReleaseNotes now 😃

dylan-smith (approved):

dylan-smith (approved):

Issue #274, opened 2022-03-17
Unhandled exception with blank environment variables


digimangos:
## Description

Unhandled exception: System.ArgumentException: String cannot be of zero length. (Parameter 'oldValue')
   at System.String.Replace(String oldValue, String newValue)
   at OctoshiftCLI.OctoLogger.MaskSecrets(String msg)
   at OctoshiftCLI.OctoLogger.LogError(Exception ex)
   at OctoshiftCLI.AdoToGithub.Program.<>c.<BuildParser>b__2_0(Exception ex, InvocationContext _)
   at System.CommandLine.Builder.CommandLineBuilderExtensions.<>c__DisplayClass14_0.<<UseExceptionHandler>b__0>d.MoveNext()
--- End of stack trace from previous location ---
   at System.CommandLine.Builder.CommandLineBuilderExtensions.<>c__DisplayClass14_0.<<UseExceptionHandler>b__0>d.MoveNext()

## Reproduction Steps

I'm using ado2gh, and received the following exception because my environment variables were set as empty strings.

dylan-smith:
I'm trying to reproduce this error so I can give some advice on testing, but I'm struggling to repro it.

What command were you running to get this error?

digimangos:
Hi @dylan-smith, basically if you do something like:

```
export ADO_PAT=""
./ado2gh migrate-repo --ado-org xxx --ado-team-project xxx --ado-repo xxx --github-org xxx --github-repo xxx
```

In the previous build you should see the error, and after my change you should get the message that:
`ADO_PAT environment variable is not set.`

🤞

PR #275, opened 2022-03-17
Treat empty variables as not being set

digimangos:
Proposed change to avoid the problem reported in issue https://github.com/github/gh-gei/issues/274.

  - Unfortunately it doesn't seem easily testable when using .net for cross platform - see issue
- [X] Issue linked

### DIFF
--- a/src/ado2gh/EnvironmentVariableProvider.cs
+++ b/src/ado2gh/EnvironmentVariableProvider.cs
@@ -20,8 +20,9 @@ public class EnvironmentVariableProvider
 
     private string GetSecret(string secretName)
     {
-        var secret = Environment.GetEnvironmentVariable(secretName) ??
-                     throw new OctoshiftCliException($"{secretName} environment variable is not set.");
+        var secret = Environment.GetEnvironmentVariable(secretName);
+        if (string.IsNullOrEmpty(secret))
+            throw new OctoshiftCliException($"{secretName} environment variable is not set.");
 
         _logger?.RegisterSecret(secret);
 


dylan-smith:
CI doesn't work properly when external contributers create a PR, I have a trick I can do to make it run properly. Will do it in an hour or so.

digimangos:
> CI doesn't work properly when external contributers create a PR, I have a trick I can do to make it run properly. Will do it in an hour or so.

Glad you have a trick, I was trying to figure out why it couldn't read back the JSON from the API.

dylan-smith:
This works good for ado2gh, but ideally we'd do the same thing for gei. But this approach won't work without changes in gei, because there are some scenarios where if the secret is blank, we fall back to a different env var (e.g. if GH_SOURCE_PAT is blank we'll grab GH_PAT instead, this change - if we made it in gei - would break that code).

dylan-smith:
> This works good for ado2gh, but ideally we'd do the same thing for gei. But this approach won't work without changes in gei, because there are some scenarios where if the secret is blank, we fall back to a different env var (e.g. if GH_SOURCE_PAT is blank we'll grab GH_PAT instead, this change - if we made it in gei - would break that code).

After doing a little digging IMO we should copy how we've done it in at least one place in gei EnvironmentVariableProvider, and do a similar thing everywhere:

```
public virtual string TargetGithubPersonalAccessToken() =>
  GetSecret(TARGET_GH_PAT) ??
  throw new OctoshiftCliException($"{TARGET_GH_PAT} environment variable is not set.");
```

https://github.com/github/gh-gei/blob/main/src/gei/EnvironmentVariableProvider.cs

We can/should do the same thing on AdoPersonalAccessToken()/AzureStorageConnectionString() in gei.EnvironmentVariableProvider, and also on GithubPersonalAccessToken()/AdoPersonalAccessToken() in ado2gh.EnvironmentVariableProvider

@digimangos if you want to make those changes I can merge this in.

dylan-smith:
We should also have unit tests, there's an example unit test we can copy here: https://github.com/github/gh-gei/blob/76470079e6161a65cda39dc8ec0125de013f6dd6/src/OctoshiftCLI.Tests/gei/EnvironmentVariableProviderTests.cs#L78

although that example unit test seems to have a misleading (wrong) name, but it's doing the right stuff.

digimangos:
> We should also have unit tests, there's an example unit test we can copy here:
> 
> https://github.com/github/gh-gei/blob/76470079e6161a65cda39dc8ec0125de013f6dd6/src/OctoshiftCLI.Tests/gei/EnvironmentVariableProviderTests.cs#L78
> 
> although that example unit test seems to have a misleading (wrong) name, but it's doing the right stuff.

As I mentioned in the issue, I can't see how to make a unit test for this scenario because:

`Environment.SetEnvironmentVariable("GH_PAT", null);`
and
`Environment.SetEnvironmentVariable("GH_PAT", "");`

Both do the same thing, which is to unset the variable. In the scenario that I have, the variable in set in OSX with an empty string, but there is not the .net library to do this for the purpose of writing the tests.

digimangos:
@dylan-smith, I've also just checked with "gei" and get the same error:

`[17:33] [INFO] Actor type is valid...
Unhandled exception: System.ArgumentException: String cannot be of zero length. (Parameter 'oldValue')
   at System.String.Replace(String oldValue, String newValue)
   at OctoshiftCLI.OctoLogger.MaskSecrets(String msg)
   at OctoshiftCLI.OctoLogger.LogError(Exception ex)
   at OctoshiftCLI.GithubEnterpriseImporter.Program.<>c.<BuildParser>b__2_0(Exception ex, InvocationContext _)
   at System.CommandLine.Builder.CommandLineBuilderExtensions.<>c__DisplayClass14_0.<<UseExceptionHandler>b__0>d.MoveNext()
--- End of stack trace from previous location ---
   at System.CommandLine.Builder.CommandLineBuilderExtensions.<>c__DisplayClass14_0.<<UseExceptionHandler>b__0>d.MoveNext()
`

digimangos:
@dylan-smith, I've addressed the issue in gei and aligned the behaviour in ado2gh.EnvironmentVariableProvider. Would you also like the same environment variable fallback for GH_SOURCE_PAT to GH_PAT?

digimangos:
@dylan-smith so I think this should resolve the big, and align the implementations. I think the build is stuck as before 😢 

Not sure what we can do about the tests though. It seems in windows you can do it by using a DllImport on kernal32.dll, but that's not going to enable us to test in osx and ubuntu, unfortunately. Looking for an idea on that one, or we'll just have to accept that we can't readily test that scenario.

digimangos:
@dylan-smith, I was discussing this with @igwejk earlier and concluded that the way `EnvironmentVariableProvider.cs` is tested is not ideal because it's relying on the behaviour of `System.Environment`.

My proposal therefore is that use of the static `Environment.GetEnvironmentVariable` should be covered by a wrapper class since it's an external dependancy and then mocked out in the test:
https://github.com/github/gh-gei/blob/39045ea2a74686d1776940591f7cc78e6b49db5a/src/ado2gh/EnvironmentVariableProvider.cs#L23

How do you feel about those changes before I make them? Also, I'm thinking this should live within the Octoshift project and be referenced in Ado2Gh project? That way the wrapper can be used by both.

ArinGhazarian:
> @dylan-smith, I was discussing this with @igwejk earlier and concluded that the way `EnvironmentVariableProvider.cs` is tested is not ideal because it's relying on the behaviour of `System.Environment`.
> 
> My proposal therefore is that use of the static `Environment.GetEnvironmentVariable` should be covered by a wrapper class since it's an external dependancy and then mocked out in the test:
> 
> https://github.com/github/gh-gei/blob/39045ea2a74686d1776940591f7cc78e6b49db5a/src/ado2gh/EnvironmentVariableProvider.cs#L23
> 
> How do you feel about those changes before I make them? Also, I'm thinking this should live within the Octoshift project and be referenced in Ado2Gh project? That way the wrapper can be used by both.

Hi @digimangos 👋 , the reason we created the `EnvironmentVariableProvider` is to create a wrapper around the static `GetEnvironmentVairable()` method.  So anywhere else we can just mock it. There is just one unit test class that tests the `EnvironmentVariableProvider` itself in both `gei` and `ado2gh` and they won't interfere with each other since both are in a same collection.

digimangos:
Hi @ArinGhazarian, I did realise that the `EnvironmentVariableProvider` is a wrapper class of sorts, but there is a flaw... that the business logic that's been added to this class cannot be tested in isolation from the implementation of `System.Environment` by MSFT, which should be assumed to work correctly. Further `SetEnvironmentVariable()` introduces further logic, which prevents the testing of the scenario that I've mentioned. See:
https://github.com/dotnet/runtime/blob/3e8e0b8cf05636a91555be69c7b74b3fd84dc786/src/mono/System.Private.CoreLib/src/System/Environment.Unix.Mono.cs#L34

The windows implementation is similar, that you are unable to set an empty string value, but we do need to test the scenario that an empty string is returned by `GetEnvironmentVariable()`, because that's the issue I'm trying to resolve. However, with the current arrangement, that's not possible.

I'm completely happy to write the code to resolve this in anyway acceptable to the team, but as it stands I don't see a way forward at the moment without a suggestion as to how I might go about it. 😢

igwejk:
@digimangos have discussed additionally implementing `GetEnvironmentVairable` and `SetEnvironmentVairable` methods in the existing wrapper [EnvironmentVariableProvider.cs]. Consequently we can use a `MockEnvironmentVariableProvider` for tests.

@ArinGhazarian @dylan-smith do you agree with updating the change according to that proposition or would you like to offer a different solution?

[EnvironmentVariableProvider.cs]: https://github.com/github/gh-gei/blob/main/src/ado2gh/EnvironmentVariableProvider.cs

ArinGhazarian:
> Hi @ArinGhazarian, I did realise that the `EnvironmentVariableProvider` is a wrapper class of sorts, but there is a flaw... that the business logic that's been added to this class cannot be tested in isolation from the implementation of `System.Environment` by MSFT, which should be assumed to work correctly. Further `SetEnvironmentVariable()` introduces further logic, which prevents the testing of the scenario that I've mentioned. See: https://github.com/dotnet/runtime/blob/3e8e0b8cf05636a91555be69c7b74b3fd84dc786/src/mono/System.Private.CoreLib/src/System/Environment.Unix.Mono.cs#L34
> 
> The windows implementation is similar, that you are unable to set an empty string value, but we do need to test the scenario that an empty string is returned by `GetEnvironmentVariable()`, because that's the issue I'm trying to resolve. However, with the current arrangement, that's not possible.
> 
> I'm completely happy to write the code to resolve this in anyway acceptable to the team, but as it stands I don't see a way forward at the moment without a suggestion as to how I might go about it. 😢

@digimangos thanks for the explanation. I wasn't aware of the fact that the API doesn't let you set an empty string as a value! One thing that we can do is to use a `Func` to wrap `Environment.GetEnvironmentVariable` (I believe we only use `Environment.SetEnvironmentVariable` in the unit tests only, so no need to worry about it), this way we can easily make it return whatever we want, so the `EnvironmentVariableProvider` would look like this:
```c#
public class EnvironmentVariableProvider
{
  private readonly Func<string> _getEnvironmentVariable;

  public EnvironmentVariableProvider(OctoLogger logger) : this(logger, v => Environment.GetEnvironmentVariable(v))
  {
  }

  public EnvironmentVariableProvider(OctoLogger logger, Func<string, string> getEnvironmentVariable)
  {
    _loger = logger;
    _getEnvironmentVariable = getEnvironmentVariable;
  }

  ...
}
```

let me know what you think.

cc: @dylan-smith @igwejk

digimangos:
@ArinGhazarian, I'm from the school of thought that, anything can be solved by adding another layer of abstraction, but this could work. 😊 I've made the special constructor internal as it's only really for the purpose of overriding `System.Enviroment.GetEnvironmentVariable()` for testing.

I've added two new tests, where the environment variable is returned as an empty string.

I did look to implement the same in the `gei` project, but the implementation is more complicated because that uses null coalesced outputs from GetEnvironmentVariable, which would then mean adding logic to the new function being passed in which feels a bit more clunky. ☹️ 

Have a look at the code changes and let me know if you also want me to do something with `gei` or revert the latest changes.

@dylan-smith I think the build is stuck on the PR integration again ☹️ 

cc: @igwejk

src/OctoshiftCLI.Tests/ado2gh/EnvironmentVariableProviderTests.cs
```diff
@@ -58,6 +58,18 @@ public void GithubPersonalAccessToken_Throws_If_Github_Pat_Is_Not_Set()
                 .Should().Throw<OctoshiftCliException>();
         }
 
+        [Fact]
+        public void GithubPersonalAccessToken_Throws_If_Github_Pat_Is_Empty_String()
+        {
+            // Arrange
+            ResetEnvs();
+            var altEnvironmentVariableProvider = new EnvironmentVariableProvider(_mockLogger.Object, v => string.Empty);
```
ArinGhazarian, line 66: ❤️

src/ado2gh/EnvironmentVariableProvider.cs
```diff
@@ -9,19 +12,34 @@ public class EnvironmentVariableProvider
...
     private readonly OctoLogger _logger;
 
-    public EnvironmentVariableProvider(OctoLogger logger)
+    private readonly Func<string, string> _getEnvironmentVariable;
+
+    public EnvironmentVariableProvider(OctoLogger logger) : this(logger, v => Environment.GetEnvironmentVariable(v))
+    {
+    }
+
+    internal EnvironmentVariableProvider(OctoLogger logger, Func<string, string> getEnvironmentVariable)
```
ArinGhazarian, line 21: ❤️

ArinGhazarian (approved):
Looks solid to me 💯  🥇

digimangos:
> Looks solid to me 💯 🥇

Thanks @ArinGhazarian

@dylan-smith are you able to fix the magic which makes this build again?

PR #284, opened 2022-03-18
Idempotency - disable repo command

teja-nanduri:
Fixes: #287 

This PR makes the disable repo command idempotent

- [x] Issue linked

### DIFF
--- a/src/Octoshift/AdoApi.cs
+++ b/src/Octoshift/AdoApi.cs
@@ -64,13 +64,13 @@ namespace OctoshiftCLI
             return data.Select(x => (string)x["name"]).ToList();
         }
 
-        public virtual async Task<IEnumerable<string>> GetNonDisabledRepos(string org, string teamProject) => (await GetRepos(org, teamProject)).Where(x => !x.IsDisabled).Select(x => x.Name).ToList();
+        public virtual async Task<IEnumerable<string>> GetEnabledRepos(string org, string teamProject) => (await GetRepos(org, teamProject)).Where(x => !x.IsDisabled).Select(x => x.Name).ToList();
 
         public virtual async Task<IEnumerable<(string Id, string Name, bool IsDisabled)>> GetRepos(string org, string teamProject)
         {
             var url = $"https://dev.azure.com/{org}/{teamProject}/_apis/git/repositories?api-version=6.1-preview.1";
             var data = await _client.GetWithPagingAsync(url);
-            return data.Select(x => ((string)x["id"], (string)x["name"], ((string)x["isDisabled"]).Equals("True", StringComparison.OrdinalIgnoreCase)))
+            return data.Select(x => ((string)x["id"], (string)x["name"], ((string)x["isDisabled"]).ToBool()))
                        .ToList();
         }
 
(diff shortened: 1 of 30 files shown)


src/Octoshift/AdoApi.cs
```diff
@@ -64,12 +64,13 @@ public virtual async Task<IEnumerable<string>> GetTeamProjects(string org)
...
-        public virtual async Task<IEnumerable<string>> GetRepos(string org, string teamProject)
+        public virtual async Task<IEnumerable<string>> GetNonDisabledRepos(string org, string teamProject) => (await GetRepos(org, teamProject)).Where(x => !x.IsDisabled).Select(x => x.Name).ToList();
+
+        public virtual async Task<IEnumerable<(string Id, string Name, bool IsDisabled)>> GetRepos(string org, string teamProject)
         {
             var url = $"https://dev.azure.com/{org}/{teamProject}/_apis/git/repositories?api-version=6.1-preview.1";
             var data = await _client.GetWithPagingAsync(url);
-            return data.Where(x => ((string)x["isDisabled"]).ToUpperInvariant() == "FALSE")
-                       .Select(x => (string)x["name"])
+            return data.Select(x => ((string)x["id"], (string)x["name"], ((string)x["isDisabled"]).Equals("True", StringComparison.OrdinalIgnoreCase)))
```
ArinGhazarian, line 73: nit: since we're refactoring, it'd be better to just parse the bool instead of comparison:
`bool.Parse((string)x["isDisabled"])`, it will parse to a bool value regardless of casing. We can even use `TryParse` if we expect an invalid bool value. You can also take it a step further if you like with an extension method for any string 😃  like this:

```c#
public static bool ToBool(this string s) => bool.TryParse(s, out var result) && result;
```

src/Octoshift/AdoApi.cs
```diff
@@ -64,12 +64,13 @@ public virtual async Task<IEnumerable<string>> GetTeamProjects(string org)
             return data.Select(x => (string)x["name"]).ToList();
         }
 
-        public virtual async Task<IEnumerable<string>> GetRepos(string org, string teamProject)
+        public virtual async Task<IEnumerable<string>> GetNonDisabledRepos(string org, string teamProject) => (await GetRepos(org, teamProject)).Where(x => !x.IsDisabled).Select(x => x.Name).ToList();
```
ArinGhazarian, line 67: wouldn't it be better to call this `GetEnabledRepos`?

src/Octoshift/AdoApi.cs
```diff
@@ -275,6 +276,7 @@ public virtual async Task<string> GetRepoId(string org, string teamProject, stri
                 url = $"https://dev.azure.com/{org}/{teamProject}/_apis/git/repositories?api-version=4.1";
 
                 var response = await _client.GetWithPagingAsync(url);
+                Console.WriteLine(response);
```
ArinGhazarian, line 279: This is probably a mistake!

teja-nanduri: Oops. Forgot to remove it.

ArinGhazarian (approved):
👍

PR #289, opened 2022-03-22
Accept PATs as CLI args

### LINKED ISSUE #4: Accept PAT's as cli args in addition to env vars
I did this as env vars to start because it's usually a more convenient way to pass around secrets, especially if you may be screen sharing while using the CLI. And you definately don't want secrets to be included in the generated script from the generate-script command.

For convenience though, an option to specify them directly as CLI args would be a nice addition.

ArinGhazarian:
Closes #4 

### Description

1. Refactored API factories in both `ado2gh` and `gei` projects to support passing in the PATs as args. 
2. For simplicity and ease of use, I unified the `Create` methods in API factories into a single one with `null` default values. 
3. In the `gh gei` project, I had to refactor the `migrate-repo` command in order to:
a) make it possible to have more than 16 command args. For more that 16 args we cannot use the `Func` invoke anymore and we need to switch to a container class that has all the command line args.  See [this](https://github.com/dotnet/command-line-api/issues/458) issue as reference.
b) simplify the method a bit to reduce the [cyclomatic complexity](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1502). I did the bare minimum and while I could go even further, I thought it would be better to do them in a separate `tech-debt` issue if need be.

### Sample command output
```
./ado2gh add-team-to-repo --github-org "org" --github-repo "repo" --team "team" --role "maintain" --github-pat "gh-pat"
[11:17 AM] [INFO] Adding team to repo...
[11:17 AM] [INFO] GITHUB ORG: org
[11:17 AM] [INFO] GITHUB REPO: repo
[11:17 AM] [INFO] TEAM: team
[11:17 AM] [INFO] ROLE: maintain
[11:17 AM] [INFO] GITHUB PAT: ***
[11:17 AM] [INFO] Successfully added team to repo
```

- [x] Issue linked

### DIFF
--- a/src/ado2gh/Commands/AddTeamToRepoCommand.cs
+++ b/src/ado2gh/Commands/AddTeamToRepoCommand.cs
@@ -36,6 +36,10 @@ namespace OctoshiftCLI.AdoToGithub.Commands
                 IsRequired = true,
                 Description = "The only valid values are: pull, push, admin, maintain, triage. For more details see https://docs.github.com/en/rest/reference/teams#add-or-update-team-repository-permissions, custom repository roles are not currently supported."
             };
+            var githubPat = new Option<string>("--github-pat")
+            {
+                IsRequired = false
+            };
             var verbose = new Option("--verbose")
             {
                 IsRequired = false
@@ -45,12 +49,13 @@ namespace OctoshiftCLI.AdoToGithub.Commands
             AddOption(githubRepo);
             AddOption(team);
             AddOption(role.FromAmong("pull", "push", "admin", "maintain", "triage"));
+            AddOption(githubPat);
             AddOption(verbose);
 
-            Handler = CommandHandler.Create<string, string, string, string, bool>(Invoke);
+            Handler = CommandHandler.Create<string, string, string, string, string, bool>(Invoke);
         }
 
-        public async Task Invoke(string githubOrg, string githubRepo, string team, string role, bool verbose = false)
+        public async Task Invoke(string githubOrg, string githubRepo, string team, string role, string githubPat = null, bool verbose = false)
         {
             _log.Verbose = verbose;
 
@@ -59,8 +64,12 @@ namespace OctoshiftCLI.AdoToGithub.Commands
             _log.LogInformation($"GITHUB REPO: {githubRepo}");
             _log.LogInformation($"TEAM: {team}");
             _log.LogInformation($"ROLE: {role}");
+            if (githubPat is not null)
+            {
+                _log.LogInformation("GITHUB PAT: ***");
+            }
 
-            await _githubApiFactory.Create().AddTeamToRepo(githubOrg, githubRepo, team, role);
+            await _githubApiFactory.Create(personalAccessToken: githubPat).AddTeamToRepo(githubOrg, githubRepo, team, role);
 
             _log.LogSuccess("Successfully added team to repo");
         }
--- a/src/ado2gh/Commands/AddTeamToRepoCommand.cs
+++ b/src/ado2gh/Commands/AddTeamToRepoCommand.cs
@@ -17,7 +17,7 @@ namespace OctoshiftCLI.AdoToGithub.Commands
 
             Description = "Adds a team to a repo with a specific role/permission";
             Description += Environment.NewLine;
-            Description += "Note: Expects GH_PAT env variable to be set.";
+            Description += "Note: Expects GH_PAT env variable or --github-pat option to be set.";
 
             var githubOrg = new Option<string>("--github-org")
             {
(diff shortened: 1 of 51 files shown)


src/ado2gh/GithubApiFactory.cs
```diff
@@ -15,15 +17,11 @@ public GithubApiFactory(OctoLogger octoLogger, HttpClient client, EnvironmentVar
...
-        {
-            return Create("https://api.github.com");
-        }
-
-        public virtual GithubApi Create(string apiUrl)
+        public virtual GithubApi Create(string apiUrl = null, string personalAccessToken = null)
         {
-            var githubPat = _environmentVariableProvider.GithubPersonalAccessToken();
-            var githubClient = new GithubClient(_octoLogger, _client, githubPat);
+            apiUrl ??= DEFAULT_API_URL;
```
iomekam, line 22: Cool, so this should allow us to eventually do GHES migrations?

src/gei/Commands/GenerateScriptCommand.cs
```diff
@@ -175,18 +195,17 @@ public GenerateScriptCommand(OctoLogger log, ISourceGithubApiFactory sourceGithu
             }
         }
 
-        private async Task<string> InvokeGithub(string githubSourceOrg, string githubTargetOrg, string ghesApiUrl, string azureStorageConnectionString, bool noSslVerify, bool sequential)
+        private async Task<string> InvokeGithub(string githubSourceOrg, string githubTargetOrg, string ghesApiUrl, string azureStorageConnectionString, bool noSslVerify, bool sequential, string githubSourcePat)
         {
-            var targetApiUrl = "https://api.github.com";
-            var repos = await GetGithubRepos(_sourceGithubApiFactory.Create(targetApiUrl), githubSourceOrg);
+            var repos = await GetGithubRepos(_sourceGithubApiFactory.Create(sourcePersonalAccessToken: githubSourcePat), githubSourceOrg);
```
iomekam, line 200: Why don't we need to pass in a url here anymore? Because the PAT has is enough information? I've seen this in other places as well, so this question is more of a general one.

iomekam (approved):
Changes look good. I left a few minor questions, but overall, this change looks to just be plumbing. I'm going to sign off, but just an fyi that I'm still learning the codebase.

src/ado2gh/GithubApiFactory.cs
ArinGhazarian: I guess it should be supported right now even 😃

src/ado2gh/Commands/AddTeamToRepoCommand.cs
```diff
@@ -36,6 +36,10 @@ public AddTeamToRepoCommand(OctoLogger log, GithubApiFactory githubApiFactory) :
                 IsRequired = true,
                 Description = "The only valid values are: pull, push, admin, maintain, triage. For more details see https://docs.github.com/en/rest/reference/teams#add-or-update-team-repository-permissions, custom repository roles are not currently supported."
             };
+            var githubPat = new Option<string>("--github-pat")
+            {
+                IsRequired = false
```
dylan-smith, line 41: we should update the description on the command and/or on this option to explain that the user can provide the PAT either as an arg or env var. We should do this for all the commands.

src/gei/Commands/MigrateRepoCommand.cs
```diff
@@ -103,6 +104,18 @@ public MigrateRepoCommand(OctoLogger log, ISourceGithubApiFactory sourceGithubAp
                 IsRequired = false,
                 Description = "Synchronously waits for the repo migration to finish."
             };
+            var githubSourcePat = new Option<string>("--github-source-pat")
+            {
+                IsRequired = false
+            };
+            var githubPat = new Option<string>("--github-pat")
```
dylan-smith, line 111: I'd suggest calling this --github-target-pat, the env var is GH_PAT because we wanted to be consistent across many commands with env var names. For in-line args that matters less, and would be more consistent with --github-source-org --github-target-org

dylan-smith (changes requested):
Once the descriptions and the one arg name are changed this'll be good to go

src/gei/Commands/MigrateRepoCommand.cs
ArinGhazarian: Addressed in e4c861708a7a68437149e65c058ee3825a020e0f

src/ado2gh/Commands/AddTeamToRepoCommand.cs
ArinGhazarian: addressed in 2209c660837f608ba28a10ce49fbc00dad93887f

dylan-smith (approved):
:shipit:

Issue #292, opened 2022-03-28
Improving Waiting for Migration Logic


oakeyc:
## Description

When we receive a 502 on waiting for migration, we should wait a little and retry. We should only fail out if the status is that it returned failed or success or after a timeout period

cc @roferg

PR #309, opened 2022-04-01
Retry when wait-for-migration encounters 502 errors

dylan-smith:
Issue #292 

Implemented this using the Polly library.

In the GithubApi layer, it can decide on a call by call basis whether retries are appropriate, and if so wrap the call to GithubClient in the RetryPolicy:

```
var response = await _retryPolicy.Retry(async () => await _client.PostAsync(url, payload),
                                        ex => ex.StatusCode == HttpStatusCode.BadGateway);
```

The RetryPolicy class is a simple wrapper around Polly to set the retry logic so it's consistent whenever we use it. In this case 5 retries with a 200ms wait between each one.

- [x] Issue linked

### DIFF
--- a/src/Octoshift/RetryPolicy.cs
+++ b/src/Octoshift/RetryPolicy.cs
@@ -2,7 +2,6 @@
 using System.Net.Http;
 using System.Threading.Tasks;
 using Polly;
-using Polly.Contrib.WaitAndRetry;
 
 namespace OctoshiftCLI
 {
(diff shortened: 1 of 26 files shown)


src/Octoshift/RetryPolicy.cs
```diff
@@ -0,0 +1,30 @@
...
+        private readonly OctoLogger _log;
+
+        public RetryPolicy(OctoLogger log)
+        {
+            _log = log;
+        }
+
+        public async Task<T> Retry<T>(Func<Task<T>> func, Func<HttpRequestException, bool> filter)
+        {
+            var delay = Backoff.LinearBackoff(TimeSpan.FromMilliseconds(200), retryCount: 5, fastFirst: true);
```
ArinGhazarian, line 20: While there is nothing wrong with using `Polly.Contrib.WaitAndRetry`, we can also just use the `sleepDurationProvider` of `WaitAndRetryAsync` and not include that package at all. Something like this:

```c#
var retryPolicy = Policy
    .Handle<HttpRequestException>(filter)
    .WaitAndRetryAsync(5, retryCount => retryCount * TimeSpan.FromSeconds(200));
```

 with `sleepDurationProvder` we can cover liner and exponential backoffs. But if you think we're going to end up with more complicated scenarios then it's ok 😃

ArinGhazarian (approved):
The implementation is solid 💯  IMO, I just had a nitpick but will leave it up to you 😃

src/Octoshift/RetryPolicy.cs
dylan-smith: either way is fine with me, 1 less nuget package sounds like a win though so I'll make the change.

ArinGhazarian: Yeah exactly, then if need be we can always add it later 😃

ArinGhazarian (approved):
🥇

Issue #215, opened 2022-01-26
Notify customers if they are using an old version of the CLI


dylan-smith:
## Description

Would require the CLI to call out to the CLI repo to check for latest version.

dylan-smith:
Possibly even prevent the usage of old versions entirely.

iomekam:
CLI version was published on 3/28/22

PR #331, opened 2022-04-06
Version check against latest release

dylan-smith:
Issue #215 

Every time the CLI runs it will check the CLI Releases page to get the latest version and compare against the running version. It will notify the user whether they are on the latest, and if not how to get the latest.

Any failure with trying to perform the version check will be swallowed and a generic message about being unable to determine latest version will be output. The CLI command will continue execution as normal.

Note: I deliberately did not include unit tests for this change. IMO the amount of effort required to get this covered by unit tests is not worth it.  Looking for other opinions on that though.

- [x] Issue linked

### DIFF
--- a/src/Octoshift/AdoClient.cs
+++ b/src/Octoshift/AdoClient.cs
@@ -15,7 +15,7 @@ namespace OctoshiftCLI
         private readonly OctoLogger _log;
         private double _retryDelay;
 
-        public AdoClient(OctoLogger log, HttpClient httpClient, string personalAccessToken)
+        public AdoClient(OctoLogger log, HttpClient httpClient, VersionChecker versionChecker, string personalAccessToken)
         {
             _log = log;
             _httpClient = httpClient;
@@ -25,6 +25,7 @@ namespace OctoshiftCLI
                 _httpClient.DefaultRequestHeaders.Add("accept", "application/json");
                 var authToken = Convert.ToBase64String(Encoding.ASCII.GetBytes($":{personalAccessToken}"));
                 _httpClient.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Basic", authToken);
+                _httpClient.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue("OctoshiftCLI", versionChecker?.GetCurrentVersion()));
             }
         }
 
--- a/src/Octoshift/AdoClient.cs
+++ b/src/Octoshift/AdoClient.cs
@@ -15,7 +15,7 @@ namespace OctoshiftCLI
         private readonly OctoLogger _log;
         private double _retryDelay;
 
-        public AdoClient(OctoLogger log, HttpClient httpClient, VersionChecker versionChecker, string personalAccessToken)
+        public AdoClient(OctoLogger log, HttpClient httpClient, IVersionProvider versionProvider, string personalAccessToken)
         {
             _log = log;
             _httpClient = httpClient;
@@ -25,7 +25,7 @@ namespace OctoshiftCLI
                 _httpClient.DefaultRequestHeaders.Add("accept", "application/json");
                 var authToken = Convert.ToBase64String(Encoding.ASCII.GetBytes($":{personalAccessToken}"));
                 _httpClient.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Basic", authToken);
-                _httpClient.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue("OctoshiftCLI", versionChecker?.GetCurrentVersion()));
+                _httpClient.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue("OctoshiftCLI", versionProvider?.GetCurrentVersion()));
             }
         }
 
--- a/src/Octoshift/GithubClient.cs
+++ b/src/Octoshift/GithubClient.cs
@@ -26,7 +26,7 @@ namespace OctoshiftCLI
             {
                 _httpClient.DefaultRequestHeaders.Add("Accept", "application/vnd.github.v3+json");
                 _httpClient.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue("OctoshiftCLI", versionChecker?.GetCurrentVersion()));
-                _httpClient.DefaultRequestHeaders.Add("GraphQL-Features", "import_api");
+                _httpClient.DefaultRequestHeaders.Add("GraphQL-Features", "import_api,mannequin_claiming");
                 _httpClient.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Bearer", personalAccessToken);
             }
         }
--- a/src/Octoshift/GithubClient.cs
+++ b/src/Octoshift/GithubClient.cs
@@ -17,7 +17,7 @@ namespace OctoshiftCLI
         private readonly HttpClient _httpClient;
         private readonly OctoLogger _log;
 
-        public GithubClient(OctoLogger log, HttpClient httpClient, VersionChecker versionChecker, string personalAccessToken)
+        public GithubClient(OctoLogger log, HttpClient httpClient, IVersionProvider versionProvider, string personalAccessToken)
         {
             _log = log;
             _httpClient = httpClient;
@@ -25,7 +25,7 @@ namespace OctoshiftCLI
             if (_httpClient != null)
             {
                 _httpClient.DefaultRequestHeaders.Add("Accept", "application/vnd.github.v3+json");
-                _httpClient.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue("OctoshiftCLI", versionChecker?.GetCurrentVersion()));
+                _httpClient.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue("OctoshiftCLI", versionProvider?.GetCurrentVersion()));
                 _httpClient.DefaultRequestHeaders.Add("GraphQL-Features", "import_api,mannequin_claiming");
                 _httpClient.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Bearer", personalAccessToken);
             }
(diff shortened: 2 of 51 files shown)


src/Octoshift/AdoClient.cs
```diff
@@ -15,7 +15,7 @@ public class AdoClient
         private readonly OctoLogger _log;
         private double _retryDelay;
 
-        public AdoClient(OctoLogger log, HttpClient httpClient, string personalAccessToken)
+        public AdoClient(OctoLogger log, HttpClient httpClient, VersionChecker versionChecker, string personalAccessToken)
```
ArinGhazarian, line 18: Instead of introducing a new dependency just to get the assembly's version, we can just create an extension method for `HttpRequestHeaders` and get rid of the unnecessary dependency:
```c#
public static void AddUserAgent(this HttpRequestHeaders httpRequestHeaders)
{
    httpRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue("OctoshiftCLI", Assembly.GetExecutingAssembly().GetName().Version.ToString()));
}
```

src/Octoshift/GithubClient.cs
```diff
@@ -17,15 +17,15 @@ public class GithubClient
         private readonly HttpClient _httpClient;
         private readonly OctoLogger _log;
 
-        public GithubClient(OctoLogger log, HttpClient httpClient, string personalAccessToken)
+        public GithubClient(OctoLogger log, HttpClient httpClient, VersionChecker versionChecker, string personalAccessToken)
```
ArinGhazarian, line 20: same, unnecessary dependency.

src/Octoshift/VersionChecker.cs
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        public string GetCurrentVersion()
+        {
+            return Assembly.GetExecutingAssembly().GetName().Version.ToString();
+        }
+
+        public async Task<string> GetLatestVersion()
+        {
+            if (string.IsNullOrWhiteSpace(_latestVersion))
```
ArinGhazarian, line 30: nit: we have an extension method for this now :) --> `_latestVersion.IsNullOrWhiteSpace()`, it reads a bit nicer nothing more :)

src/Octoshift/VersionChecker.cs
```diff
@@ -0,0 +1,51 @@
...
+            {
+                using var client = new HttpClient();
+                client.DefaultRequestHeaders.Add("Accept", "application/vnd.github.v3+json");
+                client.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue("OctoshiftCLI", GetCurrentVersion()));
```
ArinGhazarian, line 34: This instantiation should go into the `ctor`.  I would also inject it into the `ctor` just like the way we do in out Api classes.

src/ado2gh/Program.cs
```diff
@@ -23,15 +25,40 @@ public static async Task Main(string[] args)
...
+                Logger.LogWarning($"Could not retrieve latest ado2gh version from github.com, please ensure you are using the latest version by downloading it from https://github.com/github/gh-gei/releases/latest");
+                Logger.LogVerbose(ex.ToString());
+            }
+
             await parser.InvokeAsync(args);
         }
 
+        private static async Task LatestVersionCheck()
+        {
+            var versionChecker = new VersionChecker();
```
ArinGhazarian, line 49: It's better to instantiate using the `serviceProvider`, pass it as a parameter the call `GerRequriedService`

src/gei/Program.cs
```diff
@@ -43,12 +46,36 @@ public static async Task Main(string[] args)
...
+            {
+                Logger.LogInformation($"You are running the latest version of the gei CLI [v{await versionChecker.GetLatestVersion()}]");
+            }
+            else
+            {
+                Logger.LogWarning($"You are running an older version of the gei CLI [v{versionChecker.GetCurrentVersion()}]. The latest version is v{await versionChecker.GetLatestVersion()}.");
+                Logger.LogWarning($"Please update by running: gh extension upgrade gei");
+            }
+        }
+
```
ArinGhazarian, line 78: same comments

ArinGhazarian (changes requested):

src/Octoshift/VersionChecker.cs
dylan-smith: I just made this change and I don't like it. I don't like concepts like HTTP headers leaking out, IMO any knowledge about headers and how they are constructed should be encapsulated in this class (just like how we do with the auth header).

ArinGhazarian: That's fine, but I'd say def. get rid of that dependency, it's really unnecessary to be dependent on it just to get the assembly version.

src/Octoshift/AdoClient.cs
dylan-smith: I'm not sure I agree. Having it encapsulated in the VersionChecker centralizes a few things around how we deal with version numbers. For example is it the 4-part version number, or 2 part like in the GH releases? Does the version number include the "v" before it or not? I tried to keep all those details hidden away in VersionChecker so other classes wouldn't need to worry about it.

src/Octoshift/VersionChecker.cs
ArinGhazarian: Also regarding the extension, I see it just as a helper nothing more, the framework itself has lots of extensions like that. But again I am fine either way 😃

dylan-smith: my opinion on this depends on what happens with the other comment thread above about the VersionChecker dependency

src/Octoshift/AdoClient.cs
ArinGhazarian: Ok, but in that case I would create an interface `IVersionProvider` which has that one method in it, the implementation would still be inside the `VersionChecker` but then we would only inject `IVersionProvider` instead of the entire `VersionChecker`. This will help making future refactors, etc. easier since we don't depend on functionality that we are not ever going to need.

src/Octoshift/VersionChecker.cs
dylan-smith: injecting IVersionProvider instead now

src/Octoshift/GithubClient.cs
dylan-smith: changed to IVersionProvider being injected

src/ado2gh/Program.cs
```diff
@@ -23,15 +25,41 @@ public static async Task Main(string[] args)
...
+                Logger.LogWarning($"Could not retrieve latest ado2gh version from github.com, please ensure you are using the latest version by downloading it from https://github.com/github/gh-gei/releases/latest");
+                Logger.LogVerbose(ex.ToString());
+            }
+
             await parser.InvokeAsync(args);
         }
 
+        private static async Task LatestVersionCheck(ServiceProvider sp)
+        {
+            var versionChecker = sp.GetService<VersionChecker>();
```
ArinGhazarian, line 50:
```suggestion
            var versionChecker = sp.GetRequiredService<VersionChecker>();
```
this way it will throw an exception if it wasn't able to resolve the `VersionChecker` for any reason.

src/gei/Program.cs
```diff
@@ -43,12 +47,36 @@ public static async Task Main(string[] args)
...
+                Logger.LogWarning($"Could not retrieve latest gei CLI version from github.com, please ensure you are using the latest version by running: gh extension upgrade gei");
+                Logger.LogVerbose(ex.ToString());
+            }
+
             await parser.InvokeAsync(args);
         }
 
+        private static async Task LatestVersionCheck(ServiceProvider sp)
+        {
+            var versionChecker = sp.GetService<VersionChecker>();
```
ArinGhazarian, line 67:
```suggestion
            var versionChecker = sp.GetRequiredService<VersionChecker>();
```

ArinGhazarian (changes requested):

src/ado2gh/Program.cs
```diff
@@ -23,15 +25,41 @@ public static async Task Main(string[] args)
                 .AddSingleton<AdoApiFactory>()
                 .AddSingleton<GithubApiFactory>()
                 .AddSingleton<RetryPolicy>()
+                .AddSingleton<VersionChecker>()
+                .AddSingleton<IVersionProvider, VersionChecker>()
```
ArinGhazarian, line 29: This will create another singleton instance of `VersionChecker` so we will end up with two which probably not what we want, to fix it we can resolve the singleton instance like this:
```suggestion
                .AddSingleton<IVersionProvider, VersionChecker>(sp => sp.GetRequiredService<VersionChecker>())
```

src/gei/Program.cs
```diff
@@ -26,6 +28,8 @@ public static async Task Main(string[] args)
                 .AddSingleton<IBlobServiceClientFactory, BlobServiceClientFactory>()
                 .AddSingleton<IAzureApiFactory, AzureApiFactory>()
                 .AddSingleton<RetryPolicy>()
+                .AddSingleton<VersionChecker>()
+                .AddSingleton<IVersionProvider, VersionChecker>()
```
ArinGhazarian, line 32:
```suggestion
                .AddSingleton<IVersionProvider, VersionChecker>(sp => sp.GetRequiredService<VersionChecker>())
```

ArinGhazarian (changes requested):

src/Octoshift/VersionChecker.cs
```diff
@@ -0,0 +1,52 @@
...
+            return Assembly.GetExecutingAssembly().GetName().Version.ToString();
+        }
+
+        public async Task<string> GetLatestVersion()
+        {
+            if (_latestVersion.IsNullOrWhiteSpace())
+            {
+                using var client = new HttpClient();
+                client.DefaultRequestHeaders.Add("Accept", "application/vnd.github.v3+json");
+                client.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue("OctoshiftCLI", GetCurrentVersion()));
```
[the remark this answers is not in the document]
dylan-smith: I was avoiding putting this in GithubApi because I think it will make things messy because the target GH for this call is always github.com, even if the source/target github instances are something different, the version check will always go to github.com, and it shouldn't have any PAT token set on this api call.

I'll make the change to inject HttpClient though, that makes sense.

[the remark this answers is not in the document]
dylan-smith: done

PR #341, opened 2022-04-14
Fix generate command to allow GHES migrations

juanmrad:
- [ ] Issue linked

### DIFF
--- a/src/gei/Commands/GenerateScriptCommand.cs
+++ b/src/gei/Commands/GenerateScriptCommand.cs
@@ -122,7 +122,7 @@ namespace OctoshiftCLI.GithubEnterpriseImporter.Commands
           string adoTeamProject,
           string githubTargetOrg,
           FileInfo output,
-          string ghesApiUrl = "",
+          string ghesApiUrl = null,
           string azureStorageConnectionString = "",
           bool noSslVerify = false,
           bool skipReleases = false,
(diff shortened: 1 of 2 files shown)


src/gei/Commands/GenerateScriptCommand.cs
```diff
@@ -212,7 +212,9 @@ public GenerateScriptCommand(OctoLogger log, ISourceGithubApiFactory sourceGithu
 
         private async Task<string> InvokeGithub(string githubSourceOrg, string githubTargetOrg, string ghesApiUrl, string azureStorageConnectionString, bool noSslVerify, bool sequential, string githubSourcePat, bool skipReleases)
         {
-            var repos = await GetGithubRepos(_sourceGithubApiFactory.Create(sourcePersonalAccessToken: githubSourcePat), githubSourceOrg);
+            var repos = string.IsNullOrWhiteSpace(ghesApiUrl) ? 
+                await GetGithubRepos(_sourceGithubApiFactory.Create(sourcePersonalAccessToken: githubSourcePat), githubSourceOrg) :
+                await GetGithubRepos(_sourceGithubApiFactory.Create(apiUrl: ghesApiUrl, sourcePersonalAccessToken: githubSourcePat), githubSourceOrg);
```
roferg, line 217: On line 125 in this file you'll need to set `string ghesApiUrl = ""` to be `string ghesApiUrl = null`. Otherwise the `??=` operand in the create function will pick the empty string. `??=` only checks if the left hand side is null.

roferg (approved):
Left one comment on changing the default ghesApiUrl value from empty string to null.

src/gei/Commands/GenerateScriptCommand.cs
juanmrad: thought of this but wasn't sure if it would break something else lol.

roferg: When I tried it locally it broke things :)

ArinGhazarian:
new PR --> https://github.com/github/gh-gei/pull/342

Issue #216, opened 2022-01-26
Sometimes integrate-boards requires All Orgs scope


dylan-smith:
## Description

Home Point ran into a 401 error when calling integrate-boards. Changing their ADO PAT to have All Orgs scope fixed it.

However, when we tested and documented the scopes, All Orgs was not required. Why do some customers need All Orgs and others don't?

modularity:
This might be from [linked remote work items](https://docs.microsoft.com/en-us/azure/devops/boards/backlogs/add-link?view=azure-devops&tabs=browser#link-to-a-remote-work-item) that span multiple Azure DevOps organizations.

PR #345, opened 2022-04-15
Removing All Orgs scope requirement on PAT for integrate-boards

dylan-smith:
Issue #216 

`integrate-boards` used to require the ADO PAT to have the `All Accessible Organizations` setting, which some customers (e.g. any MSFT org) cannot choose due to security policies.

The reason this was required was the way we were calling the boards integration API's required us to pass in an Org ID, and the API's to retrieve an Org ID required that PAT permission.

Now we have a simplified way to call the boards integration API's that do not require the Org ID to be passed in.

- [x] Issue linked


liang2zhu1:
Took a very rough peek - do you plan to remove the calls to SPS APIs for fetching organization?

dylan-smith:
> Took a very rough peek - do you plan to remove the calls to SPS APIs for fetching organization?

we do still use those in a legit way when you call generate-script without specifying an org, in that case requiring all orgs scope makes sense, because we do actually want to iterate over all the orgs the PAT can access

ArinGhazarian (approved):
💯

PR #348, opened 2022-04-19
Extract base GHES url from GHES api url

ArinGhazarian:
Fixes https://github.com/github/octoshift/issues/4352

### Description
The CLI was calling the `startRepositoryMigration` mutation with a hardcoded base url in `sourceRepositoryUrl`. It works for most scenarios but when migrating from GHES to GitHub, the base url needs to be the actual GHES base url not `github.com`. Functionality wise, since the CLI expects a `--ghes-api-url` to be passed for GHES migrations, the migration was working fine. The only problem was that in the final migration log, instead of pointing to the correct GHES source it was always pointing to `github.com`. 

### Solution
Since we didn't want to introduce yet another command line arg for the base url, we are going to extract the base url from the provided `ghesApiUrl`. We expect the url format to be either `http(s)://hostname/api/v3` or `http(s)://api.hostname.com` so we can extract the base url. In case of failure, we will always fallback to the `ghesApiUrl` and use it as the base url. 

- [x] Issue linked

### DIFF
--- a/src/gei/Commands/MigrateRepoCommand.cs
+++ b/src/gei/Commands/MigrateRepoCommand.cs
@@ -2,6 +2,7 @@
 using System.CommandLine;
 using System.CommandLine.Invocation;
 using System.Linq;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using OctoshiftCLI.Extensions;
 
@@ -18,6 +19,7 @@ namespace OctoshiftCLI.GithubEnterpriseImporter.Commands
         private const int CHECK_STATUS_DELAY_IN_MILLISECONDS = 10000; // 10 seconds
         private const string GIT_ARCHIVE_FILE_NAME = "git_archive.tar.gz";
         private const string METADATA_ARCHIVE_FILE_NAME = "metadata_archive.tar.gz";
+        private const string DEFAULT_GITHUB_BASE_URL = "https://github.com";
 
         public MigrateRepoCommand(OctoLogger log, ISourceGithubApiFactory sourceGithubApiFactory, ITargetGithubApiFactory targetGithubApiFactory, EnvironmentVariableProvider environmentVariableProvider, IAzureApiFactory azureApiFactory) : base("migrate-repo")
         {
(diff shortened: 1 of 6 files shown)


dylan-smith:
That integration test failure looks like it's caught a legit problem

dylan-smith:
I would suggest adding a bullet to release notes

src/gei/Commands/MigrateRepoCommand.cs
```diff
@@ -67,7 +69,7 @@ public MigrateRepoCommand(OctoLogger log, ISourceGithubApiFactory sourceGithubAp
             var ghesApiUrl = new Option<string>("--ghes-api-url")
             {
                 IsRequired = false,
-                Description = "Required if migrating from GHES. The api endpoint for the hostname of your GHES instance. For example: http(s)://api.myghes.com"
+                Description = "Required if migrating from GHES. The API endpoint for the hostname of your GHES instance. For example: http(s)://myghes.com/api/v3"
```
dylan-smith, line 72: more realistic example

```suggestion
                Description = "Required if migrating from GHES. The API endpoint for your GHES instance. For example: http(s)://ghes.contoso.com/api/v3"
```

src/gei/Commands/MigrateRepoCommand.cs
```diff
@@ -243,9 +247,27 @@ private async Task<bool> RepoExists(GithubApi githubApi, string org, string repo
...
 
+        private string ExtractGhesBaseUrl(string ghesApiUrl)
+        {
+            // We expect the GHES url template to be either http(s)://hostname/api/v3 or http(s)://api.hostname.com.
+            // We are either going to be able to extract and return the base url based on the above templates or 
+            // will fallback to ghesApiUrl and return it as the base url. 
+
+            ghesApiUrl = ghesApiUrl.Trim().TrimEnd('/');
+
+            var baseUrl = Regex.Match(ghesApiUrl, @"(?<baseUrl>https?:\/\/.+)\/api\/v3", RegexOptions.IgnoreCase).Groups["baseUrl"].Value;
```
dylan-smith, line 261: does this work for both http and https?

dylan-smith (changes requested):

src/gei/Commands/MigrateRepoCommand.cs
ArinGhazarian: yes the `s?` indicates that the last `s` can appear 0 or 1 times.

dylan-smith (approved):

Issue #354, opened 2022-04-20
Add Azure DevOps Server URL support


roferg:
## Todo
Add support for an Azure DevOps Server customer to pass in their server URL to generate script and migrate repo. When this happens it would override the default assumption of dev.azure.com for all Azure DevOps URLs. 

We'll assume that the customer would add a `--azure-devops-server-url` option followed by the servers hostname. For example: 

```
--azure-devops-server-url https://roganazdo.westus2.cloudapp.azure.com
```

The URL will NOT include the collection name, project name, or repo name. Instead we'll take normally provided options for ADO organization, project, and repo and add them to the URL after the host name. This option will be a hidden option that isn't clearly visible to customers running the CLI help. For this take we we'll assume the customer is using port 443 and not a custom port like 8080.

PR #361, opened 2022-04-25
add ado server url to gei migrate-repo

dylan-smith:
Issue #354 

This adds support to `gh gei generate-script` and `gh gei migrate-repo` for passing in an `--ado-server-url` arg to support migrating from ADO Server.

For now going to leave this option as a hidden option, as such will not be mentioning anything in the release notes, nor updating the docs.

This PR only adds this capability to `gh gei` NOT to `ado2gh`. There are some API calls we make from ado2gh that hit the SPS endpoint (e.g. LockRepo, GetOrganizations), and not sure what that would be when using ADO Server (or if there are even equivalents for all of them).

- [x] Issue linked

### DIFF
--- a/src/Octoshift/AdoApi.cs
+++ b/src/Octoshift/AdoApi.cs
@@ -13,8 +13,13 @@ namespace OctoshiftCLI
     public class AdoApi
     {
         private readonly AdoClient _client;
+        private readonly string _adoBaseUrl;
 
-        public AdoApi(AdoClient client) => _client = client;
+        public AdoApi(AdoClient client, string adoServerUrl)
+        {
+            _client = client;
+            _adoBaseUrl = adoServerUrl?.Trim('/');
+        }
 
         public virtual async Task<string> GetUserId()
         {
@@ -59,7 +64,7 @@ namespace OctoshiftCLI
 
         public virtual async Task<IEnumerable<string>> GetTeamProjects(string org)
         {
-            var url = $"https://dev.azure.com/{org}/_apis/projects?api-version=6.1-preview";
+            var url = $"{_adoBaseUrl}/{org}/_apis/projects?api-version=6.1-preview";
             var data = await _client.GetWithPagingAsync(url);
             return data.Select(x => (string)x["name"]).ToList();
         }
@@ -68,7 +73,7 @@ namespace OctoshiftCLI
 
         public virtual async Task<IEnumerable<(string Id, string Name, bool IsDisabled)>> GetRepos(string org, string teamProject)
         {
-            var url = $"https://dev.azure.com/{org}/{teamProject}/_apis/git/repositories?api-version=6.1-preview.1";
+            var url = $"{_adoBaseUrl}/{org}/{teamProject}/_apis/git/repositories?api-version=6.1-preview.1";
             var data = await _client.GetWithPagingAsync(url);
             return data.Select(x => ((string)x["id"], (string)x["name"], ((string)x["isDisabled"]).ToBool()))
                        .ToList();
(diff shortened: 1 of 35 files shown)


src/Octoshift/AdoApi.cs
```diff
@@ -13,8 +13,13 @@ namespace OctoshiftCLI
     public class AdoApi
     {
         private readonly AdoClient _client;
+        private readonly string _adoBaseUrl;
 
-        public AdoApi(AdoClient client) => _client = client;
+        public AdoApi(AdoClient client, string adoServerUrl)
+        {
+            _client = client;
+            _adoBaseUrl = adoServerUrl?.Trim('/');
```
ArinGhazarian, line 21: nit
```suggestion
            _adoBaseUrl = adoServerUrl?.TrimEnd('/');
```

ArinGhazarian: not sure if it's a good idea but wouldn't it make more sense to name this `adoServerBaseUrl`?

src/OctoshiftCLI.Tests/AdoApiTests.cs
```diff
@@ -35,7 +35,7 @@ public async Task GetUserId_Should_Return_UserId()
 
             mockClient.Setup(x => x.GetAsync(endpoint).Result).Returns(userJson.ToJson());
 
-            var sut = new AdoApi(mockClient.Object);
+            var sut = new AdoApi(mockClient.Object, "https://dev.azure.com");
```
ArinGhazarian, line 38: nit: Could you create a `const` for this?

src/gei/AdoApiFactory.cs
```diff
@@ -17,11 +19,17 @@ public AdoApiFactory(OctoLogger octoLogger, HttpClient client, EnvironmentVariab
             _versionProvider = versionProvider;
         }
 
-        public virtual AdoApi Create(string personalAccessToken = null)
+        public virtual AdoApi Create()
+        {
+            return Create(null, null);
+        }
+
+        public virtual AdoApi Create(string adoServerUrl, string personalAccessToken)
```
dylan-smith, line 27: I deliberately got rid of the optional args. I personally think the overrides is clearer, but also with 2 optional strings its way to easy to do factory.Create(adoPat) not realizing that you actually passed a PAT into the URL field.

ArinGhazarian (approved):
Aside from a few nits, all looks good 💯

src/gei/AdoApiFactory.cs
ArinGhazarian: The only problem with this is as you said it would be impossible to only provide `personalAccessToken` and we need to provide both now. So we will now end up with lots of calls like following which I personally don't like:
```c#
var adoApi = adoApiFactory.Create(null, "PAT");
```

dylan-smith: what do you suggest?

ArinGhazarian: There is one other option too which is to set a default value for `adoServerUrl` but then we need to put the default urls (for both GitHub and Ado) into one place (i.e. inside `Defaults.cs`)

src/Octoshift/AdoApi.cs
dylan-smith: it's also the base URL when hitting ado SERVICE (i.e. https://dev.azure.com)

src/gei/AdoApiFactory.cs
ArinGhazarian: My ultimate suggestion is to leave the default values as is and just make the `Create()` internal since it's only going to be used in unit tests anyway.

ArinGhazarian (approved):
Left a nitpick comment but I am 👍

src/gei/AdoApiFactory.cs
dylan-smith: I made it real simple, 2 args, neither optional. In our code we're pretty much always passing both anyways, and now the test mocks match thee signature of the code under test.

Issue #335, opened 2022-04-07
Bulk reclaiming of mannequins


dylan-smith:
The `reclaim-mannequin` command lets customers reclaim one at a time. We should expand on that command to allow customers to do this in bulk easier. Probably by updating that command to accept a csv mapping file.

tspascoal:
@dylan-smith  That is the plan, accept a CSV and (most likely) add a command to generate the CSV as well

PR #373, opened 2022-04-29
Add --csv option to  reclaim-mannequin command

tspascoal:
Issue #335 

Code has been refactored to extract reclaiming functionality into a service to avoid duplication of command code

CSV header extracted into a constant

There is an inconsistency of semantics between reclaiming with a CSV and by specifying user on the cmd line
 Reclaming without a CSV is done with a login only and not with a login and user id.

- [x] Issue linked


admiralAwkbar:
@tspascoal Dope code!

RELEASENOTES.md
```diff
@@ -0,0 +1 @@
+Add capability to reclaim mannequins in bulk by using a CSV file.
```
dylan-smith, line 1:
```suggestion
- Add capability to reclaim mannequins in bulk by using a CSV file
  - Added new command `generate-mannequin-csv`
  - Updated `reclaim-mannequin` command to accept `--csv` argument
```

src/Octoshift/Mannequins.cs
```diff
@@ -0,0 +1,73 @@
...
+                );
+        }
+
+        /// <summary>
+        /// Checks if the user has been claimed at least once (regardless of the last reclaiming result)
+        /// </summary>
+        /// <param name="login"></param>
+        /// <param name="userid"></param>
+        /// <returns></returns>
+        public bool IsClaimed(string login, string userid)
```
dylan-smith, line 48: is this a userid or a mannequinid?

src/Octoshift/Mannequins.cs
```diff
@@ -0,0 +1,73 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+using OctoshiftCLI.Models;
+
+namespace Octoshift
+{
+    public class Mannequins
```
dylan-smith, line 9: should this be in the Models folder?

src/Octoshift/ReclaimConstants.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Octoshift
+{
+    public static class ReclaimConstants
+    {
+        public const string CSVHEADER = "mannequin-user,mannequin-id,target-user";
```
dylan-smith, line 5: could we move this to ReclaimService and eliminate this class?

src/Octoshift/Models/Mannequin.cs
```diff
@@ -11,5 +11,7 @@ public class Mannequin
         public string Id { get; set; }
         public string Login { get; set; }
         public Claimant MappedUser { get; set; }
+
+        public bool IsUnclaimed() => MappedUser == null;
```
dylan-smith, line 15: this is never used

src/Octoshift/Mannequins.cs
```diff
@@ -0,0 +1,73 @@
...
+    {
+        private readonly Mannequin[] _mannequins;
+
+        public Mannequins(Mannequin[] mannequins)
```
dylan-smith, line 13: do we need this ctor? it's only used in tests

src/Octoshift/Services/ReclaimService.cs
```diff
@@ -0,0 +1,189 @@
+﻿using System;
+using System.Linq;
+using System.Threading.Tasks;
+using Octoshift.Models;
+using OctoshiftCLI;
+using OctoshiftCLI.Models;
+
+namespace Octoshift.Services
```
ArinGhazarian, line 8: Can we get rid of the dedicated `Services` folder. Technically all of our clients are services, so to avoid confusion let's place this alongside others in the root folder.

src/Octoshift/Services/ReclaimService.cs
```diff
@@ -0,0 +1,189 @@
...
+        private readonly GithubApi _githubApi;
+        private readonly OctoLogger _log;
+        private bool _failed;
+
+        public ReclaimService(GithubApi githubApi, OctoLogger logger)
+        {
+            _githubApi = githubApi;
+            _log = logger;
+        }
+        public virtual async Task ReclaimMannequin(string mannequinUser, string mannequinId, string targetUser, string githubOrg, bool force)
```
ArinGhazarian, line 21: nit: missing new line

src/Octoshift/Services/ReclaimService.cs
```diff
@@ -0,0 +1,189 @@
...
+        {
+            var githubOrgId = await _githubApi.GetOrganizationId(githubOrg);
+
+            var mannequins = new Mannequins((await GetMannequins(githubOrgId)).GetByLogin(mannequinUser, mannequinId));
```
ArinGhazarian, line 25: `Mannequins` is a service and needs to be injected. If the intention is to group some helper methods that are only meant to be used in this service then you can put them here as private methods also. But otherwise all services need to be properly injected and registered in the DI container.

src/Octoshift/Services/ReclaimService.cs
```diff
@@ -0,0 +1,189 @@
...
+            {
+                _log.LogError($"Failed to reclaim {mannequinUser} ({mannequin.Id}) to {targetUser} ({targetUserId})");
+                _failed = true;
+                return;
+            }
+
+            _log.LogInformation($"Successfully reclaimed {mannequinUser} ({mannequin.Id}) to {targetUser} ({targetUserId})");
+        }
+
+        private void HandleResult()
```
ArinGhazarian, line 146: what's the purpose of this method? it's weird to have a method just to raise an exception!

src/Octoshift/Mannequins.cs
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        public bool IsClaimed(string login)
+        {
+            return _mannequins.FirstOrDefault(m =>
+                login.Equals(m.Login, StringComparison.OrdinalIgnoreCase)
+                && m.MappedUser != null)?.Login != null;
+        }
+
+        public bool Empty()
```
ArinGhazarian, line 63: naming : `IsEmpty()`

src/Octoshift/Mannequins.cs
```diff
@@ -0,0 +1,73 @@
...
+        {
+            return _mannequins.Length == 0;
+        }
+
+        public IEnumerable<Mannequin> UniqueUsers()
```
ArinGhazarian, line 68: naming: `GetUniqueUsers()`

src/Octoshift/AzureApi.cs
```diff
@@ -49,14 +49,14 @@ private Uri GetServiceSasUriForBlob(BlobClient blobClient)
                 throw new InvalidOperationException("BlobClient object has not been authorized to generate shared key credentials. Verify --azure-storage-connection-key is valid and has proper permissions.");
             }
 
-            var sasBuilder = new BlobSasBuilder()
+            var sasBuilder = new BlobSasBuilder
```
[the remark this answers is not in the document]
tspascoal: `dotnet format` is doing that :(

src/Octoshift/Mannequins.cs
```diff
@@ -0,0 +1,73 @@
...
+    public class Mannequins
+    {
+        private readonly Mannequin[] _mannequins;
+
+        public Mannequins(Mannequin[] mannequins)
+        {
+            _mannequins = mannequins;
+        }
+
+        public Mannequins(IEnumerable<Mannequin> mannequins)
```
ArinGhazarian, line 18: This class is a service and needs to be injected into other services, but this `ctor` makes it impossible to register it against the DI container. So if you really want to keep this class a service then this arg needs to be method arg rather than `ctor` arg.

src/Octoshift/Services/ReclaimService.cs
```diff
@@ -0,0 +1,189 @@
...
+            // get all unique mannequins by login and id and map them all to the same target
+            foreach (var mannequin in mannequins.UniqueUsers())
+            {
+                var result = await _githubApi.ReclaimMannequin(githubOrgId, mannequin.Id, targetUserId);
+
+                HandleResult(mannequinUser, targetUser, mannequin, targetUserId, result);
+            }
+
+            // Fail if there was at least one error
+            HandleResult();
```
ArinGhazarian, line 55: It's more clear to simply check for failure and throw an exception right here instead of using a method.

src/Octoshift/Services/ReclaimService.cs
```diff
@@ -0,0 +1,189 @@
...
+                throw new OctoshiftCliException($"User {mannequinUser} is already mapped to a user. Use the force option if you want to reclaim the mannequin again.");
+            }
+
+            var targetUserId = await _githubApi.GetUserId(targetUser);
+            if (targetUserId == null)
+            {
+                throw new OctoshiftCliException($"Target user {targetUser} not found.");
+            }
+
+            _failed = false;
```
ArinGhazarian, line 44: would it be possible not to set this instance variable? For example `HandleResult` can return a `boolean` instead of `void` so we may be able to create a local variable instead of an instance one.

src/Octoshift/Services/ReclaimService.cs
```diff
@@ -0,0 +1,189 @@
...
+
+        private void HandleResult()
+        {
+            if (_failed)
+            {
+                throw new OctoshiftCliException("Failed to reclaim mannequin(s).");
+            }
+        }
+
+        private (string, string, string) ParseLine(string line)
```
ArinGhazarian, line 154: It'd be way more clear to name the fields:
```suggestion
        private (string Login, string UserId, string ClaimantLogin) ParseLine(string line)
```

ArinGhazarian (changes requested):

src/Octoshift/Mannequins.cs
tspascoal: mannequin.

Renamed to `id` to remove ambiguity.

src/Octoshift/Mannequins.cs
tspascoal: Yeah, I struggled a bit on that one....

Moved to models

ArinGhazarian: I swear GitHub ate my comment 😢  about this. anyways, clearly this class is not a model and is a service since it bundled bunch of helper methods.

src/Octoshift/Models/Mannequin.cs
tspascoal: removed

src/Octoshift/Mannequins.cs
tspascoal: removd. It shouldn't haver survived a refactor

src/Octoshift/Services/ReclaimService.cs
tspascoal: done

src/Octoshift/Mannequins.cs
tspascoal: renamed

src/Octoshift/Mannequins.cs
tspascoal: renamed

src/Octoshift/Services/ReclaimService.cs
tspascoal: done, but used different naming to be more consistent with CSV header

src/Octoshift/Services/ReclaimService.cs
tspascoal: It goes together with the other `HandleResult` overflow.

It encapsulates resulting handling to make the core easier to read

src/Octoshift/Services/ReclaimService.cs
tspascoal: I think it's more of a model, how do you consider it a service?

src/Octoshift/AzureApi.cs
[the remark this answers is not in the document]
dylan-smith: it doesn't do it for me

src/Octoshift/Services/ReclaimService.cs
ArinGhazarian: This is by definition a service b/c it provide bunch of functionality via methods to another service. Whereas a model is an entity with bunch of fields.

ArinGhazarian: Again if you feel like you don't want to inject it then it's a indicator that you need to create bunch of private methods instead of the service class.

admiralAwkbar:
@tspascoal This is awesome thanks for working on this! the people want it... they need it...

src/Octoshift/AzureApi.cs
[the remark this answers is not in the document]
tspascoal: image

src/Octoshift/Services/ReclaimService.cs
tspascoal: Injecting seems more effort than it's worth it.

So either inline everything into reclaim service and make it more complicate or make this an nested class of reclaim service.

ArinGhazarian: How about just moving the methods to the service class as private methods? Creating a nested class seems like an overkill to me but it also works.

src/Octoshift/Services/ReclaimService.cs
tspascoal: fixed

src/Octoshift/Services/ReclaimService.cs
tspascoal: nested class added

src/Octoshift/Services/ReclaimService.cs
tspascoal: removed and now only uses a local variable

src/Octoshift/Models/Mannequins.cs
```diff
@@ -0,0 +1,68 @@
+﻿//using System;
```
ArinGhazarian, line 1: You probably wanted to delete this class I guess 😄

src/Octoshift/ReclaimService.cs
```diff
@@ -0,0 +1,248 @@
...
+
+[assembly: InternalsVisibleTo("OctoshiftCLI.Tests")]
+namespace Octoshift
+{
+    public class ReclaimService
+    {
+        private readonly GithubApi _githubApi;
+        private readonly OctoLogger _log;
+
+        public const string CSVHEADER = "mannequin-user,mannequin-id,target-user";
```
ArinGhazarian, line 18: is this supposed to be accessed outside of this class?

src/Octoshift/ReclaimService.cs
```diff
@@ -0,0 +1,248 @@
...
+
+        internal class Mannequins
```
ArinGhazarian, line 20: why `internal` and not `private`?  if the intention is to mock it in tests then this this needs to be 100% injected and as a separate service.

src/Octoshift/ReclaimService.cs
```diff
@@ -0,0 +1,246 @@
...
+            }
+
+            var failed = false;
+
+            // get all unique mannequins by login and id and map them all to the same target
+            foreach (var mannequin in mannequins.GetUniqueUsers())
+            {
+                var result = await _githubApi.ReclaimMannequin(githubOrgId, mannequin.Id, targetUserId);
+
+                failed |= HandleResult(mannequinUser, targetUser, mannequin, targetUserId, result);
```
ArinGhazarian, line 114: This is not going to be a blocking comment so I'll leave it up to you, but it took me a while to figure out what happens here. It's weird if `HandleResult()` returns `true` it actually means failure! So I would inverse the logic and instead of `failed` would create `success` flag and then if `HandleResult` returned `true` it would mean `success` instead of `failure`. Something like this:
```c#
var success = true;
foreach (...)
{
    ...
    success &= HandleResult(...);
}

if (!success)
{
    throw ...
}
```

ArinGhazarian (approved):
I am 👍  just left a comment but it's totally up to you to change it or not 😃

src/Octoshift/ReclaimService.cs
tspascoal: changed

PR #371, opened 2022-04-27
Add download-logs command to gei and ado2gh

synthead:
Depends on https://github.com/github/octoshift/pull/4521
Depends on https://github.com/github/octoshift/pull/4520
Depends on https://github.com/github/github/pull/219142
Part of https://github.com/github/octoshift/issues/4380

Adds support for downloading migration logs for migrations.

- [x] Did you write/update appropriate tests
- [x] Release notes updated (if appropriate)
- [x] Appropriate logging output
- [x] Issue linked
- [x] Docs updated (or issue created) (https://github.com/github/gh-gei/issues/381)

PR #382, opened 2022-05-17
Add --download-migration-logs option to generate-script command for gei and ado2gh

synthead:
Blocked by ~~https://github.com/github/gh-gei/pull/371~~.
Blocked by ~~https://github.com/github/octoshift/issues/4702~~.
Blocked by ~~https://github.com/github/octoshift/issues/4751~~.
Closes https://github.com/github/octoshift/issues/4380!

This PR adds `--download-migration-logs` to the `generate-script` command.  When this option is issued, it appends `download-logs` commands after the "wait" commands in the generated scripts.

This PR should ship after https://github.com/github/gh-gei/pull/371 ships!  While this PR doesn't depend on any code from it, shipping this before it would give users an option that isn't supported yet.

- [x] Issue linked
- [x] Docs updated (or issue created) (https://github.com/github/gh-gei/issues/383)

## To-do
- [x] Remove `--wait` and make it the default behavior for `download-logs`
- [x] Add a timeout for checking the log URL (~1 minute should be fine)
- [x] Write to the log and console that the logs are only available for 24 hours
- [x] Add HTTP request debug output for `HttpDownloadService`


synthead:
Should we include this option in the integration tests? :thinking: 

https://github.com/github/gh-gei/blob/018b69c94b018dc0a47a45e226d9e6dff620f0c0/src/OctoshiftCLI.IntegrationTests/GithubToGithub.cs#L50-L50

https://github.com/github/gh-gei/blob/018b69c94b018dc0a47a45e226d9e6dff620f0c0/src/OctoshiftCLI.IntegrationTests/AdoToGithub.cs#L63-L63

synthead:
Also, should we add these changes to ado2gh? https://github.com/github/gh-gei/blob/018b69c94b018dc0a47a45e226d9e6dff620f0c0/src/ado2gh/Commands/GenerateScriptCommand.cs

dylan-smith:
> Should we include this option in the integration tests? 🤔
> 
> https://github.com/github/gh-gei/blob/018b69c94b018dc0a47a45e226d9e6dff620f0c0/src/OctoshiftCLI.IntegrationTests/GithubToGithub.cs#L50-L50
> 
> https://github.com/github/gh-gei/blob/018b69c94b018dc0a47a45e226d9e6dff620f0c0/src/OctoshiftCLI.IntegrationTests/AdoToGithub.cs#L63-L63

IMO yes!

dylan-smith:
> Also, should we add these changes to ado2gh? https://github.com/github/gh-gei/blob/018b69c94b018dc0a47a45e226d9e6dff620f0c0/src/ado2gh/Commands/GenerateScriptCommand.cs

Also yes

synthead:
@dylan-smith, sounds good!  I added the option to ado2gh and also added `--download-migration-logs` to the integration tests.

Keep in mind that the integration tests won't pass until https://github.com/github/gh-gei/pull/371 gets merged 😉

src/ado2gh/Commands/GenerateScriptCommand.cs
```diff
@@ -600,6 +619,11 @@ private void AppendLine(StringBuilder sb, string content)
 
         private string WaitForMigrationScript(string repoMigrationKey) => $"./ado2gh wait-for-migration --migration-id $RepoMigrations[\"{repoMigrationKey}\"]";
 
+        private string DownloadMigrationLogScript(string githubOrg, string githubRepo)
+        {
+            return $"./ado2gh download-logs --github-org \"{githubOrg}\" --github-repo \"{githubRepo}\"";
```
dylan-smith, line 624: I think we should but a conditional in here like all the functions above it, where it checks if _generateScriptOptions.DownloadMigrationLogs is set (then we can remove all the other conditionals used elsewhere)

synthead: Added in b1065614fe55041e29a8ae29edfdc420fc22daea!

I left a conditional for the parallel script generation, since it appends some spaces before the command:

https://github.com/github/gh-gei/blob/b1065614fe55041e29a8ae29edfdc420fc22daea/src/ado2gh/Commands/GenerateScriptCommand.cs#L476-L485

synthead:
I slipped a `--wait` option to the download-logs command in this PR since it tackles a race condition related to what's happening in our script.  Happy to move this to another PR if you'd like, but it's small enough that I figured it wouldn't matter too much :slightly_smiling_face: 

I checked our code in Octoshift to see how logs are uploaded, and there is a small gap: the migrations are marked as complete to have a complete log, and then the log file gets uploaded afterwards.  This means that there is a small race condition: it's possible for us to check the status of a migration, see that it is complete, and attempt to download the log before it gets uploaded.

This is where `--wait` comes into play.  Instead of raising an error about the log not existing yet, this option polls the GraphQL endpoint for a new migration URL every ten seconds until it finds a URL to download.  This mitigates this race condition: if we land on it, it'll wait ten seconds and likely download it on the second attempt.  If things are really jammed up in prod, this will let the user download the log whenever it becomes available, too.

In addition, this can be useful to the user, because it can also serve as a "wait for my migration to finish and get the log" without necessarily having to explicitly wait for the migration to finish.  It can also be called in an interactive prompt if users are running migrations by hand without attempting to download multiple times.

However, as of this writing, the integration tests seem to be waiting for that log to populate indefinitely because it appears that uploading the logs to Azure is disabled in prod.  We should probably enable this feature.  It requires a secret to be available to GitHub, but as far as I know, this should be shipped.  Slack thread here: https://github.slack.com/archives/C03BW98SSCA/p1652959517380859

src/OctoshiftCLI.IntegrationTests/TestHelper.cs
```diff
@@ -585,5 +585,15 @@ public async Task AssertBoardsIntegrationConfigured(string adoOrg, string teamPr
...
         }
+
+        public async Task AssertMigrationLogFileExists(string githubOrg, string repo)
+        {
+            _output.WriteLine("Checking that the migration log was downloaded...");
+
+            var migrationLogFile = $"migration-log-{githubOrg}-{repo}.log";
+            var migrationLogExists = await Task.Run(() => File.Exists(migrationLogFile));
+
+            migrationLogExists.Should().Be(true);
```
dylan-smith, line 596: This can be simplified
```suggestion
            File.Exists(migrationLogFile).Should().Be(true);
```

synthead: I used `Task.Run` here because of this warning: 

> This async method lacks 'await' operators and will run synchronously. Consider using the 'await' operator to await non-blocking API calls, or 'await Task.Run(...)' to do CPU-bound work on a background thread.

I added the async `Task.Run` in https://github.com/github/gh-gei/pull/382/commits/ca9e42928c70c02763bed76f6bbc9866e2e73337 in a response to this CI failure:

https://github.com/github/gh-gei/runs/6503637347?check_suite_focus=true

There isn't an async equivalent of `File.Exists` or similar, so I'm wrapping it in a `Task.Run` here :slightly_smiling_face: 

What are your thoughts?

synthead:
I also removed the `isHidden` options from `download-logs` introduced in https://github.com/github/gh-gei/pull/396 with https://github.com/github/gh-gei/pull/382/commits/9f5144bee7d57454939159c1477ed222b0883402 :+1:

src/OctoshiftCLI.IntegrationTests/TestHelper.cs
dylan-smith: When this is closer to ready (when prod is updated and int tests are passing) we can pair on this to fix it. But the error is basically saying that you declared this method as async (AssertMigrationLogFileExists), but you're not doing any async stuff in it.  You can fix that in 2 ways, either do some async stuff (which is what you tried to do), or just don't declare the method as async, which I  think is what you should do.

ArinGhazarian: two things:
1. def. don't need to use `Task.Run` 
2. nit: there is a neater syntax for checking booleans: `Should().BeTrue()`

src/ado2gh/Commands/DownloadLogsCommand.cs
```diff
@@ -60,6 +61,12 @@ HttpDownloadService httpDownloadService
                 Description = "Local file to write migration log to (default: migration-log-ORG-REPO.log)."
             };
 
+            var wait = new Option("--wait")
+            {
+                IsRequired = false,
+                Description = "Waits for log to be ready to download."
+            };
```
ArinGhazarian, line 68: Based on the live feedback it'd be better to get rid of `--wait` and always wait with a timeout.

src/ado2gh/Commands/DownloadLogsCommand.cs
```diff
@@ -127,21 +136,36 @@ HttpDownloadService httpDownloadService
...
             }
 
-            _log.LogInformation($"Downloading log for repository {githubRepo} to {migrationLogFile}...");
-
             var githubApi = _githubApiFactory.Create(githubApiUrl, githubPat);
-            var logUrl = await githubApi.GetMigrationLogUrl(githubOrg, githubRepo);
+            string logUrl;
 
-            if (logUrl == null)
+            while (true)
```
ArinGhazarian, line 142: Instead of a `while (true)` it's be better to do `while (logUrl is null)` (assuming we shouldn't expect to get an empty or whitespace string). And of course you can add a timeout to the predicate too.

src/ado2gh/Commands/DownloadLogsCommand.cs
```diff
@@ -127,21 +136,36 @@ HttpDownloadService httpDownloadService
...
             {
-                throw new OctoshiftCliException($"Migration not found for repository {githubRepo}!");
-            }
+                logUrl = await githubApi.GetMigrationLogUrl(githubOrg, githubRepo);
```
ArinGhazarian, line 144: nit: there is a syntactic sugar to throw an exception if the returned value is null:
```suggestion
                logUrl = await githubApi.GetMigrationLogUrl(githubOrg, githubRepo) ?? throw new OctoshiftCliException($"Migration not found for repository {githubRepo}");
```

src/ado2gh/Commands/DownloadLogsCommand.cs
```diff
@@ -127,21 +136,36 @@ HttpDownloadService httpDownloadService
...
-            {
-                throw new OctoshiftCliException($"Migration found for repository {githubRepo}, but migration log not available yet!");
+                if (logUrl.HasValue())
+                {
+                    break;
+                }
+
+                if (logUrl == null)
+                {
+                    throw new OctoshiftCliException($"Migration not found for repository {githubRepo}!");
```
ArinGhazarian, line 153: nit: IMO the `!` at the end of the error message might be confusing (might be considered as part of the repo name) so I'd just drop it.

src/OctoshiftCLI.Tests/gei/Commands/DownloadLogsCommandTests.cs
```diff
@@ -155,6 +157,62 @@ public async Task Calls_Download_With_Expected_Migration_Log_File()
             mockHttpDownloadService.Verify(m => m.Download(It.IsAny<string>(), migrationLogFile));
         }
 
+        [Fact]
+        public async Task Waits_For_URL_To_Populate_When_Wait_Requested()
```
ArinGhazarian, line 161: nit: Use `Url` instead of `URL`. More than two letter identifiers should be `Pascal` cases see [this](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions)

src/OctoshiftCLI.Tests/ado2gh/Commands/DownloadLogsCommandTests.cs
```diff
@@ -155,6 +157,62 @@ public async Task Calls_Download_With_Expected_Migration_Log_File()
             mockHttpDownloadService.Verify(m => m.Download(It.IsAny<string>(), migrationLogFile));
         }
 
+        [Fact]
+        public async Task Waits_For_URL_To_Populate_When_Wait_Requested()
```
ArinGhazarian, line 161: nit: `Url`

src/OctoshiftCLI.Tests/ado2gh/Commands/DownloadLogsCommandTests.cs
```diff
@@ -155,6 +157,62 @@ public async Task Calls_Download_With_Expected_Migration_Log_File()
...
+        {
+            // Arrange
+            var githubOrg = "FooOrg";
+            var repo = "foo-repo";
+            var logUrlEmpty = "";
+            var logUrlPopulated = "some-url";
+            var defaultFileName = $"migration-log-{githubOrg}-{repo}.log";
```
ArinGhazarian, line 168: would you make these `const`s too. Also regarding `defaultFilename` since it's an interpolated string, not sure if CodeQL supports it yet, C# def. supports making an interpolated string `const` (if possible) now but CodeQL is a bit behind the latest C# lang version so it might complain. I guess you can make the change and see if it does or not 😃

src/ado2gh/Commands/DownloadLogsCommand.cs
synthead: Added in ed1afe023dff7250306495b522a6a64db736c407!

src/OctoshiftCLI.Tests/gei/Commands/DownloadLogsCommandTests.cs
synthead: Updated with 5cd1977f6e1d19eff4bffb6c2b3e8a879eb5ab60!

src/OctoshiftCLI.Tests/ado2gh/Commands/DownloadLogsCommandTests.cs
synthead: Updated with 5cd1977f6e1d19eff4bffb6c2b3e8a879eb5ab60!

src/OctoshiftCLI.Tests/ado2gh/Commands/DownloadLogsCommandTests.cs
synthead: Updated in 93b135ba09be640054c604d89a7591bf7e40f673!

src/ado2gh/Commands/DownloadLogsCommand.cs
synthead: I updated the grammar instead in 7e402a2af69542e9241d83b4076787b2339d5f2f :slightly_smiling_face: How's this look?

https://github.com/github/gh-gei/blob/7e402a2af69542e9241d83b4076787b2339d5f2f/src/gei/Commands/DownloadLogsCommand.cs#L141-L141

src/ado2gh/Commands/DownloadLogsCommand.cs
synthead: I replaced the `while(true)` with a for loop in 20b4a78bf1a10bbec8e1cb1e1bbf2ff801b7d7c4.  How's this look?

src/ado2gh/Commands/DownloadLogsCommand.cs
synthead: Updated in 1a2da58aa27d4397aa73600386cb23f500637225!

synthead:
Okay @ArinGhazarian and @dylan-smith, I have addressed everyone's concerns, and this is ready for another round of reviews!

Please note that the integration tests will not pass unless the `UPLOAD_MIGRATION_LOGS_ENABLED` feature flag env var is set to `true` in production, and we should definitely ensure that this is enabled before shipping this feature.

src/OctoshiftCLI.IntegrationTests/AdoToGithub.cs
```diff
@@ -60,7 +60,7 @@ public async Task Basic()
             commitId = await _helper.InitializeAdoRepo(adoOrg, teamProject2, adoRepo2);
             await _helper.CreatePipeline(adoOrg, teamProject2, adoRepo2, pipeline2, commitId);
 
-            await _helper.RunAdoToGithubCliMigration($"generate-script --github-org {githubOrg} --ado-org {adoOrg} --all");
```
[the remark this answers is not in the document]
synthead: Good catch!  Updated with 7c776012c768784c0e587cc1a011b69c9d50b67b!

src/ado2gh/Commands/DownloadLogsCommand.cs
```diff
@@ -95,8 +98,9 @@ HttpDownloadService httpDownloadService
         {
             _log.Verbose = verbose;
 
-            _log.LogInformation($"Downloading logs for organization {githubOrg}...");
+            _log.LogWarning("Migration logs are only available for 24 hours after a migration finishes!");
 
+            _log.LogInformation($"Downloading logs for organization {githubOrg}...");
```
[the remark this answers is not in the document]
synthead: Removed the org name in the log output with 21cb385a659220c5e7caa31611c66e7dffbf15aa!

src/gei/Commands/GenerateScriptCommand.cs
```diff
@@ -188,6 +195,10 @@ public async Task Invoke(GenerateScriptCommandArgs args)
             {
                 _log.LogInformation("SKIP RELEASES: true");
             }
+            if (args.DownloadMigrationLogs)
+            {
+                _log.LogInformation("DOWNLOAD MIGRATION LOGS: true");
```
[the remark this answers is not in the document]
synthead: Ah you're right!  Added in 9f6e3cd2d6f856faac179554d488bffe48cfce37!

src/ado2gh/Commands/DownloadLogsCommand.cs
```diff
@@ -11,11 +11,15 @@ namespace OctoshiftCLI.AdoToGithub.Commands
...
     {
+        internal const int WaitIntervalInSeconds = 10;
+        internal const int RetryMaxCount = 6;
+
         private readonly OctoLogger _log;
         private readonly GithubApiFactory _githubApiFactory;
         private readonly HttpDownloadService _httpDownloadService;
 
         internal Func<string, bool> FileExists = path => File.Exists(path);
+        internal Func<int, Task> Delay = milliSeconds => Task.Delay(milliSeconds);
```
ArinGhazarian: Why do we need to abstract out the `Delay`? For testing we can set `WaitIntervalInSeconds` and `RetryMaxCount`

src/ado2gh/Commands/DownloadLogsCommand.cs
```diff
@@ -127,24 +131,29 @@ HttpDownloadService httpDownloadService
                 _log.LogWarning($"Overwriting {migrationLogFile} due to --overwrite option.");
             }
 
-            _log.LogInformation($"Downloading log for repository {githubRepo} to {migrationLogFile}...");
-
             var githubApi = _githubApiFactory.Create(githubApiUrl, githubPat);
-            var logUrl = await githubApi.GetMigrationLogUrl(githubOrg, githubRepo);
 
-            if (logUrl == null)
+            for (var attempt = 1; attempt <= RetryMaxCount; attempt++)
```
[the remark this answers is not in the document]
ArinGhazarian: I agree, Polly is a better approach here.

src/ado2gh/Commands/DownloadLogsCommand.cs
ArinGhazarian: Also as a [best practice](https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#prefer-asyncawait-over-directly-returning-task) the delegate has to be async and the `Task.Delay` should be awaited:
```c#
internal Func<int, Task> Delay = async milliSeconds => await Task.Delay(milliSeconds);
```

src/ado2gh/Commands/DownloadLogsCommand.cs
```diff
@@ -127,24 +131,29 @@ HttpDownloadService httpDownloadService
...
+                    _log.LogInformation($"Downloading log for repository {githubRepo} to {migrationLogFile}...");
+                    await _httpDownloadService.Download(logUrl, migrationLogFile);
 
-            await _httpDownloadService.Download(logUrl, migrationLogFile);
+                    _log.LogSuccess($"Downloaded {githubRepo} log to {migrationLogFile}.");
+                    break;
+                }
 
-            _log.LogSuccess($"Downloaded {githubRepo} log to {migrationLogFile}.");
+                if (attempt == RetryMaxCount)
```
ArinGhazarian: nit: i'd prefer `if (attempt >= RetryMaxCount` just in case

dylan-smith:
I just noticed hte Polly changes (and the test refactoring) was only done in the ado2gh command, need to make the same changes to the gei command

dylan-smith (approved):
Looks good to me! :shipit:

PR #386, opened 2022-05-18
fix: Prefer DateTime.UtcNow instead of DateTime.Now

nelsonprsousa:
- [X] Did you write/update appropriate tests
- [X] Release notes updated (if appropriate)
- [X] Appropriate logging output
- [X] Issue linked
- [X] Docs updated (or issue created)

### Explanation / Reason behind this PR
We should prefer the use of `DateTime.UtcNow`  instead of `DateTime.Now`. Even tho I prefer UtcNow vs Now, I still do not recommend the use of this struct across the codebase. As a better solution, e.g., a wrapper (like an interface IDateTimeProvider) might be added as a singleton service. That way we could properly unit test our codebase. WDYT?

### Further Details
While I was studying this solution, I did some improvements that you may want to benefit from. If it doesn't add value to your project feel free to close the PR 👍

### DIFF
--- a/src/Octoshift/OctoLogger.cs
+++ b/src/Octoshift/OctoLogger.cs
@@ -29,7 +29,7 @@ namespace OctoshiftCLI
 
         public OctoLogger()
         {
-            var logStartTime = DateTime.Now;
+            var logStartTime = DateTime.UtcNow;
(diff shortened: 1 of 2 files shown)


dylan-smith:
I think we prefer the logs to remain using the local time. I can see arguments for either option, but I think we'll leave it as is for now.

PR #390, opened 2022-05-18
fix: Take advantage of asynchronous code on LockRepoCommand

nelsonprsousa:
- [X] Did you write/update appropriate tests
- [X] Release notes updated (if appropriate)
- [X] Appropriate logging output
- [X] Issue linked
- [X] Docs updated (or issue created)

### Explanation / Reason behind this PR
We can take advantage of asynchronous programming and, since getting team project id and repo id are idempotent, we can paralyze those calls.

### Further Details
While I was studying this solution, I did some improvements that you may want to benefit from. If it doesn't add value to your project feel free to close the PR 👍

### DIFF
--- a/src/ado2gh/Commands/LockRepoCommand.cs
+++ b/src/ado2gh/Commands/LockRepoCommand.cs
@@ -64,8 +64,12 @@ namespace OctoshiftCLI.AdoToGithub.Commands
 
             var ado = _adoApiFactory.Create(adoPat);
 
-            var teamProjectId = await ado.GetTeamProjectId(adoOrg, adoTeamProject);
-            var repoId = await ado.GetRepoId(adoOrg, adoTeamProject, adoRepo);
+            var teamProjectIdTask = ado.GetTeamProjectId(adoOrg, adoTeamProject);
+            var repoIdTask = ado.GetRepoId(adoOrg, adoTeamProject, adoRepo);
+
+            var teamProjectId = await teamProjectIdTask;
+            var repoId = await repoIdTask;
+
             var identityDescriptor = await ado.GetIdentityDescriptor(adoOrg, teamProjectId, "Project Valid Users");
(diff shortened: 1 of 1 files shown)


src/ado2gh/Commands/LockRepoCommand.cs
```diff
@@ -64,8 +64,12 @@ public async Task Invoke(string adoOrg, string adoTeamProject, string adoRepo, s
...
             var ado = _adoApiFactory.Create(adoPat);
 
-            var teamProjectId = await ado.GetTeamProjectId(adoOrg, adoTeamProject);
-            var repoId = await ado.GetRepoId(adoOrg, adoTeamProject, adoRepo);
+            var teamProjectIdTask = ado.GetTeamProjectId(adoOrg, adoTeamProject);
+            var repoIdTask = ado.GetRepoId(adoOrg, adoTeamProject, adoRepo);
+
+            var teamProjectId = await teamProjectIdTask;
+            var repoId = await repoIdTask;
+
```
ArinGhazarian, line 72: This code is going to exactly run the same as previous code. If you want to make them run in parallel you should use `Task.WhenAll()`:
```C#
await Task.WhenAll(new[] {  teamProjectIdTask, teamProjectIdTask };
```
Now that being said, I don't think we need to micro-optimize this code unless we have a very strong reason to do so. Making things unnecessarily parallel usually will makes the code base harder to maintain.

ArinGhazarian (changes requested):

ArinGhazarian (approved):
Please disregard my previous comment (I deleted it already). My only two cents is that to me this is a bit of a micro-optimization but I don't think it is going to be a big deal.

nelsonprsousa:
> Please disregard my previous comment (I deleted it already). My only two cents is that to me this is a bit of a micro-optimization but I don't think it is going to be a big deal.

No worries! If you prefer to keep the execution sequential, feel free to close this PR 👍

PR #421, opened 2022-06-01
Dylan smith/enrich repos csv

dylan-smith:
- [ ] Issue linked

### DIFF
--- a/src/Octoshift/AdoApi.cs
+++ b/src/Octoshift/AdoApi.cs
@@ -57,6 +57,17 @@ namespace OctoshiftCLI
             return $"{ownerName} ({ownerEmail})";
         }
 
+        public Task GetLastPushDate(string org, string teamProject, string repo)
+        {
+            var url = $"{_adoBaseUrl}/{org}/{teamProject}/_apis/git/repositories/{repo}/pushes?$top=1&api-version=7.1-preview.2";
+            var response = await _client.GetAsync(url);
+
+            var data = JObject.Parse(response);
+            var pushDate = (string)data["value"].First()["date"];
+
+            return DateTime.
+        }
+
         public virtual async Task<int> GetPullRequestCount(string org, string teamProject, string repo)
         {
             var url = $"{_adoBaseUrl}/{org}/{teamProject}/_apis/git/repositories/{repo}/pullrequests?searchCriteria.status=all&api-version=7.1-preview.1";
--- a/src/Octoshift/AdoApi.cs
+++ b/src/Octoshift/AdoApi.cs
@@ -57,7 +57,7 @@ namespace OctoshiftCLI
             return $"{ownerName} ({ownerEmail})";
         }
 
-        public Task GetLastPushDate(string org, string teamProject, string repo)
+        public async Task<DateTime> GetLastPushDate(string org, string teamProject, string repo)
         {
             var url = $"{_adoBaseUrl}/{org}/{teamProject}/_apis/git/repositories/{repo}/pushes?$top=1&api-version=7.1-preview.2";
             var response = await _client.GetAsync(url);
@@ -65,7 +65,7 @@ namespace OctoshiftCLI
             var data = JObject.Parse(response);
             var pushDate = (string)data["value"].First()["date"];
 
-            return DateTime.
+            return DateTime.Parse(pushDate);
         }
 
         public virtual async Task<int> GetPullRequestCount(string org, string teamProject, string repo)
--- a/src/ado2gh/Services/ReposCsvGeneratorService.cs
+++ b/src/ado2gh/Services/ReposCsvGeneratorService.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
 
--- a/src/ado2gh/Services/ReposCsvGeneratorService.cs
+++ b/src/ado2gh/Services/ReposCsvGeneratorService.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Collections.Generic;
 using System.Linq;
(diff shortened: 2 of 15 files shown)


src/Octoshift/AdoClient.cs
```diff
@@ -133,6 +148,44 @@ public virtual async Task<JArray> GetWithPagingAsync(string url, string continua
             return data;
         }
 
+        public virtual async Task<IEnumerable<T>> GetWithPagingTopSkipAsync<T>(string url, Func<JToken, T> selector) => await GetWithPagingTopSkipAsync(url, 0, selector);
+
+        public virtual async Task<IEnumerable<T>> GetWithPagingTopSkipAsync<T>(string url, int skip, Func<JToken, T> selector)
+        {
+            if (string.IsNullOrWhiteSpace(url))
```
ArinGhazarian, line 155: nit: We do have an extension method for this
```suggestion
            if (url.IsNullOrWhiteSpace())
```

src/Octoshift/AdoClient.cs
```diff
@@ -133,6 +148,44 @@ public virtual async Task<JArray> GetWithPagingAsync(string url, string continua
             return data;
         }
 
+        public virtual async Task<IEnumerable<T>> GetWithPagingTopSkipAsync<T>(string url, Func<JToken, T> selector) => await GetWithPagingTopSkipAsync(url, 0, selector);
```
ArinGhazarian, line 151: Is this basically same as `GetWithPagingAsync` only with an additional `selector`? If so it'd be better to make this an overload of it.

src/ado2gh/Services/ReposCsvGeneratorService.cs
```diff
@@ -32,13 +32,29 @@ public virtual async Task<string> Generate(string adoPat)
...
-                        result.AppendLine($"\"{org}\",\"{teamProject}\",\"{repo}\",\"{url}\",{pipelineCount},{prCount}");
+                        result.AppendLine($"\"{org}\",\"{teamProject}\",\"{repo}\",\"{url}\",{pipelineCount},{prCount},\"{lastPushDate:dd-MMM-yyyy hh:mm tt}\",{commitsPastYear},\"{mostActiveContributor}\"");
                     }
                 }
             }
 
             return result.ToString();
         }
+
+        private async Task<string> GetMostActiveContributor(string org, string teamProject, string repo, AdoApi adoApi)
```
ArinGhazarian, line 47: Since you have already done for the `Generate` method, here we should also pass in `adoPat` and create the `AdoApi` using the factory:
```c#
private async Task<string> GetMostActiveContributor(string org, string teamProject, string repo, string adoPat)
{
    var adoApi = _adoApiFactory.Create(adoPat);
    ...
}
```

src/ado2gh/Services/ReposCsvGeneratorService.cs
```diff
@@ -32,13 +32,29 @@ public virtual async Task<string> Generate(string adoPat)
...
+        {
+            var pushers = await adoApi.GetPushersSince(org, teamProject, repo, DateTime.Today.AddYears(-1));
+            pushers = pushers.Where(x => !x.Contains("Service"));
```
ArinGhazarian, line 50: need to add `System.Linq`

ArinGhazarian (changes requested):

src/Octoshift/AdoClient.cs
dylan-smith: it is not. I've learned that some ADO API's do paging differently than others. The previous GetWithPagingAsync uses a continuation token returned as a HTTP header to get the next page. This uses the top/skip query params to perform paging. Depending on which ADO endpoint you are calling you will need to use one or the other paging method.

src/ado2gh/Services/ReposCsvGeneratorService.cs
dylan-smith: huh? this code compiles as-is

ArinGhazarian: Yeah I also wondered why! I noticed it since CodeQL was complaining about it.

src/ado2gh/Services/ReposCsvGeneratorService.cs
dylan-smith: this is a private method though. Do you still feel the same way even for private methods?

ArinGhazarian: Oh! I didn't notice it was a `private` method. I def. don't feel that same way for private method 😄

src/Octoshift/AdoApi.cs
```diff
@@ -57,6 +57,31 @@ public virtual async Task<string> GetOrgOwner(string org)
             return $"{ownerName} ({ownerEmail})";
         }
 
+        public virtual async Task<DateTime> GetLastPushDate(string org, string teamProject, string repo)
+        {
+            var url = $"{_adoBaseUrl}/{org}/{teamProject}/_apis/git/repositories/{repo}/pushes?$top=1&api-version=7.1-preview.2";
+            var response = await _client.GetAsync(url);
+
+            var data = JObject.Parse(response);
+            var pushDate = data.ContainsKey("value") && data["value"].Any() ? (string)data["value"].First()["date"] : DateTime.MinValue.ToString();
```
ArinGhazarian, line 66: CodeQL's complain about this is legit (although in our case can be considered micro-optimization), it can be rewritten like the following just to make CodeQL happy 😄 :
```suggestion
            var pushDate = data.TryGetValue("value", out var values) && values.Any() ? (string)values.First()["date"] : DateTime.MinValue.ToString();
```

ArinGhazarian (approved):

PR #423, opened 2022-06-01
Optimize repo exists check

### LINKED ISSUE #378: Repo Already Exists check is inefficient/slow
When we check if a repo already exists in the target during `migrate-repo` we are retrieving the list of all repos in the target org (and paging them). If it's a large list this can be slow. We can probably do this check in a more performant way without retrieving the list of all repos.

ArinGhazarian:
Closes #378 

Our repo exists logic was not efficient since it was using `GithubApi#GetRepos` to find out if a repo exists. Since the GitHub's REST api already has a [get repo](https://docs.github.com/en/rest/repos/repos#get-a-repository) endpoint, we just need to call it and if it returns `200` status code it means that the repo exists and if it returns `404` it means that it doesn't exist. 

_NOTE:_ There is a caveat here that is if the PAT is not SSO enabled for and org the get repo call will return 404, so we need to make sure that the PAT is SSO enabled for the org.   

- [x] Issue linked

### DIFF
--- a/src/Octoshift/GithubApi.cs
+++ b/src/Octoshift/GithubApi.cs
@@ -97,6 +97,21 @@ namespace OctoshiftCLI
             return await _client.GetAllAsync(url).Select(x => (string)x["name"]).ToListAsync();
         }
 
+        public virtual async Task<bool> RepoExists(string org, string repo)
+        {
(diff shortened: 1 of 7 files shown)


dylan-smith (approved):
:shipit:

PR #428, opened 2022-06-04
Add validation states to migration state logic

### LINKED ISSUE #427: Migrate Repo should not treat unexpected statuses as success
Currently `migrate-repo` looks for the statuses: IN_PROGRESS, QUEUED, and FAILED. If it sees any other status it assumes it is success.

This is not very robust logic (e.g. if the backend introduces a new status like VALIDATION_FAILED). We should specifically look for the SUCCESS status, and if it's a status we don't recognize we should error.

tambling:
Closes #427 

- [x] Issue linked


src/Octoshift/RepositoryMigrationStatus.cs
```diff
@@ -6,5 +8,29 @@ public static class RepositoryMigrationStatus
...
+            var failedStates = new string[]
+            {
+                Failed,
+                FailedValidation
+            };
+
+            return Array.Exists(failedStates, e => e == migrationState.Trim()
+                                                     .ToUpper());
+
+        }
```
dylan-smith, line 34:
```suggestion
        public static bool IsPending(string migrationState) => migrationState?.Trim().ToUpper() is Queued or InProgress or PendingValidation;

        public static bool IsFailed(string migrationState) => migrationState?.Trim().ToUpper() is Failed or FailedValidation;
```

dylan-smith:
we should also handle unrecognized statuses as failures. And will need to apply the same changes to ado2gh and gei.

Will need to add some unit tests for this new behavior also.

src/Octoshift/RepositoryMigrationStatus.cs
```diff
@@ -1,3 +1,5 @@
+using System;
```
ArinGhazarian, line 1: No need for this using directive anymore 😃

ArinGhazarian (commented):
The `WaitForMigrationCommand` in both `ado2gh` and `gei` needs to be updated with these changes otherwise the in case of `FAILED_VALIDATION` the wait command will go on forever.

src/Octoshift/RepositoryMigrationStatus.cs
```diff
@@ -6,5 +6,11 @@ public static class RepositoryMigrationStatus
         public const string InProgress = "IN_PROGRESS";
         public const string Failed = "FAILED";
         public const string Succeeded = "SUCCEEDED";
+        public const string PendingValidation = "PENDING_VALIDATION";
+        public const string FailedValidation = "FAILED_VALIDATION";
+
+        public static bool IsSucceeded(string migrationState) => migrationState?.Trim().ToUpper() is Succeeded;
+        public static bool IsPending(string migrationState) => migrationState?.Trim().ToUpper() is Queued or InProgress or PendingValidation;
+        public static bool IsFailed(string migrationState) => !(IsPending(migrationState) || IsSucceeded(migrationState));
```
ArinGhazarian, line 14: Not a huge deal but I guess it'd be better to add unit tests for these methods too, like the following:
```c#
public void IsPending_Returns_True_When_Status_Is_Queued() { ... }

public void IsPending_Returns_False_When_Statues_Is_Not_Queued_Or_InProgress_Or_PendindValidation() { ... }
```

ArinGhazarian (approved):
Overall looks good, just a small non blocking suggestion for adding more unit tests but I will leave the decision up to you 😃

tambling:
@ArinGhazarian/@dylan-smith, does this change merit release notes/doc updates? And once this merges and releases, can we rely on customers to upgrade, and ship the `PENDING_VALIDATION`/`FAILED_VALIDATION` change to Octoshift sooner than later?

ArinGhazarian:
> does this change merit release notes/doc updates?

I am not quite sure but leaning toward no, since this is not a change that a user can see, I mean they would run the exact same command(s) but the command behaves differently now. 

> And once this merges and releases, can we rely on customers to upgrade, and ship the PENDING_VALIDATION/FAILED_VALIDATION change to Octoshift sooner than later?

We absolutely cannot! If you check out our Splunk dashboard you can see that people are using different versions and unfortunately not always the latest version.

dylan-smith:
I don't think this warrants any release notes or doc changes.

But we def can't ship the backend changes yet. We'll need to monitor customer adoption of the new CLI version and wait until we hit some threshold of % users that have this version or newer?

@roferg what do you think is a good target? 70% over a week before we ship the backend change?

Good news is that now that we have the version nag message, our customers are much better about keeping up to date with cli versions.

tambling:
Alright, sounds good! I'm going to fix the merge conflicts on this (they might break tests, so I might need to re-request reviews), then toss the back-end changes in the "Blocked" column until we hit a TBD adoption threshold.

roferg:
> @roferg what do you think is a good target? 70% over a week before we ship the backend change?

I think 70% is fine. We can warn support as well once it's live in case any tickets are filed. We really want the validation logic as it should help prevent the large volume of customer PAT misconfigurations.

src/ado2gh/Commands/WaitForMigrationCommand.cs
```diff
@@ -60,6 +60,7 @@ public async Task Invoke(string migrationId, string githubPat = null, bool verbo
                 switch (state)
                 {
                     case RepositoryMigrationStatus.Failed:
+                    case RepositoryMigrationStatus.FailedValidation:
```
dylan-smith, line 63: Can we update this to use the IsFailed/IsPending methods, and while we're at it we should explicitly check for IsPending instead of using default like we are here. Default should throw an exception (NOT an OctoshiftCliException) saying it's an unexpected state.

src/gei/Commands/MigrateRepoCommand.cs
```diff
@@ -221,14 +221,14 @@ public async Task Invoke(MigrateRepoCommandArgs args)
 
             var (migrationState, _, failureReason) = await githubApi.GetMigration(migrationId);
 
-            while (migrationState.Trim().ToUpper() is "IN_PROGRESS" or "QUEUED")
```
dylan-smith, line 224: We should do this in ado2gh MigrateRepoCommand also. And can we explicitly check for IsSucceeded (and throw a non-OctoshiftCliException if it's an unexpected state).

src/gei/Commands/WaitForMigrationCommand.cs
```diff
@@ -60,6 +60,7 @@ public async Task Invoke(string migrationId, string githubTargetPat = null, bool
                 switch (state)
                 {
                     case RepositoryMigrationStatus.Failed:
+                    case RepositoryMigrationStatus.FailedValidation:
```
dylan-smith, line 63: can we use IsFailed() here? And same comment about checking IsPending explicitly and failing on unexpected state values.

dylan-smith (changes requested):

src/ado2gh/Commands/WaitForMigrationCommand.cs
tambling: Since `IsSucceeded` is true for succeeded states, `IsPending` is true for pending states, and `IsFailed` is true for both failed and unexpected states, wouldn't the state always trip one of those three? Or does your suggestion posit a future where maybe we make `IsFailed` a bit stricter?

dylan-smith: My bad, I didn't look close enough at the implementation of IsFailed

tambling:
@dylan-smith Did a pass on using the `RepositoryMigrationStatus` methods for everything — pretty basic, just swapped out the `case` for an `if..else if`, then combined terms accordingly. Let me know if there's anything else I can polish!

dylan-smith:
> @dylan-smith Did a pass on using the `RepositoryMigrationStatus` methods for everything — pretty basic, just swapped out the `case` for an `if..else if`, then combined terms accordingly. Let me know if there's anything else I can polish!

I think the only outstanding comment is to do the same thing in ado2gh MigrateRepoCommand

ArinGhazarian (approved):
🥇

dylan-smith (approved):

Issue #432, opened 2022-06-07
Add .projections.json to support alternate files


synthead:
If we add a `.projections.json` file, we can quickly and easily switch between functional code and its tests.  This file and its contents are well-supported, and works in many editors:

- VS Code: https://marketplace.visualstudio.com/items?itemName=will-wow.vscode-alternate-file
- Vim: https://github.com/tpope/vim-projectionist

Here's an example of this configuration at work with VS Code:

![vscode-alternate-file](https://user-images.githubusercontent.com/820984/172441798-8e6ebe42-c00c-4e32-92d2-abd35ef73255.gif)

PR #433, opened 2022-06-07
Add .projections.json to support alternate files

synthead:
Closes https://github.com/github/gh-gei/issues/432!

This PR adds `.projections.json` to configure alternate files.  This allows us to easily switch between functional code and their tests.

From the the related issue:

> If we add a `.projections.json` file, we can quickly and easily switch between functional code and its tests.  This file and its contents are well-supported, and works in many editors:
> 
> - VS Code: https://marketplace.visualstudio.com/items?itemName=will-wow.vscode-alternate-file
> - Vim: https://github.com/tpope/vim-projectionist
> 
> Here's an example of this configuration at work with VS Code:
> 
> vscode-alternate-file

- [ ] ~~Did you write/update appropriate tests~~ (n/a)
- [ ] ~~Release notes updated (if appropriate)~~ (n/a)
- [ ] ~~Appropriate logging output~~ (n/a)
- [x] Issue linked
- [ ] ~~Docs updated (or issue created)~~ (n/a)

### DIFF
--- a/.projections.json
+++ b/.projections.json
@@ -1,0 +1,25 @@
+{
+  "src/Octoshift/*.cs": {
+    "alternate": "src/OctoshiftCLI.Tests/{}Tests.cs"
+  },
+
+  "src/OctoshiftCLI.Tests/*Tests.cs": {
+    "alternate": "src/Octoshift/{}.cs"
+  },
+
+  "src/gei/*.cs": {
+    "alternate": "src/OctoshiftCLI.Tests/gei/{}Tests.cs"
+  },
+
+  "src/OctoshiftCLI.Tests/gei/*Tests.cs": {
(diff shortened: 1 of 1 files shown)


dylan-smith:
> While I guess it probably won't hurt but not sure if we need to add these plug in configs to the solution. If people want to they install and configure the plug-in. IMO all we need to add is the mainstream IDE configs (VS, VSCode and Rider) the rest is kinda up to ppl. Not sure what @dylan-smith thinks though :D

I think this is a pretty niche functionality. I don't expect many people to use it. But I don't see any harm in adding it either. If we end up with a bunch of noisy random config files in the root of the repo, we can always come back and prune some later. I don't think it's a problem yet.

dylan-smith (approved):

PR #449, opened 2022-06-14
Add INT test for GHES to GHEC migration path

### LINKED ISSUE #353: Integration Test with GHES
We should create at least 1 integration test that performs a GHES -> GHEC migration.

ArinGhazarian:
Closes #353 

This PR adds an integration test to cover GHES to GHEC migration path. For the GHES migration path to work we needed to store GHES' PAT in GitHub but in order to not interfere with GHEC to GHEC migration path, we renamed the `GH_SOURCE_PAT` secret key to `GHES_PAT` and consequently did the same for `GH_PAT` and renamed it to `GHEC_PAT`. Then in the test code we do the mapping and properly map the PATs and create a dictionary of env variables to be used by the process that eventually runs the migration.

- [x] Issue linked

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -159,8 +159,8 @@ jobs:
     - name: Integration Test
       env: 
         ADO_PAT: ${{ secrets.ADO_PAT }}
-        GH_PAT: ${{ secrets.GH_PAT }}
-        GH_SOURCE_PAT: ${{ secrets.GH_SOURCE_PAT }}
+        GHEC_PAT: ${{ secrets.GHEC_PAT }}
+        GHES_PAT: ${{ secrets.GHES_PAT }}
         AZURE_STORAGE_CONNECTION_STRING_LINUX: ${{ secrets.AZURE_STORAGE_CONNECTION_STRING_LINUX }}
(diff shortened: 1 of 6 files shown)


.github/workflows/CI.yml
```diff
@@ -160,6 +160,10 @@ jobs:
       env: 
         ADO_PAT: ${{ secrets.ADO_PAT }}
         GH_PAT: ${{ secrets.GH_PAT }}
+        GH_SOURCE_PAT: ${{ secrets.GH_SOURCE_PAT }}
```
dylan-smith, line 163: isn't this going to cause the GHEC -> GHEC integration test to use this as the source PAT also?

ArinGhazarian: Oops! you're right, I totally missed it. Maybe the way to go here is to name it differently like `GHES_PAT` and then pass it to the CLI as an arg in the test.

dylan-smith: that sounds like it would work

dylan-smith: That might be messy to make work. You probably need to put some code in the test to override/reset the envvars properly for the test scenario.

dylan-smith (approved):

PR #454, opened 2022-06-16
Delete migration logs before kicking off a new migration

ArinGhazarian:
For our INT tests we need to delete the migration logs before kicking off a new migration. Additionally it fails the INT test if there is any error in CLI logs. 

- [x] Issue linked

### DIFF
--- a/src/OctoshiftCLI.IntegrationTests/TestHelper.cs
+++ b/src/OctoshiftCLI.IntegrationTests/TestHelper.cs
@@ -471,7 +471,7 @@ steps:
         public async Task RunGeiCliMigration(string generateScriptCommand, IDictionary<string, string> tokens) =>
             await RunCliMigration($"gei {generateScriptCommand}", "gh", tokens);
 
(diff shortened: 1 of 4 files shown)


src/OctoshiftCLI.IntegrationTests/TestHelper.cs
```diff
@@ -86,6 +86,9 @@ public async Task ResetGithubTestEnvironment(string githubOrg)
 
             foreach (var repo in githubRepos)
             {
+                _output.WriteLine($"Deleting migration log for repo: {githubOrg}\\{repo}");
+                DeleteMigrationLog(githubOrg, repo);
```
ArinGhazarian, line 90: Even though we only need to delete the logs for the target GitHub org, I thought for our INT tests it'd be just easier to shove this here, it won't do anything for the source GitHub org since it wouldn't find any logs.

dylan-smith (approved):

PR #463, opened 2022-06-22
made --ado-pipeline able to work with just name

### LINKED ISSUE #462: Make `--ado-pipeline` arg in `rewire-pipeline` accept either name or path
In #413 we changed the behavior of the `--ado-pipeline` arg to accept a pipeline path rather than a name. We also updated `generate-script` to put the path values rather than the name values in the generated script. However, if a customer generated a script with a previous version of the CLI, then tried to run it with the new CLI, this could have been a breaking change for them.

We may be able to reduce the impact by updating `rewire-pipeline` to see if it can find a pipeline with a matching name if a path isn't passed in. If it finds more than one pipeline with a matching name (in different pipeline folders), it will need to error out. But if there's only one pipeline with a matching name we should be able to proceed.

dylan-smith:
Closes #462 

When we shipped the change so that `--ado-pipeline` accepted a pipeline path instead of just a name, we broke some customers that already had scripts generated using the name.

This change makes `rewire-pipeline` more resilient, so that if you pass a name and there is only one pipeline that matches that name, then we can still proceed.  If there are multiple pipelines that match that name (in different pipeline folders), then we will still fail and the user will need to provide the full path.

- [x] Issue linked

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -1,1 +1,1 @@
-
+- `--ado-pipeline` arg in the `ado2gh rewire-pipeline` command will now accept a pipeline name without the full pipeline path, so long as there is only one pipeline found that matches that name. If there are multiple pipelines with the same name (in different pipeline folders), you will need to provide the full pipeline path.


src/Octoshift/AdoApi.cs
```diff
@@ -431,7 +431,11 @@ public virtual async Task<int> GetPipelineId(string org, string teamProject, str
                 }
             }
 
-            return _pipelineIds[(org.ToUpper(), teamProject.ToUpper(), pipelinePath.ToUpper())];
+            return _pipelineIds.TryGetValue((org.ToUpper(), teamProject.ToUpper(), pipelinePath.ToUpper()), out result)
```
ArinGhazarian, line 434: Maybe I am missing something here but I think the `TryGetValue` here is redundant since if we reach this far it means that we have already done the `TryGetValue` earlier in line 413

dylan-smith: if line 413 had a value we would have returned on line 415. We need to TryGet again because we could have added a new value on line 426.

dylan-smith: The login in this method is basically 1. Do we have the ID in the cache? 2. Lets get all the values from the API, and lookup by path 3. Didn't find it? lets look up by name only 4. Still didn't find it, or found more than one? error

ArinGhazarian (approved):
💯

PR #467, opened 2022-06-22
Handle GQL error response

### LINKED ISSUE #464: Better error handling if startRepositoryMigration fails
Recently startRepositoryMigration endpoint implemented some pre-migration validation. If the validation fails (e.g. workflow scope missing), the API call to startRepositoryMigration will not return a migration ID and will instead return an error message (but still an HTTP 200).

Right now this will cause the CLI to crash, because when it gets a 200 response it expects to find a migration ID.

It would be better to look for the error message in the response, and raise an OctoshiftCliException.

You can see an example of this failure here: https://github.com/github/gh-gei/runs/6995914753?check_suite_focus=true#step:17:168

ArinGhazarian:
Fixes #464 

- [x] Issue linked

### Screenshot 
image

### DIFF
--- a/src/Octoshift/GithubApi.cs
+++ b/src/Octoshift/GithubApi.cs
@@ -619,7 +619,7 @@ namespace OctoshiftCLI
                                     : null
             };
         }
-        
+
         private void EnsureSuccessGraphQLResponse(JObject response)
         {
             if (response.ContainsKey("errors") && response["errors"] is JArray { Count: > 0 } errors)
--- a/src/Octoshift/GithubApi.cs
+++ b/src/Octoshift/GithubApi.cs
@@ -278,6 +278,8 @@ namespace OctoshiftCLI
             var response = await _client.PostAsync(url, payload);
             var data = JObject.Parse(response);
 
+            EnsureSuccessGraphQLResponse(data);
+
             return (string)data["data"]["startRepositoryMigration"]["repositoryMigration"]["id"];
         }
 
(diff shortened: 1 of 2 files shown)


dylan-smith:
should this PR be non-draft now?

dylan-smith:
Could you attach a screenshot of calling migrate-repo with  a PAT without workflow scope

src/Octoshift/GithubApi.cs
```diff
@@ -617,5 +619,16 @@ private static Mannequin BuildMannequin(JToken mannequin)
...
         }
+
+        private void EnsureSuccessGraphQLResponse(JObject response)
+        {
+            if (response.TryGetValue("errors", out var jErrors) && jErrors is JArray { Count: > 0 } errors)
+            {
+                var error = (JObject)errors[0];
+                var errorType = error.TryGetValue("type", out var jType) ? (string)jType : null;
+                var errorMessage = error.TryGetValue("message", out var jMessage) ? (string)jMessage : null;
+                throw new OctoshiftCliException($"GraphQL response does not indicate success. Error Type: {errorType ?? "UNKNOWN"}, Error Message: {errorMessage ?? "UNKNOWN"}");
```
dylan-smith, line 630: IMO we should make the error simpler. Probably just `new OctoshiftCliException(errorMessage ?? "UNKNOWN")`  If users want all the gory details they can look in the verbose logs.

dylan-smith (changes requested):

src/Octoshift/GithubApi.cs
ArinGhazarian: addressed in c3064ce7d63ca11b181d2ab77bde4376b04f2472

dylan-smith (approved):

PR #470, opened 2022-06-27
GithubApi#RepoExists returns false on 301

### LINKED ISSUE #469: GEI CLI reports 301 and won't queue a migration for a recently renamed repo
## Description

The GEI CLI checks if a repo name is already in use before calling the GEI backend to queue a migration. If you rename a repo instead of deleting it then you end up with a 301 response when trying to queue a migration. Which stops the CLI from moving forward. 

## Reproduction Steps

1. Migrate a repo. For example, migrate a repo into a new repo called `foo`. 
2. Rename that repo after the migration completes. For example, rename `foo` to `foo-1`. 
3. Try to queue the migration again with the original repo name as the target. For example, try to migrate into `foo` again. 
4. Note that the CLI errors out on the 301 response from GitHub.

ArinGhazarian:
Fixes #469 

When a repo gets renamed, the GitHub API returns `301 (Moved)` instead of `404 (Not Found)`. This PR fixed it so on both `404` and `301` responses the `GithubApi#RepoExists()` returns `false`. 

- [x] Issue linked

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -1,3 +1,3 @@
 - `generate-script` command in `ado2gh` and `gh gei` will no longer generate an empty script if no migratable repos were found.
 - `--ado-pipeline` arg in the `ado2gh rewire-pipeline` command will now accept a pipeline name without the full pipeline path, so long as there is only one pipeline found that matches that name. If there are multiple pipelines with the same name (in different pipeline folders), you will need to provide the full pipeline path.
-- Fixed a bug where the `migrate-repo` command would return `301` for a recently renamed repo. 
+- Fixed a bug where the `migrate-repo` command would return an error when migrating to a recently renamed GitHub repo which the old/original name was used as the target repo.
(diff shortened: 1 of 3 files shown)


RELEASENOTES.md
```diff
@@ -1,2 +1,3 @@
 - `generate-script` command in `ado2gh` and `gh gei` will no longer generate an empty script if no migratable repos were found.
 - `--ado-pipeline` arg in the `ado2gh rewire-pipeline` command will now accept a pipeline name without the full pipeline path, so long as there is only one pipeline found that matches that name. If there are multiple pipelines with the same name (in different pipeline folders), you will need to provide the full pipeline path.
+- Fixed a bug where the `migrate-repo` command would return `301` for a recently renamed repo.
```
dylan-smith, line 3: not sure exactly the wording I would use, but I would be less specific about `301` codes. Just that it used to error, and now it doesn't.

ArinGhazarian: addressed in cf9038cb5c8dde33e99996ccc5c494e1406381cf

Issue #481, opened 2022-07-01
`generate-script` commands does not include source-pat / target-pat in outputed script file


madkoo:
## Description
When using the `generate-script` command and providing the" --github-source-pat" argument then that value is not included in the outputted PowerShell script file. The generated commands in the file don't have tokens in the command so it is required to add those as env variables. To simplify the script running after generated I have included the "github-source-pat" in the generated output. 
Also, I added the option to provide the" --github-target-pat" for the `generate-script` command so that would be part of it as well because the migrate-repo command itself allows it to be used.  
This gives options if you want to provide the token inline or env variables.

## Reproduction Steps
1. run generate-script command with --github-source-pat argument 
2. the generated script does not contain the provided argument in the generated migrate-repo commands, which requires modifying the files or adding the pat as an env variable.

dylan-smith:
This was an intentional design choice to do it this way. In general putting secrets in environment variables is a more secure practice than as command-line args. We added the ability to specify them as command-line args for convenience, but we deliberately did not want to persist secrets into the generated script file. If a customer uses generate-script it is expected that they will need to set the secrets as env vars before running the script.

PR #482, opened 2022-07-01
Use source pat in genscript

madkoo:
- [x] Issue linked

This is a fix for issue #481


ArinGhazarian:
👋 @madkoo, 
This was not a bug and is by design. We thought about whether to include the PATs in the generated script and decided not to do it because it can be a potential security risk to directly include the PATs in the script. Customers can of course always edit the generated script and include the PATs manually.

madkoo:
@ArinGhazarian  thanks for feeback, I was assuming that might be the case, why it was like that :)

PR #495, opened 2022-07-14
sanitize pr number

dylan-smith:
Sanitizes the PR Number in publish-test-results.yml workflow. Uses `sed` to strip any non numeric characters from the string.

- [x] Issue linked

### DIFF
--- a/.github/workflows/publish-test-results.yml
+++ b/.github/workflows/publish-test-results.yml
@@ -38,7 +38,7 @@ jobs:
           eventjson=`cat 'artifacts/Event File/event.json'`
           prnumber=`echo $(jq -r '.pull_request.number' <<< "$eventjson")`
           echo $prnumber
-          echo "PR_NUMBER=$(echo $prnumber | sed 's/[*0-9]*//g')" >> $GITHUB_ENV
+          echo "PR_NUMBER=$(echo $prnumber | sed 's/[^0-9]*//g')" >> $GITHUB_ENV
       
       - name: Publish Unit Test Results
         uses: EnricoMi/publish-unit-test-result-action@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-test-results.yml
```diff
@@ -37,10 +37,8 @@ jobs:
         run: |
           eventjson=`cat 'artifacts/Event File/event.json'`
           prnumber=`echo $(jq -r '.pull_request.number' <<< "$eventjson")`
-          prorg=`echo $(jq -r '.pull_request.head.repo.owner.login' <<< "$eventjson")`
           echo $prnumber
-          echo "PR_NUMBER=$(echo $prnumber)" >> $GITHUB_ENV
-          echo "PR_ORG=$(echo $prnumber)" >> $GITHUB_ENV
+          echo "PR_NUMBER=$(echo $prnumber | sed 's/[^0-9]*//g')" >> $GITHUB_ENV
```
anticomputer, line 41: One concern I have here is that this since sed is line processing, it does not strip trailing newlines here and there's still room for newline injection shenanigans. i.e. you can inject arbitrary sequences of newline separated decimal sequences. While that should be fairly useless to an attacker, we still don't want to introduce unexpected data I reckon.

If we don't want to do an explicit check for non-numeric prior to actually passing to the echo invocation with something like (or some equivalent that checks the contents prior to passing the untrusted data to a shell callout):

```
if [[ "$prnumber" =~ [^0-9] ]]; then echo "Invalid PR number ... aborting"; return 1; else return 0; fi
```

Then would suggest we at least chop of any newlines explicitly by piping through `tr -d '\n'` as well, e.g:

```suggestion
          echo "PR_NUMBER=$(echo $prnumber | sed 's/[^0-9]*//g' | tr -d '\n')" >> $GITHUB_ENV
```

anticomputer (approved):

ArinGhazarian (approved):

Issue #540, opened 2022-08-02
Support for sha mapping file to update PRs during migration


robandpdx:
## Description

Before migration from GHES to GHEC, a customer with large files in a repo must first migrate those large files to LFS. This rewrites the git repo history, affecting PRs. The PR shas will be outdated. I'm told that with `git lfs migrate` you can output a mapping of the old shas to the new ones. And if using ECI, a user can manipulate the pull_request json files in the tarball to rewrite the previous shas with the new ones.

Many migrations into GitHub cloud require LFS migration. GEI should support a smooth way for customers to provide this mapping file generated from `git lfs migrate` in order to update the PRs during migration.

dylan-smith:
This would be a good feature to consider if/when we add LFS support to the CLI. Tracked in #440 

I don't have an ETA when we might tackle LFS support, but not this quarter.

PR #545, opened 2022-08-10
Add --lfs-mapping-file cli option

robandpdx:
This change will address #540. This change allows a user to pass in a sha mapping file produced during lfs migration. Users can add the --lfs-mapping-file <filename> option to specify the mapping file. During the migration, the mapping file shall be used to update the PR SHAs according to the mapping.
- [x] Issue linked


src/gei/Commands/MigrateRepoCommand.cs
```diff
@@ -132,6 +135,11 @@ public MigrateRepoCommand(OctoLogger log, ISourceGithubApiFactory sourceGithubAp
             {
                 IsRequired = false
             };
+            var lfsMappingFile = new Option<string>("--lfs-mapping-file")
+            {
+                IsRequired = false,
+                Description = "Rewrite PR SHAs using lfs mapping file during migration."
```
dylan-smith, line 141:
```suggestion
                Description = "Rewrite PR SHAs using LFS mapping file during migration (only supported when migrating from GHES)."
```

robandpdx: This works when migrating from GHEC also.

dylan-smith: But there is no archive file when migrating from GHEC? how does that work?

dylan-smith: Or at least there's no archive file that the CLI has access to AFAIK

robandpdx: It works the same as when migrating from GHES. The [metadataArchiveContent is downloaded from GHES or GHEC](https://github.com/robandpdx/gh-gei/blob/robandpdx/src/gei/Commands/MigrateRepoCommand.cs#L322) as a byte[], then I [extract the tar.gz contents to a local directory](https://github.com/robandpdx/gh-gei/blob/robandpdx/src/gei/Commands/MigrateRepoCommand.cs#L483-L486). I then [modify each file extracted as needed](https://github.com/robandpdx/gh-gei/blob/robandpdx/src/gei/Commands/MigrateRepoCommand.cs#L489-L500). Lastly, I create a new [.tar.gz file from the modified files, read that new .tar.gz file into a byte[]](https://github.com/robandpdx/gh-gei/blob/robandpdx/src/gei/Commands/MigrateRepoCommand.cs#L504-L521) and return it to [metadataArchiveContent](https://github.com/robandpdx/gh-gei/blob/robandpdx/src/gei/Commands/MigrateRepoCommand.cs#L327)

robandpdx: I plan to do something similar with the [gitArchiveContent](https://github.com/robandpdx/gh-gei/blob/robandpdx/src/gei/Commands/MigrateRepoCommand.cs#L316) to implement an option to convert the repo to LFS during the migration process.

src/gei/Commands/MigrateRepoCommand.cs
```diff
@@ -454,6 +472,54 @@ private void LogAndValidateOptions(MigrateRepoCommandArgs args)
...
+                _log.LogInformation($"LFS MAPPING FILE: {args.LfsMappingFile}");
+            }
+        }
+
+        private byte[] applyLfsMappingFileToMetadata(byte[] metadataArchiveContent, string lfsMappingFile)
+        {
+            // extract the metadata archive
+            var gzipDataStream = new MemoryStream(metadataArchiveContent);
+            Stream inStream = new GZipInputStream(gzipDataStream);
+            TarArchive archive = TarArchive.CreateInputTarArchive(inStream, TarBuffer.DefaultBlockFactor);
```
ArinGhazarian, line 486: 1. All of these classes are `IDisposable` so they should be properly disposed. In C# you can do it with `using`: 
```c#
using var gZipData = new MemoryStream(metadataArchiveContent);
```
2. Prefer `var variable` over `Type varialbe` --> `var inStream = ...`
3. Generally speaking we should avoid Hungarian notation --> use `gZipData`  instead of `gZipDataStream`.

We follow these guidelines for naming conventions (with some exceptions, for example we use all caps for constants): 
1. [Microsoft C# naming guidelines](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/naming-guidelines)
2. [The most widely used guideline (non Microsoft)](https://github.com/ktaranov/naming-convention/blob/master/C%23%20Coding%20Standards%20and%20Naming%20Conventions.md)

src/gei/Commands/MigrateRepoCommand.cs
```diff
@@ -454,6 +472,54 @@ private void LogAndValidateOptions(MigrateRepoCommandArgs args)
...
+            {
+                string text = File.ReadAllText(fileName);
+                var lfsMappingLines = File.ReadLines(lfsMappingFile);
+                foreach (var lfsMappingLine in lfsMappingLines)
+                {
+                    string[] lfsMapping = lfsMappingLine.Split(','); 
+                    text = text.Replace(lfsMapping[0], lfsMapping[1]);
+                }
+                File.WriteAllText(fileName, text);
+            }
```
ArinGhazarian, line 501: 1. This is a bit inefficient because you traverse the entire file for each LFS mapping where as it'd be better to go through each file once and perform the mapping.
2. For each file in the extracted archive directory you load the LFS mapping file and do the split logic! It's be way more optimized to read the LFS mapping file once and create a `Dictionary` (aka hash map) and then for each file, traverse it line by line and do the mapping.

src/gei/Commands/MigrateRepoCommand.cs
```diff
@@ -454,6 +472,54 @@ private void LogAndValidateOptions(MigrateRepoCommandArgs args)
...
+            newArchive.RootPath = "./archiveExtracted";
+            
+            if (fileNames.Length > 0) {
+                foreach (string name in fileNames) {
+                    TarEntry entry = TarEntry.CreateEntryFromFile(name);
+                    newArchive.WriteEntry(entry, true);
+                }
+            } else {
+                _log.LogInformation("No files found in archiveExtracted directory");
+            }
```
ArinGhazarian, line 517: This can be simplified:
```suggestion
            if (!fileNames.Any()) 
            {
                _log.LogInformation("No files found in archiveExtracted directory");
            }
            
            foreach (string name in fileNames) 
            {
                    TarEntry entry = TarEntry.CreateEntryFromFile(name);
                    newArchive.WriteEntry(entry, true);
            }
```

src/gei/Commands/MigrateRepoCommand.cs
```diff
@@ -454,6 +472,54 @@ private void LogAndValidateOptions(MigrateRepoCommandArgs args)
...
+            newArchive.Close();
```
ArinGhazarian, line 518: I believe the `Close` method would be automatically called upon dispose.

ArinGhazarian (changes requested):

src/gei/Commands/MigrateRepoCommand.cs
```diff
@@ -454,6 +472,54 @@ private void LogAndValidateOptions(MigrateRepoCommandArgs args)
...
                 _log.LogInformation($"METADATA ARCHIVE URL: {args.MetadataArchiveUrl}");
             }
+            if (args.LfsMappingFile != "")
+            {
+                _log.LogInformation($"LFS MAPPING FILE: {args.LfsMappingFile}");
+            }
+        }
+
+        private byte[] applyLfsMappingFileToMetadata(byte[] metadataArchiveContent, string lfsMappingFile)
+        {
```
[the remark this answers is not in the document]
robandpdx: @ArinGhazarian This all sounds like solid advice. I wasn't super stoked about checking in `metadataArchiveOrig.tar.gz` so I'm glad you can steer this in a better direction. I'm new to c# so I'll need to lean on you and your team to help get this over the finish line. Thanks for all the help here!

src/gei/Commands/MigrateRepoCommand.cs
```diff
@@ -453,6 +471,54 @@ private void LogAndValidateOptions(MigrateRepoCommandArgs args)
                 _log.LogInformation($"GIT ARCHIVE URL: {args.GitArchiveUrl}");
                 _log.LogInformation($"METADATA ARCHIVE URL: {args.MetadataArchiveUrl}");
             }
+            if (args.LfsMappingFile != "")
+            {
+                _log.LogInformation($"LFS MAPPING FILE: {args.LfsMappingFile}");
+            }
```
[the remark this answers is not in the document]
robandpdx: This is in `LogAndValidateOptions`.

src/gei/Commands/MigrateRepoCommand.cs
robandpdx: Some of this is Greek (or Hungarian 😂 ) to me.

roferg:
@dylan-smith @ArinGhazarian @robandpdx - What are the next steps to get this PR completed?

robandpdx:
@roferg What is needed to get this merged?

roferg:
> @roferg What is needed to get this merged?

@robandpdx - I'm actually rolling off being the migrations PM and onto another project. @timrogers will be the new PM for migrations moving forward. You should work with him and @dylan-smith to get this over the finish line. Note, @dylan-smith is OOO this week and will be back next week.

dylan-smith:
Closing this because it's super stale. If the team decides to pick it back up we can always re-open.

PR #564, opened 2022-08-12
added cache to version checker

### LINKED ISSUE #555: Cache version check results
When running a bunch of CLI commands (e.g. in a script) you'll start to see warnings about the version check failing. This is because the version check makes an anonymous API call, and anonymous api calls get throttled very agressively.

When we have a successful version check we could stuff the date/time that it was done in an environment variable, then subsequent CLI commands could check the env var before making the API call.

dylan-smith:
Using environment variables to cache the results of the version check. Will store the latest version found, and the last time the check was made against github.com. When the CLI runs it will use the env vars to see if a check has been made in the past 24 hours, and if so use the cached latest version number for the check. If no env vars are found, of if the cached date is older than 24 hours it will make the version check API call and update the env vars.

Closes #555 

- [x] Issue linked

### DIFF
--- a/src/Octoshift/VersionChecker.cs
+++ b/src/Octoshift/VersionChecker.cs
@@ -1,4 +1,5 @@
-﻿using System.Net.Http;
+﻿using System;
+using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Reflection;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


src/Octoshift/VersionChecker.cs
```diff
@@ -23,7 +24,24 @@ public VersionChecker(HttpClient httpClient, OctoLogger log)
...
+                if (cacheDate.AddDays(1) > DateTime.Now)
+                {
+                    _log.LogVerbose("Version check cache is valid");
+                    latestVersion = Environment.GetEnvironmentVariable("OCTOSHIFT_VERSION_CHECK_RESULT");
+                }
+            }
+            else
+            {
+                latestVersion = await GetLatestVersion();
+                Environment.SetEnvironmentVariable("OCTOSHIFT_VERSION_CHECK_DATE", DateTime.Now.ToString());
```
ArinGhazarian, line 42: I believe this will only work in Windows. It won't be able to physically create an env variable in macOS and Linux!

PR #591, opened 2022-08-22
CLI support for Org to Org

teja-nanduri:
Fixes: https://github.com/github/octoshift/issues/5339

This PR adds Org to Org migration to CLI. This work is to only enable org to org migrations on GHEC there will be a follow up batch and work for GHES related work.

- [x] Issue linked


src/gei/Commands/MigrateOrgCommand.cs
```diff
@@ -0,0 +1,137 @@
...
+            _environmentVariableProvider = environmentVariableProvider;
+
+            Description = "Invokes the GitHub APIs to migrate the org, its teams and the repositories.";
+
+            var githubSourceOrg = new Option<string>("--github-source-org")
+            {
+                IsRequired = true,
+                Description = "Uses GH_SOURCE_PAT env variable or --github-source-pat option. Will fall back to GH_PAT or --github-target-pat if not set."
+            };
+            var githubTargetOrg = new Option<string>("--github-target-org")
```
juanmrad, line 31: We need the target enterprise also.

src/gei/Commands/MigrateOrgCommand.cs
```diff
@@ -0,0 +1,137 @@
...
+                throw new ArgumentNullException(nameof(args));
+            }
+
+            _log.Verbose = args.Verbose;
+
+            LogAndValidateOptions(args);
+
+            var githubApi = _targetGithubApiFactory.Create(targetPersonalAccessToken: args.GithubTargetPat);
+
+            var githubOrgId = await githubApi.GetOrganizationId(args.GithubTargetOrg);
```
juanmrad, line 79: We do not need the target org id as it won't exist we need the enterprise id instead.

juanmrad (changes requested):

src/Octoshift/GithubApi.cs
```diff
@@ -143,6 +143,23 @@ public virtual async Task<string> GetOrganizationId(string org)
             return (string)data["data"]["organization"]["id"];
         }
 
+        public virtual async Task<string> GetEnterpriseId(string enterpriseName)
+        {
+            var url = $"{_apiUrl}/graphql";
+
+            var payload = new
+            {
+                // TODO: this is super ugly, need to find a graphql library to make this code nicer
```
dylan-smith, line 152:
```suggestion
```

dylan-smith: I think you copied this comment from somewhere else in the code. I think at this point we've accepted that the GQL calls are as good as they're gonna get for now.

src/gei/Commands/MigrateOrgCommand.cs
```diff
@@ -0,0 +1,143 @@
...
+            {
+                IsRequired = true,
+                Description = "Name of the target enterprise."
+            };
+            var ssh = new Option("--ssh")
+            {
+                IsRequired = false,
+                IsHidden = true,
+                Description = "Uses SSH protocol instead of HTTPS to push a Git repository into the target repository on GitHub."
+            };
```
dylan-smith, line 44: I would remove this, this is long since obsolete.

src/gei/Commands/MigrateOrgCommand.cs
```diff
@@ -0,0 +1,143 @@
...
+            var targetToken = args.GithubTargetPat ?? _environmentVariableProvider.TargetGithubPersonalAccessToken();
+
+            var migrationId = await githubApi.StartOrganizationMigration(
+                sourceOrgUrl,
+                args.GithubTargetOrg,
+                githubEnterpriseId,
+                sourceToken,
+                targetToken);
+
+            _log.LogSuccess($"Org Migration has been initiated (ID: {migrationId}).");
```
dylan-smith, line 95: Are we not going to have the CLI poll for status and give the user some indication of success/failure and/or progress?

src/gei/Commands/MigrateOrgCommand.cs
```diff
@@ -0,0 +1,143 @@
...
+            _log.LogInformation($"GITHUB SOURCE ORG: {args.GithubSourceOrg}");
+            _log.LogInformation($"GITHUB TARGET ORG: {args.GithubTargetOrg}");
+            _log.LogInformation($"GITHUB TARGET ENTERPRISE: {args.GithubTargetEnterprise}");
+
+            if (args.Ssh)
+            {
+                _log.LogWarning("SSH mode is no longer supported. --ssh flag will be ignored");
+            }
+
+            if (args.GithubSourcePat is not null)
```
dylan-smith, line 115:
```suggestion
            if (args.GithubSourcePat.HasValue())
```

src/gei/Commands/MigrateOrgCommand.cs
```diff
@@ -0,0 +1,143 @@
...
+            {
+                _log.LogInformation("GITHUB SOURCE PAT: ***");
+            }
+
+            if (args.GithubTargetPat is not null)
```
dylan-smith, line 120:
```suggestion
            if (args.GithubTargetPat.HasValue())
```

src/gei/Commands/MigrateOrgCommand.cs
```diff
@@ -0,0 +1,143 @@
...
+            {
+                _log.LogInformation("GITHUB TARGET PAT: ***");
+
+                if (args.GithubSourcePat is null)
```
dylan-smith, line 124:
```suggestion
                if (args.GithubSourcePat.IsNullOrWhitespace())
```

src/gei/Commands/MigrateOrgCommand.cs
roferg: We need to implement some kind of --wait option for org migrations. As well as a making sure the watch migration command works when give the migration ID. 

Customers will want to monitor the status.

dylan-smith:
Since this isn't usable by customers yet, I would set IsHidden on the command itself so customers don't see it when they do `--help`

src/gei/Commands/MigrateOrgCommand.cs
dylan-smith: I suggest waiting should be the default behavior of migrate-org. Unlike migrate-repo customers won't be generating a migration script when migrating whole orgs (at least I don't think so).

There could be some optional flag to not wait (e.g. --no-wait)

teja-nanduri: I have added the `--wait` option and it's behaviour is similar to the `--wait` option in `migrate-repo` command.

src/Octoshift/GithubApi.cs
```diff
@@ -360,7 +359,23 @@ public virtual async Task<string> StartOrganizationMigration(string sourceOrgUrl
...
+        {
+            var url = $"{_apiUrl}/graphql";
+
+            var query = "query($id: ID!)";
+            var gql = "node(id: $id) { ... on OrganizationMigration { state } }";
+
+            var payload = new { query = $"{query} {{ {gql} }}", variables = new { id = migrationId } };
+
+            var response = await _retryPolicy.HttpRetry(async () => await _client.PostAsync(url, payload),
+                ex => ex.StatusCode == HttpStatusCode.BadGateway);
```
dylan-smith, line 373:
```suggestion
                _ => true);
```
we're getting more aggressive with our retry logic lately, I would change this to retry on any Http error.

src/Octoshift/OrganizationMigrationStatus.cs
```diff
@@ -0,0 +1,18 @@
...
+        public const string InProgress = "IN_PROGRESS";
+        public const string Failed = "FAILED";
+        public const string Succeeded = "SUCCEEDED";
+        public const string NotStarted = "NOT_STARTED";
+        public const string PostRepoMigration = "POST_REPO_MIGRATION";
+        public const string PreRepoMigration = "PRE_REPO_MIGRATION";
+        public const string RepoMigration = "REPO_MIGRATION";
+
+        public static bool IsSucceeded(string migrationState) => migrationState?.Trim().ToUpper() is Succeeded;
+        public static bool IsFailed(string migrationState) => migrationState?.Trim().ToUpper() is Failed;
```
dylan-smith, line 15: Can we flip the logic a bit to match RepositoryMigrationStatus. If the API returns a state that is not in this list, we want IsFailed() to return true.

We ran into a problem with repo migration status where it started returning a new state FAILED_VALIDATION, and we were interpreting any unexpected state as Pending, so it sent the CLI into an infinte loop.

We intentionally decided to interpret any unexpected state string as Failed.

src/gei/Commands/MigrateOrgCommand.cs
```diff
@@ -92,7 +93,29 @@ public async Task Invoke(MigrateOrgCommandArgs args)
...
+            {
+                _log.LogInformation($"Migration in progress (ID: {migrationId}). State: {migrationState}. Waiting 10 seconds...");
+                await Task.Delay(10000);
+                migrationState = await githubApi.GetOrganizationMigrationState(migrationId);
+            }
+
+            if (OrganizationMigrationStatus.IsFailed(migrationState))
+            {
+                _log.LogError($"Migration Failed. Migration ID: {migrationId}");
+                throw new OctoshiftCliException($"Migration Failed. Migration ID: {migrationId}");
```
dylan-smith, line 115: Can we get the failureReason from octoshift and display this to the user here? That's what we do in migrate-repo

dylan-smith:
I notice the GQL we use to check the status of an org migration is different than checking the status of a repo migration. Does this mean if a customer runs `gh gei migrate-org` without `--wait`, gets the migration ID, then tries to call `gh gei wait-for-migration --migration-id 1234` it won't work?

We could do something like `wait-for-org-migration` command, but that seems like unnecessary complexity / confusing for customers. Could we make `wait-for-migration` command work with any migration-id, regardless of whether its an org or repo migration?

src/gei/Commands/MigrateOrgCommand.cs
teja-nanduri: I was not able to find the equivalent in the graphql classes. I will check with @juanmrad to see if it exists or will be implemented.

src/Octoshift/OrganizationMigrationStatus.cs
```diff
@@ -12,7 +12,7 @@ public static class OrganizationMigrationStatus
         public const string RepoMigration = "REPO_MIGRATION";
 
         public static bool IsSucceeded(string migrationState) => migrationState?.Trim().ToUpper() is Succeeded;
-        public static bool IsFailed(string migrationState) => migrationState?.Trim().ToUpper() is Failed;
-        public static bool IsPending(string migrationState) => !IsFailed(migrationState) && !IsSucceeded(migrationState);
+        public static bool IsPending(string migrationState) => migrationState?.Trim().ToUpper() is Queued or InProgress or NotStarted or PostRepoMigration or PreRepoMigration or RepoMigration;
+        public static bool IsFailed(string migrationState) => !(IsPending(migrationState) || IsSucceeded(migrationState));
```
dylan-smith, line 16: I think this should be && not ||

teja-nanduri: I have switched it from `!A && !B` to `!( A || B )` to be consistent with the logic in `RepositoryMigrationStatus`.

dylan-smith: you're right, I missed that first !

src/OctoshiftCLI.Tests/GithubApiTests.cs
```diff
@@ -399,6 +400,38 @@ public async Task GetOrganizationId_Returns_The_Org_Id()
...
+                        ""enterprise"": 
+                            {{
+                                ""slug"": ""{GITHUB_ENTERPRISE}"",
+                                ""id"": ""{enterpriseId}""
+                            }} 
+                    }} 
+            }}";
+
+            _githubClientMock
+                .Setup(m => m.PostAsync(url, It.Is<object>(x => x.ToJson() == payload), null))
```
dylan-smith, line 425: should this be passing the extra header to PostAsync?

dylan-smith:
The custom header stuff looks good, but I'd like to see at least one unit test for it. I don't think it's necessary to add custom header unit tests for every single method in GithubClient that accepts custom headers, but we should do at least one of them.

dylan-smith:
Do you know what happens when the org-to-org stuff is no longer behind the internal schema flag? Will this break the CLI if we're still passing the internal schema header?

dylan-smith (changes requested):
Needs unit test(s) for custom headers.

teja-nanduri:
> Do you know what happens when the org-to-org stuff in the backend is no longer behind the internal schema flag? Will this break the CLI if we're still passing the internal schema header?

I was able to verify that the internal schema header is not causing any issues with the other requests, even after my JIT expired. So it will not break the CLI.

src/Octoshift/GithubClient.cs
```diff
@@ -122,6 +128,8 @@ public virtual async IAsyncEnumerable<JToken> GetAllAsync(string url)
                 _log.LogVerbose($"HTTP BODY: {body.ToJson()}");
             }
 
+            AddHeaders(customHeaders);
```
ArinGhazarian, line 131: At this point probably it's easier to create a `HttpRequestMessage` and use the `HttpClient#SendAsync()` method that accepts an http request instead of adding and removing the headers every single time. This way we wouldn't need the `switch` statement and it will be less code overall 😄 this is roughly what I mean:

```c#
...
_log.LogVerbose($"HTTP {httpMethod}: {url}");

using var request = new HttpRequestMessage(httpMethod, url).AddHeaders(customHeaders);

if (body != null)
{
    _log.LogVerbose($"HTTP BODY: {body.ToJson()}");
    request.Content = body.ToJson().ToStringContent();
}

using var response = await _httpClient.SendAsync(request);

// the rest of the method body stays exactly the same
```

`request.AddHeaders()` is gonna be an extension method like below, or if you don't feel like adding one you can use the `AddHeaders()` private method. 
```c#
public static HttpRequestMessage AddHeaders(this HttpRequestMessage request, Dictionary<string, string> headers)
{
    ...
}
```

ArinGhazarian (commented):
I suggested a fairly small change that I think is going to make the method way simpler.

src/Octoshift/GithubClient.cs
teja-nanduri: Love this approach. Switched to this one!!!

ArinGhazarian (approved):
🥇

teja-nanduri:
>Could we make wait-for-migration command work with any migration-id, regardless of whether its an org or repo migration?

I think this is the approach to take for this command. I am not sure but this might be big enough to be done as a separate PR. What do you think?

src/OctoshiftCLI.Tests/GithubClientTests.cs
```diff
@@ -53,6 +53,28 @@ public async Task GetAsync_Returns_String_Response()
...
+
+            // Act
+            var internalSchemaHeader = new Dictionary<string, string>() { { ghraphQLSchemaHeaderName, ghraphQLSchemaHeaderValue } };
+            await githubClient.GetAsync("http://example.com", internalSchemaHeader);
+
+            // Assert
+            handlerMock.Protected().Verify(
+                "SendAsync",
+                Times.Once(),
+                ItExpr.Is<HttpRequestMessage>(msg => msg.Headers.Any(kv => kv.Key == ghraphQLSchemaHeaderName && kv.Value.First() == ghraphQLSchemaHeaderValue)),
```
dylan-smith, line 74:
```suggestion
            const string graphQLSchemaHeaderName = "GraphQL-schema";
            const string graphQLSchemaHeaderValue = "internal";
            var handlerMock = MockHttpHandlerForGet();
            using var httpClient = new HttpClient(handlerMock.Object);
            var githubClient = new GithubClient(_mockOctoLogger.Object, httpClient, null, null);

            // Act
            var internalSchemaHeader = new Dictionary<string, string>() { { graphQLSchemaHeaderName, graphQLSchemaHeaderValue } };
            await githubClient.GetAsync("http://example.com", internalSchemaHeader);

            // Assert
            handlerMock.Protected().Verify(
                "SendAsync",
                Times.Once(),
                ItExpr.Is<HttpRequestMessage>(msg => msg.Headers.Any(kv => kv.Key == graphQLSchemaHeaderName && kv.Value.First() == graphQLSchemaHeaderValue)),
```

if we don't fix this typo it will drive me insane!

dylan-smith:
> > Could we make wait-for-migration command work with any migration-id, regardless of whether its an org or repo migration?
> 
> @dylan-smith I think this is the approach to take for this command. I am not sure but this might be big enough to be done as a separate PR. What do you think?

Pinging @roferg for input, but I'm ok with that.  To save @roferg some reading, the comment is that the migration ID returned from `migrate-org` will not work with the `wait-for-migration` command the way things are implemented today.

dylan-smith:
> > Could we make wait-for-migration command work with any migration-id, regardless of whether its an org or repo migration?
> 
> @dylan-smith I think this is the approach to take for this command. I am not sure but this might be big enough to be done as a separate PR. What do you think?

@teja-nanduri can you please make sure an issue is created for that (either in this repo or octoshift) and link it to this PR.

teja-nanduri:
I have created a [new issue](https://github.com/github/gh-gei/issues/606) to modify the `wait-for-migration` command to support both repo and org migration ids.

dylan-smith (approved):

Issue #525, opened 2022-07-26
Transfer the archive off BBS Server to CLI machine (using SSH)


dylan-smith:
- [x] Create a service to encapsulate ssh interaction (SCP over SSH) with the BBS server.
- [x] Add the download archive functionality to `migrate-repo` command.

PR #601, opened 2022-08-25
Add BBS archive downloader service

ArinGhazarian:
Related to #525 

This PR adds `BbsArchiveDownloader` service to be injected into the `migrate-repo` command later.

**Note:**
I created an abstraction for file and directory IO since I noticed that it's a patterns that was getting repeated a lot. Since it is only going to be used for testing, I didn't inject it to the class and just declared it as an `internal` instance field which is supposed to set only in a unit test. We should probably also refactor our other classes to use it, but it is going to b a `tech-debt` that we can tackle later. 

- [x] Issue linked


src/bbs2gh/Services/BbsArchiveDownloader.cs
```diff
@@ -0,0 +1,90 @@
...
+    private const int DOWNLOAD_PROGRESS_REPORT_INTERVAL_IN_SECONDS = 10;
+    private const string DEFAULT_BBS_SHARED_HOME_DIRECTORY = "/var/atlassian/application-data/bitbucket/shared";
+    private const string EXPORT_ARCHIVE_SOURCE_DIRECTORY = "data/migration/export";
+
+    private readonly ScpClient _scpClient;
+    private readonly OctoLogger _log;
+    private DateTime _nextProgressReport;
+
+    internal FileSystemProvider FileSystemProvider = new();
+    internal Action<string, FileInfo> ScpDownload;
```
dylan-smith, line 18: can we not mock ScpClient itself? instead of doing this indirection via function pointers

ArinGhazarian: without mocking it'd be impossible to unit test it. I had two choices, either to create an abstraction (`IScpClient` or something, that is going to be just and adapter) to be mocked later in tests or the delegate approach that I already took.
That being said if you think it is too much, I can get rid of the unit tests and make the downloader so much simpler.

dylan-smith: Why can't we do `new Mock<ScpClient>()` ?

dylan-smith: oh I think I see, it's because the functions aren't virtual in ScpClient. Interestingly if we were to use SftpClient it has an ISftpClient we could mock.

ArinGhazarian: addressed in 8c3a31526e42e969e554ca6f7fcbd5db878b2665

src/bbs2gh/Services/BbsArchiveDownloader.cs
```diff
@@ -46,22 +44,31 @@ public virtual void Download(long exportJobId, string targetDirectory = "bbs_arc
             throw new OctoshiftCliException($"Target export archive ({targetExportArchiveFullPath}) already exists.");
         }
 
-        FileSystemProvider.CreateDirectory(targetDirectory);
-
-        ScpDownload(sourceExportArchiveFullPath, new FileInfo(targetExportArchiveFullPath));
-    }
+        if (_sftpClient is BaseClient { IsConnected: false } clinet)
```
dylan-smith, line 47: cool use of pattern matching!

src/bbs2gh/Services/BbsArchiveDownloader.cs
```diff
@@ -0,0 +1,97 @@
...
+{
+    private const int DOWNLOAD_PROGRESS_REPORT_INTERVAL_IN_SECONDS = 10;
+    private const string DEFAULT_BBS_SHARED_HOME_DIRECTORY = "/var/atlassian/application-data/bitbucket/shared";
+    private const string EXPORT_ARCHIVE_SOURCE_DIRECTORY = "data/migration/export";
+
+    private readonly ISftpClient _sftpClient;
+    private readonly OctoLogger _log;
+    private DateTime _nextProgressReport;
+
+    internal FileSystemProvider FileSystemProvider = new();
```
dylan-smith, line 18: why don't we inject this?

src/OctoshiftCLI.Tests/bbs2gh/Services/BbsArchiveDownloaderTests.cs
```diff
@@ -12,72 +13,82 @@ public sealed class BbsArchiveDownloaderTests : IDisposable
...
         // Act
-        _bbsArchiveDownloader.Download(EXPORT_JOB_ID);
+        await _bbsArchiveDownloader.Download(EXPORT_JOB_ID, TARGET_DIRECTORY);
 
         // Assert
-        actualSource.Should().Be(expectedSourceArchiveFullName);
-        actualTarget.Should().Be(_exportArchiveFilename);
+        _mockSftpClient.Verify(m =>
+            m.BeginDownloadFile(
+                It.Is<string>(actual => actual == expectedSourceArchiveFullName),
```
dylan-smith, line 54:
```suggestion
                expectedSourceArchiveFullName,
```

dylan-smith: isn't that doing the exact same thing as my suggestion?

src/OctoshiftCLI.Tests/bbs2gh/Services/BbsArchiveDownloaderTests.cs
```diff
@@ -12,72 +13,82 @@ public sealed class BbsArchiveDownloaderTests : IDisposable
...
+                It.IsAny<Stream>(),
+                null,
+                null,
+                It.IsAny<Action<ulong>>()));
+
+        _mockFileSystemProvider.Verify(m => m.Open(It.Is<string>(actual => actual == expectedTargetArchiveFullName), FileMode.CreateNew));
```
dylan-smith, line 60:
```suggestion
        _mockFileSystemProvider.Verify(m => m.Open(expectedTargetArchiveFullName, FileMode.CreateNew));
```

src/OctoshiftCLI.Tests/bbs2gh/Services/BbsArchiveDownloaderTests.cs
```diff
@@ -12,72 +13,82 @@ public sealed class BbsArchiveDownloaderTests : IDisposable
...
-        _bbsArchiveDownloader.Download(EXPORT_JOB_ID);
+        await _bbsArchiveDownloader.Download(EXPORT_JOB_ID, TARGET_DIRECTORY);
 
         // Assert
-        actualSource.Should().Be(expectedSourceArchiveFullName);
-        actualTarget.Should().Be(_exportArchiveFilename);
+        _mockSftpClient.Verify(m =>
+            m.BeginDownloadFile(
+                It.Is<string>(actual => actual == expectedSourceArchiveFullName),
+                It.IsAny<Stream>(),
```
dylan-smith, line 55: Wouldn't it be better to do a _mockFileSystemProvider.Setup(...) that returns some stream, then verify that stream is what's passed into BeginDownloadFile? If we did that we wouldn't need the Verify on _mockFileSystemProvider either.

ArinGhazarian: I tried to do it in the previous commit, but unfortunately `FileStream` is hard and even impossible to mock and creating a dummy one is also impossible. In the previous commit I tried to mock the `FileStream` like the following:
```c#
var expectedTargetArchive = new Mock<FileStream>(new SafeFileHandle(IntPtr.Zero, false), FileAccess.Read);
```
but it failed our `Windows` CI because of the fact that `.NET 6` added some extra checks for `SafeFileHandle` so it's no longer possible to create invalid file handles (eg. `new SafeFileHandle(IntPtr.Zero, false)`).
The only way to test/mock `FileStream` (or any other `Stream`) would be to create abstractions (eg. `IFileStream`), but IMO it's going to be too much effort for not that much of a gain!

src/bbs2gh/Services/BbsArchiveDownloader.cs
ArinGhazarian: I wanted to keep it simple because it is only supposed to be used for tests and it's not even visible to other assemblies (except our unit tests one). But if you think it's better to inject it I'll do so.

src/OctoshiftCLI.Tests/bbs2gh/Services/BbsArchiveDownloaderTests.cs
ArinGhazarian: good catch! I was playing with that predicate and forgot to put everything back.

src/bbs2gh/Services/BbsArchiveDownloader.cs
ArinGhazarian: addressed in 508e086bbd64c5766ba77cdd00b690d1ed1ec286

src/OctoshiftCLI.Tests/bbs2gh/Services/BbsArchiveDownloaderTests.cs
ArinGhazarian: addressed in 508e086bbd64c5766ba77cdd00b690d1ed1ec286

src/OctoshiftCLI.Tests/bbs2gh/Services/BbsArchiveDownloaderTests.cs
ArinGhazarian: addressed in 508e086bbd64c5766ba77cdd00b690d1ed1ec286

dylan-smith (approved):

PR #619, opened 2022-09-08
Fix BBS download progress log's race condition

ArinGhazarian:
This is a quick PR to address the BBS download progress log's race condition by putting `LogProgress()` method's body in a critical section. The race condition was causing an unsynchronized log output (as shown below) which is at the very least a bad user experience:
```
[10:09 AM] [INFO] Download archive 38 started...
[10:09 AM] [INFO] Download archive in progress, 64 KB out of 111 KB (57%) completed...
[10:09 AM] [INFO] Download archive in progress, 32 KB out of 111 KB (28%) completed...
[10:09 AM] [INFO] Download archive in progress, 96 KB out of 111 KB (86%) completed...
```

- [x] Issue linked

### DIFF
--- a/src/bbs2gh/Services/BbsSshArchiveDownloader.cs
+++ b/src/bbs2gh/Services/BbsSshArchiveDownloader.cs
@@ -14,6 +14,7 @@ public sealed class BbsSshArchiveDownloader : IBbsArchiveDownloader, IDisposable
     private readonly ISftpClient _sftpClient;
     private readonly OctoLogger _log;
     private readonly FileSystemProvider _fileSystemProvider;
+    private readonly object _mutex = new();
     private DateTime _nextProgressReport;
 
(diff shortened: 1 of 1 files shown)


dylan-smith:
does this mean that something in that code is spinning up multiple threads? That is a bit concerning, for example I don't think we've ever really looked to see if OctoLogger is thread-safe.

ArinGhazarian:
> does this mean that something in that code is spinning up multiple threads? That is a bit concerning, for example I don't think we've ever really looked to see if OctoLogger is thread-safe.

Yes, the SFTP download code is an asynchronous code and it spins up multiple threads. Our logger also needs to be thread safe but this was the least thing I could do to fix it.

```c#
```

dylan-smith (approved):

Issue #494, opened 2022-07-13
Add rate-limit intelligence to our GitHub API calls


dylan-smith:
When we call Azure DevOps API's we look at and respect the rate limit headers returned, but we don't do the same thing for GitHub rate limits.

We should be looking at the [GitHub Rate Limit HTTP Response Headers](https://docs.github.com/en/rest/overview/resources-in-the-rest-api#rate-limit-http-headers) and logging/warning the user when the limit is close or hit, and waiting until the limit resets before making more API calls.

roferg:
When a customer queueing migrations hits a rate limit, the CLI should wait until the rate limit refreshes. During this time it should display a message to a customer that it's waiting.

PR #623, opened 2022-09-13
Use `X-RateLimit-Reset` headers to throttle GitHub requests

tambling:
This PR adds code to `GithubClient` to introspect the `X-RateLimit-Remaining` and `X-RateLimit-Reset` headers on GitHub requests (both REST and GraphQL; I checked) to pause the CLI until their rate limit resets. On each request, it checks if the `X-RateLimit-Remaining` value is `0` (indicating no requests left), then extracts the `X-RateLimit-Reset` header (which is the Unix epoch second that the rate limit resets), calculates the diff between that and the present time, and sets that as `_retryDelay`. Once `_retryDelay` is set, `GithubClient` behaves identically to `AdoClient`.

Closes https://github.com/github/gh-gei/issues/494

- [ ] Issue linked


src/Octoshift/GithubClient.cs
```diff
@@ -156,12 +162,24 @@ public virtual async IAsyncEnumerable<JToken> GetAllAsync(string url, Dictionary
             {
                 response.EnsureSuccessStatusCode();
             }
-            else if (response.StatusCode != status)
+            else if (response.StatusCode != status && RateLimitRemaining(headers) != 0)
```
tambling, line 165: Tests still fail if I simulate an HTTP response with a status code of 403 (which could happen with a rate-limited response) — how best to make this not raise an error if we get a 403 and `X-RateLimit-Remaining: 0`?

src/Octoshift/GithubClient.cs
```diff
@@ -186,8 +204,24 @@ private string GetNextUrl(KeyValuePair<string, IEnumerable<string>>[] headers)
         private string ExtractLinkHeader(IEnumerable<KeyValuePair<string, IEnumerable<string>>> headers) =>
             ExtractHeaderValue("Link", headers);
 
+        private int RateLimitRemaining(IEnumerable<KeyValuePair<string, IEnumerable<string>>> headers) =>
+            int.Parse(ExtractHeaderValue("X-RateLimit-Remaining", headers) ?? "5000");
+
+        private long RateLimitReset(IEnumerable<KeyValuePair<string, IEnumerable<string>>> headers) =>
+            long.Parse(ExtractHeaderValue("X-RateLimit-Reset", headers) ?? DateTimeOffset.Now.ToUnixTimeSeconds().ToString());
+
```
ArinGhazarian, line 212: It's customary in C# for method names to be verbs or verb phrases. Microsoft also recommends it in [design guidelines for naming methods](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-methods):
> Because methods are the means of taking action, the design guidelines require that method names be verbs or verb phrases. Following this guideline also serves to distinguish method names from property and type names, which are noun or adjective phrases.

So these methods are better to be named as `GetRemainingRateLimit` and `ResetRateLimit`.

src/Octoshift/GithubClient.cs
```diff
@@ -186,8 +204,24 @@ private string GetNextUrl(KeyValuePair<string, IEnumerable<string>>[] headers)
         private string ExtractLinkHeader(IEnumerable<KeyValuePair<string, IEnumerable<string>>> headers) =>
             ExtractHeaderValue("Link", headers);
 
+        private int RateLimitRemaining(IEnumerable<KeyValuePair<string, IEnumerable<string>>> headers) =>
+            int.Parse(ExtractHeaderValue("X-RateLimit-Remaining", headers) ?? "5000");
+
+        private long RateLimitReset(IEnumerable<KeyValuePair<string, IEnumerable<string>>> headers) =>
+            long.Parse(ExtractHeaderValue("X-RateLimit-Reset", headers) ?? DateTimeOffset.Now.ToUnixTimeSeconds().ToString());
```
ArinGhazarian, line 211: I would use the `_dateTimeProvider` abstraction:
```suggestion
            long.Parse(ExtractHeaderValue("X-RateLimit-Reset", headers) ?? _dateTimeProvider.CurrentUnixTimeSeconds().ToString());
```

src/Octoshift/GithubClient.cs
```diff
@@ -186,8 +204,24 @@ private string GetNextUrl(KeyValuePair<string, IEnumerable<string>>[] headers)
         private string ExtractLinkHeader(IEnumerable<KeyValuePair<string, IEnumerable<string>>> headers) =>
             ExtractHeaderValue("Link", headers);
 
+        private int RateLimitRemaining(IEnumerable<KeyValuePair<string, IEnumerable<string>>> headers) =>
+            int.Parse(ExtractHeaderValue("X-RateLimit-Remaining", headers) ?? "5000");
```
ArinGhazarian, line 208: I would use a `const` (eg. `DEFAULT_RATE_LIMIT_REMAINING`) here.

src/Octoshift/GithubClient.cs
```diff
@@ -186,8 +204,24 @@ private string GetNextUrl(KeyValuePair<string, IEnumerable<string>>[] headers)
...
+
+        private long RateLimitReset(IEnumerable<KeyValuePair<string, IEnumerable<string>>> headers) =>
+            long.Parse(ExtractHeaderValue("X-RateLimit-Reset", headers) ?? DateTimeOffset.Now.ToUnixTimeSeconds().ToString());
+
         private string ExtractHeaderValue(string key, IEnumerable<KeyValuePair<string, IEnumerable<string>>> headers) =>
             headers.SingleOrDefault(kvp => kvp.Key.Equals(key, StringComparison.OrdinalIgnoreCase)).Value?.FirstOrDefault();
 
+        private void CheckForRetryDelay(IEnumerable<KeyValuePair<string, IEnumerable<string>>> headers)
+        {
+            if (RateLimitRemaining(headers) == 0)
```
ArinGhazarian, line 218: nit: not sure if this can ever be `< 0` but I guess it won't hurt to write it as `if (RateLimitRemaining(headers) <= 0)` just in case

src/Octoshift/GithubClient.cs
```diff
@@ -186,8 +204,24 @@ private string GetNextUrl(KeyValuePair<string, IEnumerable<string>>[] headers)
...
+            {
+                var resetUnixSeconds = RateLimitReset(headers);
+                var currentUnixSeconds = _dateTimeProvider.CurrentUnixTimeSeconds();
+
+                _retryDelay = (resetUnixSeconds - currentUnixSeconds) * 1000.0;
```
ArinGhazarian, line 223: here also I would create a `const` for 1000, and make it clear that it's milliseconds (or whatever).

src/Octoshift/GithubClient.cs
```diff
@@ -156,12 +162,24 @@ public virtual async IAsyncEnumerable<JToken> GetAllAsync(string url, Dictionary
...
 
-            return (content, response.Headers.ToArray());
+            CheckForRetryDelay(headers);
+
+            return (content, headers);
+        }
+
+        private async Task ApplyRetryDelayAsync()
+        {
+            if (_retryDelay > 0.0)
```
ArinGhazarian, line 177: two nits:
1. there is no need to necessarily compare two `doubles` in C# the case is going to happen automatically otherwise  you're gonna get a compiler error clearly stating that there's no implicit case available for the comparison.
2. For literals it's better (more clear) to use their respective suffix, for more info take a look at [this](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/integral-numeric-types). So for double (which is the default for floating points) I'd write `0D` instead of `0.0`. Of course if you use explicit typing there is no need to use the suffix (eg. `double dbl = 1.5`)

src/OctoshiftCLI.IntegrationTests/AdoToGithub.cs
```diff
@@ -36,7 +36,7 @@ public AdoToGithub(ITestOutputHelper output)
 
             var githubToken = Environment.GetEnvironmentVariable("GHEC_PAT");
             _githubHttpClient = new HttpClient();
-            var githubClient = new GithubClient(logger, _githubHttpClient, new VersionChecker(_versionClient, logger), new RetryPolicy(logger), githubToken);
+            var githubClient = new GithubClient(logger, _githubHttpClient, new VersionChecker(_versionClient, logger), new RetryPolicy(logger), null, githubToken);
```
ArinGhazarian, line 39: wouldn't it be better to just pass in a `DateTimeProvider` instead of `null` ? it can be simply and inline instantiation.

ArinGhazarian (commented):
Just a few nitpicks but overall it's nicely implemented 💯

src/Octoshift/GithubClient.cs
ArinGhazarian: I guess the best would be to also check the error message in case of `403`

dylan-smith: I think it's because of a misunderstanding of how this if statement is working. In this code `status` is the argument passed in, which could probably be renamed to `expectedStatus`. In the vast majority of circumstances that will have the default value of HttpStatusCode.OK so the else here will never be hit, it will go into the first part of the if statement, EnsureSuccessStatusCode() will run, see a 403 and error out.

src/Octoshift/GithubClient.cs
```diff
@@ -156,12 +162,24 @@ public virtual async IAsyncEnumerable<JToken> GetAllAsync(string url, Dictionary
...
+            {
+                _log.LogWarning($"THROTTLING IN EFFECT. Waiting {(int)_retryDelay} ms");
```
dylan-smith, line 179: I would show the wait time in seconds to be more user-friendly (I even considered suggesting minutes, but seconds probably works).  I would also reword this a bit and not use the word throttling here, because the way github does things (unlike ADO) I don't really think of it as throttling.  Maybe something like "GitHub rate limit exceeded. Waiting 329 seconds before continuing".

src/Octoshift/GithubClient.cs
dylan-smith: Should probably be GetRateLimitReset based on my understanding of what it's doing.

src/OctoshiftCLI.IntegrationTests/GhesToGithub.cs
```diff
@@ -47,11 +47,11 @@ public GhesToGithub(ITestOutputHelper output)
         _versionClient = new HttpClient();
 
         _sourceGithubHttpClient = new HttpClient();
-        _sourceGithubClient = new GithubClient(logger, _sourceGithubHttpClient, new VersionChecker(_versionClient, logger), new RetryPolicy(logger), sourceGithubToken);
+        _sourceGithubClient = new GithubClient(logger, _sourceGithubHttpClient, new VersionChecker(_versionClient, logger), new RetryPolicy(logger), null, sourceGithubToken);
```
dylan-smith, line 50:
```suggestion
        _sourceGithubClient = new GithubClient(logger, _sourceGithubHttpClient, new VersionChecker(_versionClient, logger), new RetryPolicy(logger), new DateTimeProvider(), sourceGithubToken);
```

src/OctoshiftCLI.IntegrationTests/GithubToGithub.cs
```diff
@@ -33,7 +33,7 @@ public GithubToGithub(ITestOutputHelper output)
 
             _githubHttpClient = new HttpClient();
             _versionClient = new HttpClient();
-            _githubClient = new GithubClient(logger, _githubHttpClient, new VersionChecker(_versionClient, logger), new RetryPolicy(logger), githubToken);
+            _githubClient = new GithubClient(logger, _githubHttpClient, new VersionChecker(_versionClient, logger), new RetryPolicy(logger), null, githubToken);
```
dylan-smith, line 36:
```suggestion
            _githubClient = new GithubClient(logger, _githubHttpClient, new VersionChecker(_versionClient, logger), new RetryPolicy(logger), new DateTimeProvider(), githubToken);
```

src/OctoshiftCLI.IntegrationTests/GhesToGithub.cs
```diff
@@ -47,11 +47,11 @@ public GhesToGithub(ITestOutputHelper output)
...
         _sourceGithubApi = new GithubApi(_sourceGithubClient, GHES_API_URL, new RetryPolicy(logger));
 
         _targetGithubHttpClient = new HttpClient();
-        _targetGithubClient = new GithubClient(logger, _targetGithubHttpClient, new VersionChecker(_versionClient, logger), new RetryPolicy(logger), targetGithubToken);
+        _targetGithubClient = new GithubClient(logger, _targetGithubHttpClient, new VersionChecker(_versionClient, logger), new RetryPolicy(logger), null, targetGithubToken);
```
dylan-smith, line 54:
```suggestion
        _targetGithubClient = new GithubClient(logger, _targetGithubHttpClient, new VersionChecker(_versionClient, logger), new RetryPolicy(logger), new DateTimeProvider(), targetGithubToken);
```

dylan-smith (changes requested):
We should also have a unit test for when we get a 403 response and a 0 rate limit remaining header. In that case we should both wait and retry the failed HTTP call.

src/Octoshift/GithubClient.cs
ArinGhazarian: Oh yeah I though it was trying to reset the rate limit (this is exactly why we should rename it 😃), but didn't notice that it was actually returning the rate limit reset 😄

src/Octoshift/GithubClient.cs
```diff
@@ -16,13 +16,19 @@ public class GithubClient
     {
         private readonly HttpClient _httpClient;
         private readonly OctoLogger _log;
+        private double _retryDelay;
         private readonly RetryPolicy _retryPolicy;
+        private readonly DateTimeProvider _dateTimeProvider;
 
-        public GithubClient(OctoLogger log, HttpClient httpClient, IVersionProvider versionProvider, RetryPolicy retryPolicy, string personalAccessToken)
+        private const string DEFAULT_RATE_LIMIT_REMAINING = "5000";
+        private const float MILLISECONDS_PER_SECOND = 1000f;
```
ArinGhazarian, line 24: This doesn't need to be a `float`:
```suggestion
        private const int MILLISECONDS_PER_SECOND = 1000;
```

ArinGhazarian (commented):
Just a small nitpick and after that I am 👍

src/Octoshift/GithubClient.cs
tambling: CodeQL [whined](https://github.com/github/gh-gei/pull/623#pullrequestreview-1106528791) about possible loss of precision when casting the result of integer multiplication to a double, but if I should just leave it an integer and tell CodeQL to take a hike, I will!

ArinGhazarian: oh that's because the `_retryDelay` is a `double` but I guess it shouldn't be right?

dylan-smith: should probably be an int

dylan-smith (approved):

Issue #639, opened 2022-09-16
Infrastructure changes


ArinGhazarian:
### Description

- Necessary changes in `program.cs`.
- Reflection magic.
- Unit tests.
- Can do this one command at a time.

PR #652, opened 2022-09-21
CommandLine plumbing infrastructure

ArinGhazarian:
Related to #639

This is all the required infrastructure changes in form of some extension method to do the plumbing for the new CommandLine. Here is a [sample branch](https://github.com/github/gh-gei/tree/upgrade-bbs2gh-to-new-command-line) that applies these changes to `bbs2gh`. 
I went ahead and simplified it even more by creating an extension to add all commands to the `rootCommand`. Here is how it's supposed to be used:

```c#
public static async Task<int> Main(string[] args) // note that Main now returns a Task<int> instead of Task
{
    Logger.LogDebug("Execution Started");
 
    var serviceCollection = new ServiceCollection();
    ... // service registrations
    var serviceProvider = serviceCollection.BuildServiceProvider();

    var rootCommand = new RootCommand("Automate end-to-end Bitbucket Server to GitHub migrations.")
          .AddCommands(serviceProvider);

    ...
}
```

- [x] Issue linked


src/Octoshift/Extensions/CommandExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        foreach (var commandType in Assembly.GetCallingAssembly().GetAllDescendantsOfCommandBase())
+        {
+            var command = commandType.CreateInstance<Command>();
+
+            command.SetHandler(async ctx =>
+            {
+                var commandArgsType = commandType.BaseType.GetGenericArguments()[0];
+                var commandArgs = ctx.BindArgs(command, commandArgsType);
+                var handler = commandType.GetMethod("BuildHandler").Invoke(command, new[] { commandArgs, serviceProvider });
```
ArinGhazarian, line 28: Wouldn't it be better to check if `handler` implements `ICommandHander` and throw a meaningful exception? It is going to throw a `NullReferenceException` if `GetMethod` returns `null` in case it cannot find `BuildHandler` method anyways.

src/Octoshift/Commands/CommandBase.cs
```diff
@@ -0,0 +1,11 @@
+using System.CommandLine;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace OctoshiftCLI.Commands;
+
+public abstract class CommandBase<TArgs, THandler> : Command where TArgs : class where THandler : class
```
dylan-smith, line 6:
```suggestion
public abstract class CommandBase<TArgs, THandler> : Command where TArgs : class where THandler : ICommandHandler<TArgs>
```

ArinGhazarian: Great catch! I totally forgot to mark it as `ICommandHander` :)

dylan-smith (approved):

Issue #594, opened 2022-08-23
AWS S3 support for archive storage


dylan-smith:
Related to https://github.com/github/releases/issues/2532.

When migrating from GHES->GHEC you need to provide an `--azure-storage-connection-string` that is used for intermediate storage of the migration archives.

We should also support AWS S3 instead of Azure for customers that prefer that.

PR #653, opened 2022-09-21
Support uploading archives to AWS S3 buckets

synthead:
Closes https://github.com/github/gh-gei/issues/594.
Related to https://github.com/github/releases/issues/2532.

- [ ] Issue linked


src/Octoshift/AwsApi.cs
```diff
@@ -0,0 +1,51 @@
+﻿using System;
+using System.Threading.Tasks;
+using Amazon;
+using Amazon.S3;
+using Amazon.S3.Model;
+using Amazon.S3.Transfer;
+
+namespace OctoshiftCLI
```
ArinGhazarian, line 8: use file scoped namespace

src/Octoshift/AwsApi.cs
```diff
@@ -0,0 +1,51 @@
...
+                BucketName = bucketName,
+                Key = keyName,
+                Expires = expires
+            };
+
+            var urlString = amazonS3Client.GetPreSignedURL(urlRequest);
+
+            return new Uri(urlString);
+        }
+    }
```
ArinGhazarian, line 50: Since there is little bit of logic here (for GetPreSignedUrlForFile), it's better to create an `internal` constructor and pass in a `ITransferUtility` into it which is supposed to be used by out unit tests. I personally like the idea of keeping the `public` constructor as is for better encapsulation. So here is mu suggestion to re-write the class:
```suggestion
    public class AwsApi : IDisposable
    {
        private const int AUTHORIZATION_TIMEOUT_IN_HOURS = 24;
        private static readonly RegionEndpoint RegionEndpoint = RegionEndpoint.USEast1;
        
        private readonly ITransferUtility _transferUtility;

#pragma warning disable CA2000
        public AwsApi(string awsAccessKey, string awsSecretKey) : this(new TransferUtility(new AmazonS3Client(awsAccessKey, awsSecretKey, RegionEndpoint)))
#pragma warning restore CA2000
        {
        }

        internal AwsApi(ITransferUtility transferUtility) => _transferUtility = transferUtility;

        public virtual async Task<string> UploadToBucket(string bucketName, string fileName, string keyName)
        {
            await _transferUtility.UploadAsync(fileName, bucketName, keyName);
            return GetPreSignedUrlForFile(bucketName, keyName);
        }

        private string GetPreSignedUrlForFile(string bucketName, string keyName)
        {
            var expires = DateTime.Now.AddHours(AUTHORIZATION_TIMEOUT_IN_HOURS);

            var urlRequest = new GetPreSignedUrlRequest
            {
                BucketName = bucketName,
                Key = keyName,
                Expires = expires
            };

            return _transferUtility.S3Client.GetPreSignedURL(urlRequest);
        }

        protected virtual void Dispose(bool disposing)
        {
            if (disposing)
            {
                _transferUtility?.Dispose();
            }
        }

        public void Dispose()
        {
            Dispose(true);
            GC.SuppressFinalize(this);
        }
    }
```

Also note that I changed the signature of `UploadToBucket` from a `Task<Uri>` to `Task<string>` because based on the usage we just need the url string so there is no need to create a `Uri` only to convert it back to a string later.

src/bbs2gh/AwsApiFactory.cs
```diff
@@ -0,0 +1,22 @@
...
+        public AwsApiFactory(EnvironmentVariableProvider environmentVariableProvider, OctoLogger octoLogger)
+        {
+            _environmentVariableProvider = environmentVariableProvider;
+            _octoLogger = octoLogger;
+        }
+
+        public AwsApi Create(string awsAccessKey = null, string awsSecretKey = null)
+        {
+            var accessKey = string.IsNullOrWhiteSpace(awsAccessKey) ? _environmentVariableProvider.AwsAccessKey() : awsAccessKey;
+            var secretKey = string.IsNullOrWhiteSpace(awsSecretKey) ? _environmentVariableProvider.AwsSecretKey() : awsSecretKey;
```
ArinGhazarian, line 17:
```suggestion
            var accessKey = awsAccessKey ?? _environmentVariableProvider.AwsAccessKey();
            var secretKey = awsSecretKey ?? _environmentVariableProvider.AwsSecretKey();
```

src/bbs2gh/AwsApiFactory.cs
```diff
@@ -0,0 +1,22 @@
+namespace OctoshiftCLI.BbsToGithub
```
ArinGhazarian, line 1: use file scoped namespace

src/bbs2gh/AwsApiFactory.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    public sealed class AwsApiFactory : IAwsApiFactory
```
ArinGhazarian, line 3: There is no need to create a `IAwsApiFactory` interfacem the `Create` method can simply be marked as `virtual`

src/bbs2gh/IAwsApiFactory.cs
```diff
@@ -0,0 +1,7 @@
+namespace OctoshiftCLI.BbsToGithub
+{
+    public interface IAwsApiFactory
```
ArinGhazarian, line 3: Remove the interface.

src/bbs2gh/Handlers/MigrateRepoCommandHandler.cs
```diff
@@ -110,13 +115,34 @@ private async Task<long> GenerateArchive(MigrateRepoCommandArgs args)
...
-        var archiveBlobUrl = await _azureApi.UploadToBlob($"{guid}.tar", archiveData);
+        return $"{guid}.tar";
+    }
+
+    private async Task<string> UploadArchiveToAws(string bucketName, string accessKey, string secretKey, string archivePath)
+    {
+        _log.LogInformation("Uploading Archive to AWS...");
+
+        _ = accessKey ?? _environmentVariableProvider.AwsAccessKey();
+        _ = secretKey ?? _environmentVariableProvider.AwsSecretKey();
```
ArinGhazarian, line 142: What is this for? this is being taken care of by the factory already.

src/bbs2gh/Handlers/MigrateRepoCommandHandler.cs
```diff
@@ -110,13 +115,34 @@ private async Task<long> GenerateArchive(MigrateRepoCommandArgs args)
...
+
+        return archiveBlobUrl.ToString();
+    }
+
+    private string GenerateArchiveName()
+    {
         var guid = Guid.NewGuid().ToString();
-        var archiveBlobUrl = await _azureApi.UploadToBlob($"{guid}.tar", archiveData);
+        return $"{guid}.tar";
+    }
```
ArinGhazarian, line 135: nit: I would crate a one liner:
```c#
private string GenerateArchiveName() => $"{Guid.NewGuid()}.tar";
```

ArinGhazarian (changes requested):
I dropped some comments but the main one was for the `AwsApi` class and how we construct it. We should also add unit test for it to.

src/Octoshift/AwsApi.cs
tambling: @ArinGhazarian This suggestion, as you might imagine, took me the longest to wrap my head around, but I think I've pretty much got it. Two questions:
1. I assume the internal constructor that takes a `ITransferUtility` provides more flexibility to pass in a mock `TransferUtility`, but how? When testing, can we instantiate a `AwsApi` from its internal constructor?
2. Where did `IDisposable` come from?

ArinGhazarian: Very good questions: 
1. Generally speaking only abstractions can be mocked with the `moq` framework, which means that either an `interface` or `abstract` class, or `virtual` methods. The reason is simple, when you write `new Mock<Type>`, `moq` framework internally tries to subclass the `Type` and reimplement all methods that it can (you can then customize the methods that you want in terms of what they return, etc.). Abstractions are meant to be subclassed so they are mockable. `virtual` methods are also meant to be `overridden` if need be in a subclass so they are also mockable but there is no way to mock a non virtual method! If you look at the `TransferUtility`'s signature the `UploadAsync` method is non virtual so if we pass in just the `TransferUtility` it will be impossible to mock it, so instead we pass in the `ITransferUtility` abstraction.

> When testing, can we instantiate a `AwsApi` from its internal constructor?

In our actual code no, but in our unit tests yes with a simple trick. If you notice we marked the `bbs2gh` assembly, [InternalVisibleTo("Octoshift.Tests")](https://github.com/github/gh-gei/blob/b49ba197484750697bc872bd760e0f7d1cb38aac/src/bbs2gh/Program.cs#L13), with that, all internal members will be visible just to our unit tests assembly. The reason to mark that second constructor as `internal` is that we naturally want ppl to see only the constructor that takes two strings but for the sake of testability we created an internal one just for us to use.

2. Generally speaking if a class **owns** `IDisposable` objects (e.g. `ITransferUtility` or `IAmazonS3`) it needs to [properly dispose](https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose) them and release unmanaged resources. Unmanaged resources are resources that are not managed by .NET's CLR, for example, network connections, db connections, file handles etc. Now since `AwsApi` now owns the `ITransferUtility` it needs to properly dispose it so we make it disposable by subclassing `IDisposable` and implement the `Dispose` method. There is a reason why we need to do a `GC.SuppressFinalize()` and create a `Dispose(bool disposing)` method that is described [here](https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose) but the tl;dr is that these boiler plate codes are only necessary if you plan to  subclass your class otherwise if the class is marked as `sealed` you can simple just implement the `Dispose` method.  We are not planning on subclassing `AwsApi` but we cannot mark it as `sealed` because of the `virtual` methods (they are mutually exclusive), so we need to add that boilerplate code. 

** Also pay attention to the fact that we want the `AwsApi` to  **own** the `ITransferUtility` that's why we are disposing it upon disposal of the `AwsApi`. If this isn't the case then generally speaking it is the responsibility of whoever that instantiates an object to dispose it if need be.

src/bbs2gh/Program.cs
```diff
@@ -31,6 +31,7 @@ public static async Task Main(string[] args)
                 .AddSingleton<RetryPolicy>()
                 .AddSingleton<IAzureApiFactory, AzureApiFactory>()
                 .AddSingleton<IBlobServiceClientFactory, BlobServiceClientFactory>()
+                .AddSingleton<AwsApiFactory, AwsApiFactory>()
```
ArinGhazarian, line 34:
```suggestion
                .AddSingleton<AwsApiFactory>()
```

src/bbs2gh/AwsApiFactory.cs
```diff
@@ -0,0 +1,19 @@
...
+
+    public AwsApiFactory(EnvironmentVariableProvider environmentVariableProvider)
+    {
+        _environmentVariableProvider = environmentVariableProvider;
+    }
+
+    public virtual AwsApi Create(string awsAccessKey = null, string awsSecretKey = null)
+    {
+        var accessKey = _environmentVariableProvider.AwsAccessKey() ?? awsAccessKey;
+        var secretKey = _environmentVariableProvider.AwsSecretKey() ?? awsSecretKey;
```
ArinGhazarian, line 15: These won't work as intended and should be inverted because if the env variable is not set it is going to throw:
```suggestion
        var accessKey = awsAccessKey ?? _environmentVariableProvider.AwsAccessKey();
        var secretKey = awsSecretKey ?? _environmentVariableProvider.AwsSecretKey();
```

src/OctoshiftCLI.Tests/AwsApiTests.cs
```diff
@@ -0,0 +1,48 @@
...
+        string result;
+
+        try
+        {
+            result = await awsApi.UploadToBucket(bucketName, fileName, keyName);
+        }
+        finally
+        {
+            awsApi.Dispose();
+        }
```
ArinGhazarian, line 42: There's already a syntactic sugar for disposable objects in c#, behind the scenes it does exactly what you did:

```suggestion
        using var awsApi = new AwsApi(transferUtility.Object);

        // Act
        var result = await awsApi.UploadToBucket(bucketName, fileName, keyName);
```

src/Octoshift/AwsApi.cs
```diff
@@ -0,0 +1,66 @@
...
+
+#pragma warning disable CA2000
+    public AwsApi(string awsAccessKey, string awsSecretKey) : this(new TransferUtility(new AmazonS3Client(awsAccessKey, awsSecretKey, RegionEndpoint)))
+#pragma warning restore CA2000
+    {
+    }
+
+    internal AwsApi(ITransferUtility transferUtility) => _transferUtility = transferUtility;
+
+    public virtual async Task<string> UploadFileToBucket(string bucketName, string fileName, string keyName)
```
ArinGhazarian, line 26: Not a huge concern and blocker but we can use method overloading here and keep the method names the same, exactly how `TransferUtility#UploadAsync()` is overloaded. So this way we will have the following methods:
```c#
public virtual async Task<string> UploadToBucket(string bucketName, string fileName, string keyName)
{
    ...
}

public virtual async Task<string> UploadToBucket(string bucketName, byte[] bytes, string keyName)
{
    ...
}
```

I will leave the decision to change up to you though.

ArinGhazarian (approved):
Looks good 🥇

Issue #638, opened 2022-09-16
Introduce Handler as separate class


ArinGhazarian:
### Description
We are planning on separating commands from their handlers. To do so we need to create a handler for each command and basically move all of the business logic into it. The handler will be registered and injected into the command. In order to keep the PR as small as possible we can do this one command at a time.

PR #654, opened 2022-09-21
Refactor separate CommandHandler classes for the shared commands

dylan-smith:
Split out a CommandHandler class for all the shared commands in Octoshift. This is in preparation for the next step in the refactor where we will change how we wire up the command/handlers using some new binding/reflection code.

Issue #638 

- [x] Issue linked


src/Octoshift/Handlers/CreateTeamCommandBaseHandler.cs
```diff
@@ -0,0 +1,77 @@
+using System;
+using System.Linq;
+using System.Threading.Tasks;
+using OctoshiftCLI.Commands;
+using OctoshiftCLI.Contracts;
+
+namespace OctoshiftCLI.Handlers;
+
+public class CreateTeamCommandBaseHandler
```
ArinGhazarian, line 9: nit: I would rename this to just `CreateTeamCommandHandler`, because having `Base` kinda indicates that this is supposed to be subclassed but it's not the case here. The fact that these are in our shared project means that these are shared handlers already.

src/Octoshift/Handlers/GrantMigratorRoleCommandBaseHandler.cs
```diff
@@ -0,0 +1,64 @@
...
+{
+    private readonly OctoLogger _log;
+    private readonly ITargetGithubApiFactory _githubApiFactory;
+
+    public GrantMigratorRoleCommandBaseHandler(OctoLogger log, ITargetGithubApiFactory githubApiFactory)
+    {
+        _log = log;
+        _githubApiFactory = githubApiFactory;
+    }
+    public async Task Handle(GrantMigratorRoleCommandArgs args)
```
ArinGhazarian, line 18: nit: missing new line.

ArinGhazarian (approved):
Looks good, I just added two non-blocking nitpicks.

src/Octoshift/Handlers/CreateTeamCommandBaseHandler.cs
dylan-smith: Yeah I went back and forth on this. I'm gonna change it.

src/Octoshift/Handlers/GrantMigratorRoleCommandBaseHandler.cs
dylan-smith: huh I would have expected dotnet format to catch that. fixing.

PR #663, opened 2022-09-24
BBS new command binding

dylan-smith:
The final refactoring for bbs2gh. Changed the way we construct commands and the dependency graph to the final state. (these code changes were mostly taken from @ArinGhazarian branch where he figured this all out).

More PR's to follow for the other projects.

Issue #639

- [x] Issue linked


src/bbs2gh/Program.cs
```diff
@@ -37,14 +36,24 @@ public static async Task Main(string[] args)
...
-            SetContext(parser.Parse(args));
+            var commandLineBuilder = new CommandLineBuilder(rootCommand);
+            var parser = commandLineBuilder
+                .UseDefaults()
+                .UseExceptionHandler((ex, _) =>
+                {
+                    Logger.LogError(ex);
+                    Environment.ExitCode = 1;
+                }, 1)
+                .Build();
```
ArinGhazarian, line 54: Why are we using the `CommandLineBuilder` again? this is not the recommended way anymore. If you are doing this to set the exit code, the proper way is to modify the `Main` method to return `Task<int>` instead of a `Task` and it should just work. The only thing is to then wrap the `InvokeAsync` in a `try..catch` block and log the error instead of letting it bubble up.

ArinGhazarian: So I tested the `try..catch` locally and it didn't work. By looking at the docs however, I realized that the correct place to do the `try..catch` logic would be inside the `SetHander` method. But if we swallow the exception then we also need to set the exit code manually (kinda like the old way of using the `UseExceptionHandler`). If we go that route (that I guess we should) we then need to use an overload of `SetHandler` that accepts `Action<InvocationContext>` so we are going to need the `InvocationContext`. The `SetHandler` will look like the following:
```c#
command.SetHandler(async ctx =>
{
    ... // pre invocation
    
    try
    {
        // Invocation
    }
    catch (Exception ex)
    {
        sp.GetRequiredService<OctoLogger>().LogError(ex);
        ctx. ExitCode = 1;
    }
});
```

PR #679, opened 2022-10-11
Move GHES archive-downloading logic out of `AzureApi`

tambling:
When we shipped #653, we missed a case where omitting `--azure-storage-connection-string` causes the app to crash, because `AzureApiFactory.Create` was running anyway. After a bit of digging, I discovered that we were using `AzureApi` to make unauthenticated HTTP GETs to download the archives off of the GHES instance. This PR moves `AzureApi.DownloadArchive` into its own class, so that we can encapsulate making unauthenticated HTTP GETs.

- [ ] Issue linked


src/gei/Commands/MigrateRepoCommand.cs
```diff
@@ -161,7 +162,7 @@ public class MigrateRepoCommand : Command
             AddOption(adoPat);
             AddOption(verbose);
 
-            var handler = new MigrateRepoCommandHandler(log, sourceGithubApiFactory, targetGithubApiFactory, environmentVariableProvider, azureApiFactory, awsApiFactory);
+            var handler = new MigrateRepoCommandHandler(log, sourceGithubApiFactory, targetGithubApiFactory, environmentVariableProvider, azureApiFactory, awsApiFactory, new FileDownloader(new HttpClient(), log));
```
ArinGhazarian, line 165: All dependencies (except the command handlers) should be registered and injected to the constructor, so here you need to register the `FileDownloader` so you won't need to worry about constructing all dependencies and everything will be constructed for you. To register `FileDownloader` add this to our registrations:

```c#
.AddSingleton<FileDownloader>()
``` 

and then inject it to the constructor.

P.S.: With the new infrastructure changes you'll need to get the `FileDownloader` directly from the `IServiceProvider`:
```c#
var fileDownloader = sp.GetRequiredService<FileDownloader>();
```

ArinGhazarian (changes requested):

src/gei/FileDownloader.cs
```diff
@@ -0,0 +1,26 @@
...
+    private readonly HttpClient _client;
+    private readonly OctoLogger _octoLogger;
+
+    public FileDownloader(HttpClient client, OctoLogger octoLogger)
+    {
+        _client = client;
+        _octoLogger = octoLogger;
+    }
+
+    public virtual async Task<byte[]> DownloadArchive(string fromUrl)
```
ArinGhazarian, line 17: Since this is a generic file downloader I would call this method just `Download`.

dylan-smith (approved):

dylan-smith:
You should add a bullet to the release notes about the bug that we fixed (even if it was really the previous refactor that actually fixed it - because I didn't release note it then).

Issue #678, opened 2022-10-11
Add kerberos support for BBS


dylan-smith:
Add support for kerberos authentication when talking to the BBS server API's.

PR #684, opened 2022-10-15
Add Kerberos support

dylan-smith:
Issue #678 

Added support for kerberos authentication when calling BBS API's. To do this all we need to do is configure our HttpClient to have `UseDefaultCredentials = true` and NOT set the normal Authorization header that we set. This needs to be set at the time we construct the HttpClient. Similar to how we did the NoSSL support, I added a couple different named HttpClient's into our DI container ("Default" and "Kerberos") configured appropriately. Then I added a new `CreateKerberos()` function to `BbsApiFactory` that doesn't require a username/password. Finally I added a 2nd constructor to `BbsApi` that doesn't require a username/password and doesn't attempt to set the Authorization header.

I added a `--kerberos` flag to both `bbs2gh generate-script` and `bbs2gh migrate-repo`. For now this is a hidden option.

- [x] Issue linked

### DIFF
--- a/src/Octoshift/BbsClient.cs
+++ b/src/Octoshift/BbsClient.cs
@@ -14,7 +14,17 @@ public class BbsClient
     private readonly OctoLogger _log;
     private readonly RetryPolicy _retryPolicy;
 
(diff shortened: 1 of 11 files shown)


ArinGhazarian (approved):
Looks good.

PR #696, opened 2022-10-24
BBS Integration tests

### LINKED ISSUE #553: BBS Migration integration test(s)
We'll want some e2e tests that hit a real BBS environment and invoke the CLI the same way a customer would. Just like the ADO and GH tests, we'll want to have some code that "resets" all the data in the BBS environment to a known state, run the migrations, then check that the final state in github is what we expect.

We should probably do this against at least 3x different BBS environments, one with the latest version, one with the oldest version we support (5.14), and one running on a windows server.

## Tasks

- [x] #569
- [x] #571
- [x] #573
- [x] #718

dylan-smith:
Stood up a BBS server in our shared Azure Subscription (e2e-bbs-8-5-0-linux-2204.eastus.cloudapp.azure.com). It's running the latest version of BBS (8.5.0) and Ubuntu 22.04.

Manually created 2 projects and 2 repos in BBS (in the future we may automate this like we have for ADO/GH), and added a new integration test to migrate BBS->GH.

Included asserts to check if the GH repo exists and is initialized.

The default SSH cert that Azure generates didn't seem to work when using Ubuntu 22.04 (didn't work with our SSH.Net nuget library, but did work when directly using SSH in a terminal). I found a [workaround in an issue](https://github.com/sshnet/SSH.NET/issues/825#issuecomment-1139440419) from the SSH.Net repo that I applied in this PR. I don't fully understand exactly what it's doing, but it looks like it's updating the signature on the key in-memory before using it. Whatever it's doing, it makes keys that previously didn't work now work. So it seems like a good thing to me.

Closes #553 

Future work:
- automate the creation of BBS test data (like we do for ADO/GH)
- Include downloading the migration logs (and asserting that they exist)
- Add BBS windows server (once we support SMB)
- Add AWS
- Add another BBS server with the oldest version we support (5.14 I think)

Checklist
- [x] Issue linked


Jamacna (approved):

src/bbs2gh/Services/BbsSshArchiveDownloader.cs
```diff
@@ -10,15 +11,28 @@ public sealed class BbsSshArchiveDownloader : IBbsArchiveDownloader, IDisposable
...
 #pragma warning disable CA2000 // Incorrectly flagged as a not-disposing error
     public BbsSshArchiveDownloader(OctoLogger log, FileSystemProvider fileSystemProvider, string host, string sshUser, string privateKeyFileFullPath, int sshPort = 22)
-    : this(log, fileSystemProvider, new SftpClient(host, sshPort, sshUser, new PrivateKeyFile(privateKeyFileFullPath)))
     {
+        _pkRsa = new PrivateKeyFile(privateKeyFileFullPath);
+        var newKey = RsaSha256Util.ConvertToKeyWithSha256Signature(_pkRsa);
+        RsaSha256Util.UpdatePrivateKeyFile(_pkRsa, newKey);
+        _authenticationMethodRsa = new PrivateKeyAuthenticationMethod(sshUser, _pkRsa);
+        var connection = new ConnectionInfo(host, sshPort, sshUser, _authenticationMethodRsa);
+        connection.HostKeyAlgorithms["rsa-sha2-256"] = data => new KeyHostAlgorithm("rsa-sha2-256", _rsaKey, data);
```
ArinGhazarian, line 30: I just ran this on my local with a `ed25519` key (which IMO should be our recommended one b/c its a newer key format and is recommended probably anywhere else over RSA, I did a quick research, [see this](https://medium.com/risan/upgrade-your-ssh-key-to-ed25519-c6e8d60d3c54)) and it crashed. The reason is that we are assuming that the key is in RSA format and trying to convert it. While a quick workaround would be to check the key format and skip the conversion if it's not RSA I really don't like this approach for the following reasons:
1. it's too hack-y and messy.
2. IMO it's a **huge red flag** in terms of security risk. The reason is, as you also mentioned, we don't have a good understanding of what the workaround is doing, we're happy b/c it works but imagine if there's a security hole in the code, we are probably never going to find out about it and our code will suffer the security issue even if we download the latest nuget package because we took an alternate route!

My suggestion is to just remove the code and for our INT tests use `ed25519` key format that works. Later on for customers we can ask them to create new `ed25519` keys just for migration if RSA doesn't work.

src/bbs2gh/RsaSha256Util.cs
```diff
@@ -0,0 +1,101 @@
...
+using Renci.SshNet.Security;
+using Renci.SshNet.Security.Cryptography;
+using Renci.SshNet.Security.Cryptography.Ciphers;
+
+namespace OctoshiftCLI.BbsToGithub;
+
+// workaround for RSA keys on Ubuntu 22.04
+// https://github.com/sshnet/SSH.NET/issues/825#issuecomment-1139440419
+
+public class RsaWithSha256SignatureKey : RsaKey
```
ArinGhazarian, line 15: As I already mentioned this seems like a quick workaround rather than a production ready code and really hard to maintain! If we really want to go this route (I wouldn't recommend it) we need to spend some time and understand  what this is doing and probably re-write it and add unit tests, etc.

ArinGhazarian (changes requested):
The INT tests look really good. My main problem is the hack for supporting RSA keys which I think is a reg flag.

src/bbs2gh/Services/BbsSshArchiveDownloader.cs
dylan-smith: One of my big worries is support for the various key formats that are in use by our customers. Especially the more popular ones like RSA. But my bigger immediate concern with ed25519 is it doesn't look like it's an option to use that with Azure VM's at all:

image

https://learn.microsoft.com/en-us/azure/virtual-machines/linux/mac-create-ssh-keys

ArinGhazarian: Well this code as is doesn't support `ed25519` format. At the very least we should check for key type before doing the conversion. But I am more concerned about the security aspects. Like I said if we are going to go this route then we need to be able to fully understand what that conversion is doing and then add some tests for the class. IMO the way it is right now is not production ready at all.

ArinGhazarian: To unblock the INT tests, once possible solution would be to create another internal constructor overload for the `BbsSshArchiveDownloder` that accepts a `ConnctionInfo` and then instead of creating the connection inside the ctor, create it in the INT tests. This way at least we will minimize the risk and use the conversion code only in the INT tests.

dylan-smith: I like that idea as a way to make some quick progress.

src/bbs2gh/RsaSha256Util.cs
dylan-smith: I re-wrote it to be a bit simpler (IMO), but more importantly it only applies this behavior for ssh key's that are RSA (i.e. it won't break ed25519 or other key formats).

I thought about how we could unit test it, and ended up at the opinion that the amount of effort to unit test this chunk of code isn't worth it (it would be quite difficult, and probably require file system access). Hopefully at some point in the near future SSH.Net will release a new version and make this code obsolete.

src/bbs2gh/Services/BbsSshArchiveDownloader.cs
dylan-smith: I spent some time getting more familiar with exactly what's going on here, and simplified the implementation a bit (and should have fixed other key types not working).

The gist of what it's trying to do is basically modifying this one line of code inside SSH.Net:
https://github.com/sshnet/SSH.NET/blob/b9bc4750ab147948d66678af7e583878ce96f28a/src/Renci.SshNet/Security/Cryptography/RsaDigitalSignature.cs#L23

to use SHA256 instead of SHA1. It does this by creating a new DigitalSignature subclass that uses SHA256, then uses a bit of reflection to inject that into our PrivateKeyFile and ConnectionInfo objects.

ArinGhazarian (approved):
This looks good as far as I can see.

PR #698, opened 2022-10-25
Add AWS S3 support to BBS generate script

### LINKED ISSUE #693: Validate that either AWS or Azure credentials are provided
In `bbs2gh` (and possibly `gei` also) if the user doesn't pass either Azure or AWS credentials the CLI will crash with a NullReferenceException.

image

We should explicitly validate that they provided valid arguments and if not fail early with a useful error message.

ArinGhazarian:
Closes #693 

This PR fixes/adds the following:

1. Add `--aws-bucket-name` option to `bbs2gh generate-script` command to support AWS S3 upload. 
2. Fix a bug in `bbs2gh` where not providing Azure or AWS creds would cause the CLI to crash by adding more validations to `bbs2gh migrate-repo` command. 

- [x] Issue linked


RELEASENOTES.md
```diff
@@ -1,2 +1,4 @@
 - Added additional retry logic covering the case when polling for migration status fails for any reason (along with a few other situations)
-- Added `--aws-bucket-name` to `gh gei generate-script` and removed `--azure-storage-connection-string`. 
\ No newline at end of file
+- Added `--aws-bucket-name` to `gh gei generate-script` and removed `--azure-storage-connection-string`.
+- Added `--aws-bucket-name` to `bbs2gh generate-script`. 
+- Fixed a bug where not providing Azure Storage or AWS S3 connection info would cause `bbs2gh` to crash.
```
dylan-smith, line 4:
```suggestion
```

we're not mentioning bbs2gh in release note yet

dylan-smith:
We need to make sure we continue to support these 3-4 migrate-repo usage scenarios:

1. do all the things - gen, ssh, azure/aws, ingest
2. ingest only - ingest
3. local file - azure/aws, ingest
4. gen only - gen

definately (2) is broke with this PR because it forces the user to provide either azure/aws args

ArinGhazarian:
> We need to make sure we continue to support these 3-4 migrate-repo usage scenarios:
> 
> 1. do all the things - gen, ssh, azure/aws, ingest
> 2. ingest only - ingest
> 3. local file - azure/aws, ingest
> 4. gen only - gen
> 
> definately (2) is broke with this PR because it forces the user to provide either azure/aws args

- missing tests were added to cover all possible scenarios.
- To address no 2 (ingest only scenario), and extra if clause is added that check the archive url value and if present it won't validate the upload related args.
- Currently we don't support a gen only scenario.

dylan-smith:
What about if the user provides an archive path instead of a url? also what about the scenario where the user runs the CLI on the BBS server.

ArinGhazarian:
> What about if the user provides an archive path instead of a url? also what about the scenario where the user runs the CLI on the BBS server.

Both of those scenarios are already covered and there are unit tests for them:
- `Uses_Archive_Path_If_Provided` --> archive path is provided
- `It_Sets_The_Archive_Path_To_Default_Local_Path_When_Archive_Path_And_Archive_Url_Are_Not_Provided` --> running the CLI on the server where no archive path or archive url is provided.

That being said we still need some extra validations, like for example in some scenarios, if `--github-org` or `--github-repo` is not provided the code is gonna crash at some point . We need to validate them too.

dylan-smith (approved):

Issue #699, opened 2022-10-25
Output information about the progress of org-to-org migrations


timrogers:
When waiting for an organization-to-organization migration with the `wait-for-migration` command, the only feedback that the CLI provides on the progress of the migration is its `state`. 

Org-to-org migrations potentially stay in the `REPO_MIGRATION` `state` for a long time if there are many repos to be migrated, and this can be a worrying time - as a user, you are likely to wonder "is my migration actually moving forward"?.

The `node` query for `OrganizationMigration`s allows you to fetch the `remainingRepositoriesCount` and `totalRepositoriesCount` for the migration. We should grab these and output them as part of the logging in the CLI.

PR #700, opened 2022-10-25
Output information about the progress of org-to-org migrations as repos are migrated

timrogers:
When waiting for an organization-to-organization migration with the `wait-for-migration` command, the only feedback that the CLI provides on the progress of the migration is its `state`.

Org-to-org migrations potentially stay in the `REPO_MIGRATION` `state` for a long time if there are many repos to be migrated, and this can be a worrying time - as a user, you are likely to wonder "is my migration actually moving forward"?.

The `node` query for `OrganizationMigration`s allows you to fetch the `remainingRepositoriesCount` and `totalRepositoriesCount` for the migration.

This commit starts grabbing that data and logging a progress indicator (`1/10 repos migrated`) to the console.

Note that the GraphQL API currently returns `0` for the two count fields if the number of repos has not yet been completed. There is an outstanding bug to fix this and return `null` instead. This code is future-proof, able to handle both values.

Fixes https://github.com/github/gh-gei/issues/699.

_I don't intend to make it a habit to make code changes, but I wanted to see this enhancement added before org-to-org ships, and I wanted to get familiar with the codebase 😊 _

- [x] Issue linked

### DIFF
--- a/src/Octoshift/GithubApi.cs
+++ b/src/Octoshift/GithubApi.cs
@@ -375,12 +375,12 @@ namespace OctoshiftCLI
             return (string)data["data"]["startOrganizationMigration"]["orgMigration"]["id"];
         }
 
-        public virtual async Task<(string State, string SourceOrgUrl, string TargetOrgName, string FailureReason)> GetOrganizationMigration(string migrationId)
+        public virtual async Task<(string State, string SourceOrgUrl, string TargetOrgName, string FailureReason, int? RemainingRepositoriesCount, int? TotalRepositoriesCount)> GetOrganizationMigration(string migrationId)
         {
             var url = $"{_apiUrl}/graphql";
 
             var query = "query($id: ID!)";
-            var gql = "node(id: $id) { ... on OrganizationMigration { state, sourceOrgUrl, targetOrgName, failureReason } }";
+            var gql = "node(id: $id) { ... on OrganizationMigration { state, sourceOrgUrl, targetOrgName, failureReason, remainingRepositoriesCount, totalRepositoriesCount } }";
 
             var payload = new { query = $"{query} {{ {gql} }}", variables = new { id = migrationId } };
 
@@ -395,7 +395,9 @@ namespace OctoshiftCLI
                     State: (string)data["data"]["node"]["state"],
                     SourceOrgUrl: (string)data["data"]["node"]["sourceOrgUrl"],
                     TargetOrgName: (string)data["data"]["node"]["targetOrgName"],
-                    FailureReason: (string)data["data"]["node"]["failureReason"]);
+                    FailureReason: (string)data["data"]["node"]["failureReason"],
+                    RemainingRepositoriesCount: (int?)data["data"]["node"]["remainingRepositoriesCount"],
+                    TotalRepositoriesCount: (int?)data["data"]["node"]["totalRepositoriesCount"]);
             });
 
             return response.Outcome == OutcomeType.Failure
--- a/src/Octoshift/Handlers/WaitForMigrationCommandHandler.cs
+++ b/src/Octoshift/Handlers/WaitForMigrationCommandHandler.cs
@@ -76,15 +76,14 @@ public class WaitForMigrationCommandHandler : ICommandHandler<WaitForMigrationCo
                 throw new OctoshiftCliException($"Migration {migrationId} failed for {sourceOrgUrl} -> {targetOrgName}. Failure reason: {failureReason}");
             }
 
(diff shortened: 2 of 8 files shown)


src/gei/Handlers/MigrateOrgCommandHandler.cs
```diff
@@ -50,13 +50,21 @@ public async Task Handle(MigrateOrgCommandArgs args)
...
-        var (migrationState, _, _, failureReason) = await _githubApi.GetOrganizationMigration(migrationId);
+        var (migrationState, _, _, failureReason, completedRepositoriesCount, totalRepositoriesCount) = await _githubApi.GetOrganizationMigration(migrationId);
 
         while (OrganizationMigrationStatus.IsPending(migrationState))
         {
-            _log.LogInformation($"Migration in progress (ID: {migrationId}). State: {migrationState}. Waiting 10 seconds...");
+            if (OrganizationMigrationStatus.IsRepoMigration(migrationState))
+            {
+                var remainingRepositoriesCount = (int)totalRepositoriesCount - (int)completedRepositoriesCount;
+                _log.LogInformation($"Migration in progress (ID: {migrationId}). State: {migrationState}. {remainingRepositoriesCount}/{totalRepositoriesCount} repo(s) migrated. Waiting 10 seconds...");
```
dylan-smith, line 60: this logic or variable names is confusing. shouldn't the message be {completed}/{total} not {remaining}/{total} ?

timrogers: You're totally right. My brain wasn't working when I wrote this PR. 

The API returns remaining and total and we want to present it as `completed/remaining`. I'll fix this and request a re-review.

timrogers: I just misnamed the variable returned from the API in this method and then went from there 😂

dylan-smith (changes requested):
looks good, could we get a test for the null scenario?

timrogers:
> looks good, could we get a test for the null scenario?

The null scenario shouldn't be relevant now, as we're just relying on the statuses instead.

dylan-smith (approved):

src/Octoshift/GithubApi.cs
```diff
@@ -375,12 +375,12 @@ public virtual async Task<string> StartOrganizationMigration(string sourceOrgUrl
             return (string)data["data"]["startOrganizationMigration"]["orgMigration"]["id"];
         }
 
-        public virtual async Task<(string State, string SourceOrgUrl, string TargetOrgName, string FailureReason)> GetOrganizationMigration(string migrationId)
+        public virtual async Task<(string State, string SourceOrgUrl, string TargetOrgName, string FailureReason, int? RemainingRepositoriesCount, int? TotalRepositoriesCount)> GetOrganizationMigration(string migrationId)
```
ArinGhazarian, line 378: In what scenario the remaining repo count and total repo could would be null? can we just rely on zero as an exceptional case and work with that?

src/Octoshift/Handlers/WaitForMigrationCommandHandler.cs
```diff
@@ -76,11 +76,19 @@ private async Task WaitForOrgMigration(string migrationId, string githubPat, Git
                 throw new OctoshiftCliException($"Migration {migrationId} failed for {sourceOrgUrl} -> {targetOrgName}. Failure reason: {failureReason}");
             }
 
-            _log.LogInformation($"Migration {migrationId} is {state}");
+            if (OrganizationMigrationStatus.IsRepoMigration(state))
+            {
+                var completedRepositoriesCount = (int)totalRepositoriesCount - (int)remainingRepositoriesCount;
```
ArinGhazarian, line 81: since both total repo count and remaining repo count can be null, there's a potential of an exception here. So either we need to guard against that or just don't use null values.

dylan-smith: I did write (and then delete) a unit test with 2 nulls and it didn't crash (I had the same thought)

src/Octoshift/GithubApi.cs
timrogers: We discussed this on Slack, but just adding an answer here to for the record.

At the moment, the API returns `0` when the count has not yet been computed. I have filed a bug, because I think we should change it to return `null` in this case. This code is designed to support both cases.

src/Octoshift/Handlers/WaitForMigrationCommandHandler.cs
timrogers: It is guaranteed that when the repo is in the `REPO_MIGRATION` state - as is tested by the `if` condition - the counts will be present.

Issue #710, opened 2022-11-01
No help for migrate-org command


robandpdx:
## Description
Running `gh gei migrate-org --help` produces no output. I expect to see a message similar to what I see when running `gh gei migrate-repo --help`.

## Reproduction Steps
Run `gh gei migrate-org --help` on the command line. Observe no useful output message.

dylan-smith:
good catch! that command has been present in the CLI for a while now but "hidden". With org-to-org migrations now being more widely available we should change that.

PR #719, opened 2022-11-02
Remove hidden from migrate org command

juanmrad:
Fixes: https://github.com/github/gh-gei/issues/710

- [X] Issue linked
- [x] Docs updated (or issue created) https://docs.github.com/en/early-access/enterprise-importer/migrating-organizations-with-github-enterprise-importer/migrating-organizations-from-githubcom-to-github-enterprise-cloud


dylan-smith (approved):

dylan-smith:
there were 2 integration test failures here, the first one I took a look at and it looked liked it would have been solved by a CLI retry when talking to ADO ([Issue created](https://github.com/github/gh-gei/issues/720)).

I reran the build job, but this time it failed with what looks like a backend error. We've been trying really hard not to ignore integration test failure and track down what happened and file an appropriate issue to fix when appropriate. We used to file an on-call issue to investigate, but got some feedback in retro that it would be better if we did the investigation rather than relying on on-call.

Can you please do that here. Looks like the migration failed with some kind of SSL error. It's almost certainly transient (because the other integration test jobs succeeded), but we should file a bug to fix whatever needs to be fixed to prevent it.

dylan-smith:
Updated release notes looks better. :shipit:

PR #723, opened 2022-11-03
Add support to `bbs2gh migrate-repo` to only generate the archive on the BBS server

### LINKED ISSUE #729: Add support for splitting archive generation and processing
Add support for calling bbs2gh migrate-repo with a combination of args that will cause it to tell BBS to generate the archive and do nothing else.

This can be useful in a few scenarios, but for right now the goal is to support a customer that uses different proxy servers depending on if they are talking to an internal server (i.e. BBS) or an external server (i.e. Azure/GitHub). With this change you will be able to call migrate-repo once to trigger the BBS archive generation, then switch up your proxy settings and call migrate-repo a 2nd time to tell it to upload to azure and call github.com.

dylan-smith:
Closes #729 

This adds support for calling `bbs2gh migrate-repo` with a combination of args that will cause it to tell BBS to generate the archive and do nothing else.

This can be useful in a few scenarios, but for right now the goal is to support a customer that uses different proxy servers depending on if they are talking to an internal server (i.e. BBS) or an external server (i.e. Azure/GitHub). With this change you will be able to call `migrate-repo` once to trigger the BBS archive generation, then switch up your proxy settings and call `migrate-repo` a 2nd time to tell it to upload to azure and call github.com.

- [x] Issue linked


timrogers:
I'm wondering if this command might get hard to reason about with different flows implicitly triggered by the presence of different arguments. 

Validation and errors become much more complicated. 

I think there are some other ways we could potentially structure this, e.g. a discrete `generate-repo-archive` command or an explicit `--only-generate-archive` argument. 

I haven't deeply thought through the different options, but I just wanted to ask the question and make sure we get the right CLI UX for this.

timrogers:
@dylan-smith What do you think?

I don't want to slow down this getting merged given that we have a customer waiting and we can easily make breaking changes on BBS. But I do think these questions are important to ask!

dylan-smith:
> I'm wondering if this command might get hard to reason about with different flows implicitly triggered by the presence of different arguments.
> 
> Validation and errors become much more complicated.
> 
> I think there are some other ways we could potentially structure this, e.g. a discrete `generate-repo-archive` command or an explicit `--only-generate-archive` argument.
> 
> I haven't deeply thought through the different options, but I just wanted to ask the question and make sure we get the right CLI UX for this.

We've talked about this within the team several times. Specifically whether we should have one migrate-repo command with many args and different ways of using it vs splitting it into multiple commands. I think there's 2 consequences of having a complex command like this:
1. Makes the code complex and potentially more error-prone and difficult to maintain
2. Potentially make the user experience more confusing than having multiple simpler commands

For 1, I feel like we have a pretty good set of implementation patterns that makes this complexity manageable from a coding perspective. And we have even more refactors coming that will improve it even further (specifically moving validation to the Args class).

For 2 I bet if you asked 10 people you'd probably get 10 different opinions which way was best. I don't really feel strongly about it, but my opinion is one command makes more sense. I feel like it's pretty common for CLI's to have commands with many args and different ways of calling it. Just look at pretty much any `git` command.

I think the key is having good documentation - and that will apply no matter which implementation approach we take. I do think that the built-in CLI help could benefit from putting a few examples in there, I think our CLI help library we're using supports that.

src/bbs2gh/Handlers/MigrateRepoCommandHandler.cs
```diff
@@ -369,15 +387,17 @@ private void ValidateDownloadOptions(MigrateRepoCommandArgs args)
 
         if (shouldUseSsh && shouldUseSmb)
         {
-            throw new OctoshiftCliException("Either SSH or SMB download options can be specified.");
+            throw new OctoshiftCliException("You can't provide both SSH and SMB credentials together.");
         }
 
-        if (shouldUseSsh && sshArgs.Any(arg => arg.IsNullOrWhiteSpace()))
+        if ((args.SshUser.HasValue() && args.SshPrivateKey.IsNullOrWhiteSpace()) ||
+            (args.SshPrivateKey.HasValue() && args.SshUser.IsNullOrWhiteSpace()))
```
ArinGhazarian, line 394: We can use XOR here:
```suggestion
        if (args.SshUser.HasValue() ^ args.SshPrivateKey.HasValue())
```

src/bbs2gh/Handlers/MigrateRepoCommandHandler.cs
```diff
@@ -369,15 +387,17 @@ private void ValidateDownloadOptions(MigrateRepoCommandArgs args)
...
         {
             throw new OctoshiftCliException("Both --ssh-user and --ssh-private-key must be specified for SSH download.");
         }
 
-        if (shouldUseSmb && smbArgs.Any(arg => arg.IsNullOrWhiteSpace()))
+        if ((args.SmbUser.HasValue() && args.SmbPassword.IsNullOrWhiteSpace()) ||
+            (args.SmbPassword.HasValue() && args.SmbUser.IsNullOrWhiteSpace()))
```
ArinGhazarian, line 400: Same as above we can use XOR here:
```suggestion
        if (args.SmbUser.HasValue() ^ args.SmbPassword.HasValue())
```

ArinGhazarian (approved):
Looks good, I just left a couple of non blocking comments.

PR #725, opened 2022-11-03
Added: support for session tokens in the AWSAPI factory

mattbryce93:
- [x ] Issue linked

### DIFF
--- a/src/Octoshift/AwsApi.cs
+++ b/src/Octoshift/AwsApi.cs
@@ -17,12 +17,10 @@ public class AwsApi : IDisposable
 
 #pragma warning disable CA2000
     public AwsApi(string awsAccessKey, string awsSecretKey, string awsSessionToken = null) : this(new TransferUtility( 
-        if (awsSessionToken == null) 
-        { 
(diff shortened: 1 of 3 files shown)


dylan-smith:
Hi @mattbryce93, thanks for trying to contribute! Unfortunately, it looks like your PR needs a lot more work before we could consider accepting it (e.g. it doesn't compile, doesn't pass the linter, I suspect it breaks existing scenarios, and is missing a bunch of things needed to support the STS scenario such as new args).

My best suggestion would be to find a friend that knows a bit more c# to help out, or wait until this backlog item raises to the top of our teams priority list and we can help tackle it.

peteraritchie:
@mattbryce93 is there some code you've added to `EnvironmentVariableProvider` for the `AwsSessionToken()` method you're using in `AwsApiFactory`?

dylan-smith:
FYI there are some discussions going on internally about moving the blob storage configuration into GHES itself instead of passing it into the CLI. If that happens this would become kind of a moot point.

ArinGhazarian:
Closing since this is already done in #844

PR #730, opened 2022-11-06
Refactor CommandArgs to Validate and Log

dylan-smith:
This PR does a few things:
- Moves GenerateScriptCommandArgs to it's own file
- Adds a custom attribute `[LogName(string)]` to specify a user-friendly string to use when logging an Arg property
- Adds a custom attribute `[Secret]` to indicate that an Arg property is a secret and should be obfuscated from any logs
- Adds an abstract base class `BaseCommandArgs` which implements a `Log()` function using reflection, and a `RegisterSecrets()` function also using reflection
- BaseCommandArgs has an abstract `Validate()` function which needs to be overridden in each Args class
- Moved the arg validation logic from `GenerateScriptCommandHandler` to `GenerateScriptCommandArgs.Validate()`
- Moved the arg validation unit tests into their own test file
- Added the Verbose arg and the code to apply it into the `BaseCommandArgs` class
- Updated the code that wires up commands and handlers (in `CommandExtensions`) to always call the `RegisterSecrets()` `Log()` and `Validate()` functions on the args instance

Checklist
- [ ] Issue linked


src/gei/CommandArgs/GenerateScriptCommandArgs.cs
```diff
@@ -0,0 +1,79 @@
+﻿using System.IO;
+using OctoshiftCLI.Extensions;
+
+namespace OctoshiftCLI.GithubEnterpriseImporter.CommandArgs;
+
+public class GenerateScriptCommandArgs : BaseCommandArgs
+{
+    [LogName("GITHUB SOURCE ORG")]
```
ArinGhazarian, line 8: I think for almost all of the. properties we won't need the `LogName` attribute, we can simply write a code to automatically covert the property name to a log friendly name. All that method needs to do is to split the property name by capital letters (except the very first one) then join them with a space and `ToUpper()` the end result.

src/gei/CommandArgs/GenerateScriptCommandArgs.cs
```diff
@@ -0,0 +1,79 @@
...
+    public bool LockSourceRepo { get; set; }
+
+    [LogName("DOWNLOAD MIGRATION LOGS")]
+    public bool DownloadMigrationLogs { get; set; }
+
+    [LogName("SEQUENTIAL")]
+    public bool Sequential { get; set; }
+
+    [LogName("GITHUB SOURCE PAT")]
+    [Secret]
```
ArinGhazarian, line 48: Like the idea of `Secret` attribute 👍

src/Octoshift/BaseCommandArgs.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            if (log is null)
+            {
+                throw new ArgumentNullException(nameof(log));
+            }
+
+            log.Verbose = Verbose;
+            var sb = new StringBuilder();
+
+            foreach (var property in GetType().GetProperties())
```
ArinGhazarian, line 24: Might be better to get all `public` properties

src/Octoshift/BaseCommandArgs.cs
```diff
@@ -0,0 +1,57 @@
...
+            log.LogInformation(sb.ToString());
+        }
+
+        public void RegisterSecrets(OctoLogger log)
+        {
+            foreach (var property in GetType().GetProperties())
+            {
+                if (property.GetCustomAttributes(typeof(SecretAttribute), true).Any())
+                {
+                    log?.RegisterSecret((string)property.GetValue(this));
```
ArinGhazarian, line 52: I guess like the `Log` method it'd be better to add a null guard for log and throw if it's null.

src/Octoshift/BaseCommandArgs.cs
```diff
@@ -0,0 +1,57 @@
+﻿using System;
+using System.Linq;
+using System.Text;
+
+namespace OctoshiftCLI
+{
+    public abstract class BaseCommandArgs
```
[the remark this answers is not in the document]
dylan-smith: I did what I think is even better and renamed it to just CommandArgs. 

I rebased this branch on top of the one that does the folder renames, this way there isn't an awkward folder/namespace collision between a class and namespace with the same name anymore allowing me to do this.

src/Octoshift/BaseCommandArgs.cs
dylan-smith: the default behavior of GetProperties is to only return public properties

src/gei/CommandArgs/GenerateScriptCommandArgs.cs
dylan-smith: done

src/Octoshift/BaseCommandArgs.cs
dylan-smith: done

Issue #570, opened 2022-08-15
BBS Integrations tests for 5.14


dylan-smith:
- [x] Setup BBS 5.14 test server in Azure
- [x] Update BBS integration test to run against both 8.50 and 5.14

PR #733, opened 2022-11-09
Added BBS environment reset code for e2e tests

### LINKED ISSUE #573: Implement environment reset code

dylan-smith:
Added code to do a similar thing we do with our other e2e tests, where the test is responsible for resetting the test data that it expects in the test environment. This makes it easy to stand up new environments, and prevent accidentally breaking test environments by unintentional manual changes in the BBS server(s). It also makes it easy to add new data shapes to our test environments, as we just add them in one place (in the code) rather than having to manually touch each of our BBS test servers (and we will have several of them before we're done with the BBS work).

Right now these additions will check if the BBS project exists, and if so iterate and delete all repos in the project, then delete the project. Then create the project(s) and repo(s) from scratch.

It uses libgit2sharp to initialize the repos. This dependency only exists in our e2e test code, not in the CLI that we ship to customers. libgit2sharp comes with native binaries for various linux distro's, but only includes ubuntu 18.04 but our build servers use ubuntu 20.04 (or maybe 22.04). The workaround is to use the LD_LIBRARY_PATH env var when building our integration test project to force it to link to the ubuntu 18.04 binary (https://github.com/libgit2/libgit2sharp/issues/1703#issuecomment-619497921)

This has pretty much everything we need to support testing against BBS 5.14 also (#570 ) except for some reason the initialize repo code fails consistently when run against BBS 5.14. Once somebody figures out why that fails, you just need to add an extra [InlineData] attribute to the test case and it will run for both.

Closes #573 

- [x] Issue linked

### DIFF
--- a/src/OctoshiftCLI.IntegrationTests/BbsToGithub.cs
+++ b/src/OctoshiftCLI.IntegrationTests/BbsToGithub.cs
@@ -70,7 +70,7 @@ public sealed class BbsToGithub : IDisposable
     [Fact]
     public async Task Basic()
     {
-        var bbsProjectKey = "E2E-L";
+        var bbsProjectKey = $"E2E-{TestHelper.GetOsName().ToUpper()}";
         var githubTargetOrg = $"e2e-testing-{TestHelper.GetOsName()}";
         var repo1 = $"{bbsProjectKey}-repo-1";
         var repo2 = $"{bbsProjectKey}-repo-2";
--- a/src/OctoshiftCLI.IntegrationTests/BbsToGithub.cs
+++ b/src/OctoshiftCLI.IntegrationTests/BbsToGithub.cs
@@ -65,7 +65,7 @@ public sealed class BbsToGithub : IDisposable
     }
 
     [Theory]
-    [InlineData("http://e2e-bbs-8-5-0-linux-2204.eastus.cloudapp.azure.com:7990")]
+    //[InlineData("http://e2e-bbs-8-5-0-linux-2204.eastus.cloudapp.azure.com:7990")]
     [InlineData("http://e2e-bbs-5-14-0-linux-2004.eastus.cloudapp.azure.com:7990")]
     public async Task Basic(string bbsServer)
     {
--- a/src/OctoshiftCLI.IntegrationTests/BbsToGithub.cs
+++ b/src/OctoshiftCLI.IntegrationTests/BbsToGithub.cs
@@ -66,6 +66,7 @@ public sealed class BbsToGithub : IDisposable
 
     [Theory]
     [InlineData("http://e2e-bbs-8-5-0-linux-2204.eastus.cloudapp.azure.com:7990")]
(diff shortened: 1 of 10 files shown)


src/OctoshiftCLI.IntegrationTests/BbsToGithub.cs
```diff
@@ -54,35 +53,46 @@ public BbsToGithub(ITestOutputHelper output)
...
         _blobServiceClient = new BlobServiceClient(azureStorageConnectionString);
 
-        _sourceHelper = new TestHelper(_output, _sourceBbsApi, _sourceBbsClient, BBS_URL);
         _targetHelper = new TestHelper(_output, _targetGithubApi, _targetGithubClient, _blobServiceClient);
     }
 
-    [Fact]
-    public async Task Basic()
+    [Theory]
+    [InlineData("http://e2e-bbs-8-5-0-linux-2204.eastus.cloudapp.azure.com:7990")]
```
synthead, line 68: Can we add an issue to make the server use HTTPS?  We should probably rotate the credentials as part of this effort, too :wink:

src/OctoshiftCLI.IntegrationTests/TestHelper.cs
```diff
@@ -113,6 +156,60 @@ public async Task ResetGithubTestEnvironment(string githubOrg)
...
+
+            Directory.CreateDirectory(repoPath);
+
+            Repository.Init(repoPath);
+            using var repo = new Repository(repoPath);
+            File.WriteAllText(Path.Join(repo.Info.WorkingDirectory, "README.md"), "# Test Repo");
+            repo.Index.Add("README.md");
+            repo.Index.Write();
+            var author = new Signature("Octoshift", "octoshift@github.com", DateTime.Now);
+            repo.Commit("Here's a commit i made!", author, author);
```
synthead, line 176: We spoke about how you're interested in also creating PRs and such with the BBS API sometime in the future.  With this code, a unique commit ID will be created every time it is executed, which will also mean that we will need to use unique payloads to create the PRs.

We could use a Git fixture (that could even be stored as a repo in GitHub), but we could also go the full nine yards and use Atlassian's own migration tools to store test data :smile: This strategy would use the same archives that we're consuming in Octoshift, and we can use their first-party tools to do the work.

This has the potential to save us a lot of time in creating our test data.  In addition, we wouldn't need to use Git libraries, and it would reduce network requests made to set up the data.  We could also use [our own fixtures](https://github.com/github/octoshift/tree/129b1f77bda609df9c5bbc9e51d1ffba9c80ad4c/spec/fixtures/bitbucket_server) as test data, too.

Here is Atlassian's doc for migrating from and to BBS servers with migration archives: https://confluence.atlassian.com/bitbucketserver/migrating-bitbucket-server-to-another-server-776640406.html

This is certainly up for conversation and should definitely not be addressed in this PR, but I think there would be value in creating an issue to track this possibility :+1:

synthead (approved):
LGTM!  I added a couple comments for things we could improve on with over efforts of work :upside_down_face:

Issue #703, opened 2022-10-26
Move validations into the arg class


ArinGhazarian:
Currently we perform the arg validations inside the command handler class sometimes right inside the handle method or in its own dedicated method (or more than one method even). 

The problems with this approach are: 

1. We do the validations after we create the handler which is not ideal. We shouldn't even create the handler if the validation fails. 
2. There is no uniform pattern for validations. Sometimes we do them in a single place or sometimes they are scattered in a handler class inside different methods which makes it harder to keep a track and make changes, also there is a possibility of missing some. 

## Todo
- [ ] Create a `Validate` method inside each arg class (it can come from a base arg class and overridden in the arg class) and put all arg validations in it.
- [ ] Remove the validations from the command handler classes and call the `Validate` method in the command class as part of the `BuildHandler` method.
- [ ] Rethink about arg validations case by case specially for larger commands like `migrate-repo` or `generate-script`.

PR #735, opened 2022-11-11
Restructure the folders to one per command

dylan-smith:
No code was meaningfully changed in this PR, it changes our folder structure so instead of folders for Commands/Handlers we now have a folder for each command that contains up to 3 cs files Command/CommandHandler/CommandArgs. As part of this  the xxx.Handlers namespaces went away and everything lives in the Commands namespace.

This PR also split all the CommandArgs classes out into their own files.

Want to get this PR merged in before I come back to the other PR that will refactor the validation to happen in the args classes. 

Both of which are part of the work for #703 

- [x] Issue linked


ArinGhazarian (changes requested):
By convention all namespaces should match the folder structure.

dylan-smith:
> By convention all namespaces should match the folder structure.

Normally I'd agree but I think this is an example where it makes more sense to go against convention. Putting an extra part in the namespace adds no value IMO, the namespace will become redundant with the class name (e.g. OctoshiftCLI.Commands.CreateTeam.CreateTeamCommand) and I wouldn't want to try to remove the name from the class name. Having the folder name match the actual CLI command-name is a useful bit of information when navigating the code structure, and I don't believe you can have hyphens in namespace names so we couldn't make the current folder names namespace names as-is.

ArinGhazarian:
> > By convention all namespaces should match the folder structure.
> 
> 
> 
> Normally I'd agree but I think this is an example where it makes more sense to go against convention. Putting an extra part in the namespace adds no value IMO, the namespace will become redundant with the class name (e.g. OctoshiftCLI.Commands.CreateTeam.CreateTeamCommand) and I wouldn't want to try to remove the name from the class name. Having the folder name match the actual CLI command-name is a useful bit of information when navigating the code structure, and I don't believe you can have hyphens in namespace names so we couldn't make the current folder names match namespace names as-is.

Well I understand your reasoning, but still think following the convention is better. I for example usually don't navigate using the tree explorer but rather do it with the class view explorer and with improper namespacing it'd be very confusing to see the big picture so if we are creating these folders it is going to be confusing to have different namespace structure. Everything else is just fine and it makes sense to keep the class names as is with the command name included. Also I probably haven't noticed but the folder names should also follow the naming guidelines so no hyphen in the names.

dylan-smith:
Superceded by #976

PR #592, opened 2022-08-23
Adding migration of GitHub Secret Scanning Alerts

peter-murray:
- [x] Did you write/update appropriate tests
- [x] Release notes updated (if appropriate)
- [x] Appropriate logging output
- [x] Issue linked
- [x] Docs updated (or issue created) (#715 )

This is the first pass on the requirements to migrate secret scanning remediations between two repositories, provides the functionality for #536.

This is being opened to start a code review and then complete the support of this command line and service. There are some necessary manual steps required as part of the larger migration in that the repo is migrated and secret scanning has been reactivated (and the backfill scan completed) before this command line makes sense to execute.

Created a new issue to add an integration test to cover this functionality in the future: #717

PR #741, opened 2022-11-21
Adding feature for code scanning migration

### LINKED ISSUE #921: Add retries to UploadSarifReport
It seems UploadSarifReport frequently gets a 502 when POSTing to the endpoint.

```
[4:52 PM] [DEBUG] RESPONSE (BadGateway): {
  "message": "Server Error"
}
```

Especially if you're running `migrate-code-scanning-alerts` on a repo with many analysis to migrate (e.g. this repo has 431 analysis at this time).

When this happens you can usually just re-run the `migrate-code-scanning-alerts` command and it picks up where it left off, skipping all the analyses that have already been uploaded.

It would be better instead to retry the API call that fails with the 502. Could scope the retries down to only retry on 502.

davelosert:
- [x] Issue linked

This PR provides the functionality for #535 (I had to wait for @peter-murray 's changes in  #592 as I was building on top of them).

I yet have to create a documentation issue - but knowing that this is a lot of code, I wanted to put up the PR to already get some feedback early.

I already announced this PR to @dylan-smith through Slack.

Also fixes #921


gateixeira:
Hi team!

is anything holding this PR?

looking forward to use this feature for a ongoing migration.

src/gei/Commands/MigrateCodeScanningAlertsCommand.cs
```diff
@@ -0,0 +1,110 @@
...
+using OctoshiftCLI.Extensions;
+using OctoshiftCLI.GithubEnterpriseImporter.Handlers;
+
+namespace OctoshiftCLI.GithubEnterpriseImporter.Commands;
+
+public class MigrateCodeScanningAlertsCommand : CommandBase<MigrateCodeScanningAlertsCommandArgs, MigrateCodeScanningAlertsCommandHandler>
+{
+    public MigrateCodeScanningAlertsCommand() : base(
+        name: "migrate-code-scanning-alerts",
+        description: "Migrates the state and resolution of code scanning alerts.")
```
timrogers, line 14: The release note contains a bit more detail on what this command does. I think it'd be a great idea to transpose that context to the help text, so it's a bit clearer what this does and what the user can expect.

src/Octoshift/CodeScanningAlertService.cs
```diff
@@ -0,0 +1,209 @@
...
+            var defaultBranch = await _sourceGithubApi.GetDefaultBranch(sourceOrg, sourceRepo);
+            _log.LogInformation($"Found default branch: {defaultBranch} - migrating code scanning alerts only of this branch.");
+            await MigrateAnalyses(sourceOrg, sourceRepo, targetOrg, targetRepo, defaultBranch, dryRun);
+            await MigrateAlerts(sourceOrg, sourceRepo, targetOrg, targetRepo, defaultBranch, dryRun);
+        }
+
+        protected internal virtual async Task MigrateAnalyses(string sourceOrg, string sourceRepo, string targetOrg,
+            string targetRepo, string branch, bool dryRun)
+        {
+            _log.LogInformation($"Migrating Code Scanning Analyses from '{sourceOrg}/{sourceRepo}' to '{targetOrg}/{targetRepo}'");
```
timrogers, line 37: I wonder if we should make the logging more detailed here so it's clear what GitHub instance we're migrating from and to - it might just make these logs friendlier to read.

src/Octoshift/CodeScanningAlertService.cs
```diff
@@ -0,0 +1,209 @@
...
+                try
+                {
+                    await _targetGithubApi.UploadSarifReport(targetOrg, targetRepo,
+                        new SarifContainer
+                        {
+                            Sarif = sarifReport,
+                            Ref = analysis.Ref,
+                            CommitSha = analysis.CommitSha
+                        });
+                    _log.LogInformation($"Successfully Migrated report for analysis {analysis.Id}");
```
timrogers, line 70:
```suggestion
                    _log.LogInformation($"Successfully migrated report for analysis {analysis.Id}");
```

src/Octoshift/CodeScanningAlertService.cs
```diff
@@ -0,0 +1,209 @@
...
+                    ++successCount;
+                }
+                catch (HttpRequestException httpException)
+                {
+                    if (httpException.StatusCode.Equals(HttpStatusCode.NotFound))
+                    {
+                        _log.LogVerbose($"No commit found on target. Skipping Analysis {analysis.Id}");
```
timrogers, line 77: Is a missing commit the only reason why this could 404? If we're not certain, then I think it would be better to be a bit less certain in this log.

timrogers, line 77:
```suggestion
                        _log.LogVerbose($"No commit found on target. Skipping analysis {analysis.Id}");
```

src/Octoshift/CodeScanningAlertService.cs
```diff
@@ -0,0 +1,209 @@
...
+                    }
+                    else
+                    {
+                        _log.LogWarning($"Http Error {httpException.StatusCode} while migrating analysis {analysis.Id}: ${httpException.Message}");
```
timrogers, line 81:
```suggestion
                        _log.LogWarning($"HTTP Error {httpException.StatusCode} while migrating analysis {analysis.Id}: ${httpException.Message}");
```

src/Octoshift/CodeScanningAlertService.cs
```diff
@@ -0,0 +1,209 @@
...
+                {
+                    _log.LogWarning($"Fatal Error while uploading SARIF report for analysis {analysis.Id}: \n {exception.Message}");
+                    _log.LogError(exception);
+                    // Todo Maybe throw another exception here?
+                    throw;
+                }
+                _log.LogInformation($"Handled {successCount + errorCount} / {analyses.Count()} Analyses.");
+            }
+
+            _log.LogInformation($"Code Scanning Analyses done!\nSuccess-Count: {successCount}\nError-Count: {errorCount}\nOverall: {analyses.Count()}.");
```
timrogers, line 95: If some of my analyses fail to migrate, what can I do? Is it safe to run this command again?

timrogers, line 95:
```suggestion
            _log.LogInformation($"Finished migrating Code Scanning analyses done! {successCount}/{analyses.Count()} migrated successfully");
```

timrogers (commented):
Thanks for taking the time to submit this contribution ❤️ , and apologies again that it has taken us such a long time to review it.

I've reviewed this from a product perspective, and I think it looks great! I've suggested a few changes to the log output, plus I've asked some questions about edge cases.

I'll leave @dylan-smith to review the code, which he is planning to do in the next couple of days.

src/Octoshift/CodeScanningAlertService.cs
davelosert: The Problem here is that I don't have that information in that particular context / class, as the class itself only gets a finished `sourceGithubApi` and `targetGithubApi` object (which hides the server information)

To be able to log the instance here, I guess I would either have to:

- Pass the value as an argument just for logging, which feels weird
- Make it accessible from the `GitHubApi`-class - which also feels a bit weird as I guess this is a `private` field there for a reason

Happy to do either, but I guess this is something that I wanted @dylan-smith to look at first and tell me if we can/should do this and how. 🙂

src/Octoshift/CodeScanningAlertService.cs
davelosert: Well spotted - a 404 can also occur if the target token is lacking permissions.

I added this to the log and also made it a `warn` rather than only when being in `verbose` mode.

This was originally in there when I was still allowing pull-request analysis to be migrated - there it could happen that commits were missing as, during an action run, merge-commits were created on the runner itself that never got pushed back to GitHub. In that case, a skipped analysis was no problem at all.

But not having it only for the main branch, it actually is.

src/Octoshift/CodeScanningAlertService.cs
davelosert: That is indeed a good question. Actually, as of right now, running this command multiple times is **not** safe, as the chronological order of sarif upload is important to the alert states.

E.g. if I have a repository with existing alerts from recent commits, and upload a SARIF Report that has no vulnerabilities to that repository, all alerts will be closed as `fixed` - even if the commit-sha of the SARIF Report itself is a very old one.

So, if an analysis upload fails, right now it would be most safe to find the cause, fix the issue but then remigrate the entire repository.

Let me see if I can add a function to check the target for already uploaded analysis and then just skip those - however. This will greatly increase the required API Calls.

src/gei/Handlers/MigrateCodeScanningAlertsCommandHandler.cs
```diff
@@ -0,0 +1,87 @@
...
+        LogAndValidateOptions(args);
+
+        await _codeScanningAlertService.MigrateCodeScanningAlerts(
+            args.SourceOrg,
+            args.SourceRepo,
+            args.TargetOrg,
+            args.TargetRepo,
+            args.DryRun);
+
+        _log.LogSuccess($"Secret scanning alerts successfully migrated.");
```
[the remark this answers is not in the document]
davelosert: Fixed it, will not be logged in dry run anymore!

src/gei/Commands/MigrateCodeScanningAlertsCommand.cs
davelosert: Added more context - is it better now?

dylan-smith:
I commited a bunch of changes, the main one is making the models more lightweight by removing a bunch of fields that we don't actually need in GEI.

However, when I was manually testing it, it was failing to migrate any alerts, saying it couldn't find a matching alert. @timrogers  mentioned he talked with you about a known race condition that might be the cause. Can you elaborate a bit if you know what I'm talking about?

davelosert:
@dylan-smith: Thank you very much for the help!

And yes. It is less of a race condition and more the realization that uploaded SARIF Reports are not immediately yielding alerts / analyses, but have a certain processing time. Due to this, running the alert-state-migration right after the analyses-migration can lead to what you experienced, in that the alerts on the target are not yet created and thus can't be matched and the state can't be migrated.

I have another commit in store which I was working on today which will solve this by splitting the migration into two commands:

1. `migrate-code-scanning-alerts`: Will fetch all SARIF-Reports from the source and reupload them on the target
2. `migrate-code-scanning-alert-states`: Will match all alerts from the source and set the states on the target

The first command is not (yet) safe to run multiple times, as the chronological order of SARIF Uploads is crucial to the alert states - and if the same SARIF is uploaded twice, that might leave the repository in a weird state.

The second one, however, can be ran multiple times. As such, I built it that it will stop and not do anything if there is a difference in the count between source- and target alerts, logging that the user may have to wait a few seconds. I also built in a `--force` flag in case the user does not want to wait and rather apply whats already there and migrate the rest of the alerts at a later stage.

I still have to test this on monday. If you agree with the approach, I'll put it into this PR then.

dylan-smith (approved):

Issue #756, opened 2022-12-09
Update integration tests to add a new one with a GHES 3.8+ server


dylan-smith:

tambling:
As part of https://github.com/github/octoshift/issues/5741, I've spun up a couple of 3.8 instances, one uploading archives to S3, one to Azure — their creds are in the Octoshift vault in 1Password, as `GHES Export Instance (AWS)` and `GHES Export Instance (Azure)`, respectively.

PR #771, opened 2023-01-10
Allow blob credentials to be optional after GHES 3.8

### LINKED ISSUE #755: Update validation logic to require cloud storage args for GHES 3.7 or earlier, and warn that they will be ignored if 3.8+
If the source is GHES 3.8+ we probably shouldn't fail if the user passes cloud storage args (e.g. `--azure-connection-string` `--aws-access-key`) instead we should spit out warnings saying that those args are no longer necessary with GHES 3.8+ and will be ignored.

We'll need to update the arg descriptions to make it clear that they are only needed with GHES 3.7 and earlier.

And be sure that the various combinations have good unit test coverage, as the various valid/invalid arg combinations on this command have gotten quite complex.

brianaj:
Fixes #755 

I didn't want to move around too much logic so just pulled out the GHES version check to happen earlier to we can properly validate args and then pass that along to the migrations API logic so we don't call the meta API twice.

- [ ] Issue linked
- [ ] Docs updated - created https://github.com/github/gh-gei/issues/770

Functionally tested as shown below. Real integration tests will be added in https://github.com/github/gh-gei/issues/756.

#### Tested BAU behavior with older GHES version (3.6.0)
image

#### Tested new behavior with new GHES version (3.8.0) - no credentials set in UI
image

Error message in cli (moved out of verbose log) since this is an expected error we should surface.

#### Tested new behavior with new GHES version (3.8.0) - credentials set in UI
image
image

#### Tested new behavior with new GHES version (3.8.0) - credentials set in UI and in cli 
image
Can see the note that the user does not need to set in cli anymore.

### DIFF
--- a/src/Octoshift/GithubClient.cs
+++ b/src/Octoshift/GithubClient.cs
@@ -170,14 +170,20 @@ namespace OctoshiftCLI
             {
                 (content, headers) = await SendAsync(httpMethod, url, body, expectedStatus, customHeaders);
             }
-            else if (response.StatusCode == HttpStatusCode.OK)
+            else if (expectedStatus == HttpStatusCode.OK)
             {
-                response.EnsureSuccessStatusCode();
+                try
+                {
+                    response.EnsureSuccessStatusCode();
+                }
+                catch (HttpRequestException ex)
+                {
+                    throw new HttpRequestException($"GitHub API error: {content}", ex, response.StatusCode);
+                }
             }
             else if (response.StatusCode != expectedStatus)
             {
-
-                throw new HttpRequestException(content, null, response.StatusCode);
+                throw new HttpRequestException($"Expected status code {expectedStatus} but got {response.StatusCode}", null, response.StatusCode);
             }
 
             return (content, headers);
--- a/src/Octoshift/GithubClient.cs
+++ b/src/Octoshift/GithubClient.cs
@@ -170,13 +170,14 @@ namespace OctoshiftCLI
             {
                 (content, headers) = await SendAsync(httpMethod, url, body, expectedStatus, customHeaders);
             }
-            else if (expectedStatus == HttpStatusCode.OK)
+            else if (response.StatusCode == HttpStatusCode.OK)
             {
                 response.EnsureSuccessStatusCode();
             }
             else if (response.StatusCode != expectedStatus)
             {
-                throw new HttpRequestException($"Expected status code {expectedStatus} but got {response.StatusCode}", null, response.StatusCode);
+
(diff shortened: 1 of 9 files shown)


src/Octoshift/GithubClient.cs
```diff
@@ -170,13 +170,14 @@ public virtual async IAsyncEnumerable<JToken> GetAllAsync(string url, Dictionary
...
             }
-            else if (expectedStatus == HttpStatusCode.OK)
+            else if (response.StatusCode == HttpStatusCode.OK)
             {
                 response.EnsureSuccessStatusCode();
             }
             else if (response.StatusCode != expectedStatus)
             {
-                throw new HttpRequestException($"Expected status code {expectedStatus} but got {response.StatusCode}", null, response.StatusCode);
+
```
brianaj, line 179: We never got to this code until now as EnsureSuccessStatusCode was throwing an HttpRequestException for non-200 responses. Made it where we are now more explicit in the message (since the status code is already part of the exception) which gives us the opportunity to catch specific exceptions like I am doing in the API.

brianaj (commented):
Not sure if we could add more test coverage, but that can be in future PR if there are suggestions.

src/OctoshiftCLI.Tests/GithubApiTests.cs
```diff
@@ -2255,6 +2255,31 @@ public async Task StartGitArchiveGeneration_Returns_The_Initiated_Migration_Id()
...
+            _githubClientMock
+                .Setup(m => m.PostAsync(url, It.Is<object>(x => x.ToJson() == payload.ToJson()), null))
+                    .ThrowsAsync(new HttpRequestException(exception_message, null, HttpStatusCode.BadGateway));
+
+            // Act
+            Func<Task> act = async () => await _githubApi.StartGitArchiveGeneration(GITHUB_ORG, GITHUB_REPO);
+
+            // Assert
+            await act.Should().ThrowExactlyAsync<OctoshiftCliException>()
+                .WithMessage(exception_message);
```
dylan-smith, line 2279: The other unit tests in here use a slightly different pattern to test for exceptions (the way you did it here actually is probably nicer, but we should probably be consistent unless we want to change it everywhere).

```suggestion
            await _githubApi.Invoking(api => api.StartGitArchiveGeneration(GITHUB_ORG, GITHUB_REPO))
                .Should()
                .ThrowExactlyAsync<OctoshiftCliException>()
                .WithMessage(exception_message);
```

dylan-smith (approved):
:shipit:

Issue #740, opened 2022-11-17
migrate-repo: output references Azure when AWS S3 bucket is used


DI-DaveGoodine:
## Description
When migrating a repo from GHES to a github.com org, using AWS S3, the output of gei references azure blob storage.

```
[10:24 PM] [INFO] Uploading archive 2022-11-16_22-24-14-9-git_archive.tar.gz to AWS S3
[10:24 PM] [INFO] Uploading archive 2022-11-16_22-24-14-10-metadata_archive.tar.gz to AWS S3
[10:24 PM] [INFO] Archives uploaded to Azure Blob Storage, now starting migration...
```

## Reproduction Steps
1. Setup migration: create s3 bucket, create user with api keys, create source and destination PAT.
2. Issue migrate repo command `gh gei migrate-repo --github-source-org TF-Modules --github-target-org enverus-cts --source-repo tf_module_eks --target-repo sre.tf-modules.eks --ghes-api-url https://git.drillinginfo.com/api/v3 --aws-bucket-name enverus-cts-github-enterprise-importer --wait`
3. observe output.

timrogers:
Thanks for reporting this! Great spot ❇️

PR #773, opened 2023-01-11
Update log message after successful blob storage upload to not refer to a specific blob storage provider

timrogers:
When running an indirect migration, after finishing the upload of
the archives to blob storage, the CLI outputs the following line
to the logs:

> Archives uploaded to Azure Blob Storage, now starting migration...

It's great to tell the user that the blob storage upload is done,
but we now support Amazon S3 as well as Azure Blob Storage, but
this log line *always* mentions Azure.

This updates the log output to just refer to "blob storage" rather
than a specific provider. Previous lines in the logs will make it
clear which provider was used, e.g.:

> Uploading archive 2022-11-16_22-24-14-10-metadata_archive.tar.gz to AWS S3

Fixes https://github.com/github/gh-gei/issues/740.

- [x] Issue linked

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -1,2 +1,3 @@
 - Improve error messages when the specified target organization or enterprise cannot be found
 - Mask the value for `AWS_ACCESS_KEY` and `AWS_SECRET_KEY` parameters in log output
+- Fix log output so we don't say we've finished upload to Azure Blob Storage when you're actually using Amazon S3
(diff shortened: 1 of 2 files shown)


dylan-smith (approved):

timrogers:
@dylan-smith Thanks for the review and for updating the PR. Any idea why this is blocked? It looks like all of the checks passed.

dylan-smith:
> @dylan-smith Thanks for the review and for updating the PR. Any idea why this is blocked? It looks like all of the checks passed.

it's because I kicked off a bunch of other actions triggering stuff at the same time, and sometimes int tests get cancelled (since only 1 can run at a time, and actions has no queuing concept for this).  when that happens we just need to rerun it, I'll do it now

PR #780, opened 2023-01-12
Increase Signed url auth time to avoid fail migrations due to queue times

juanmrad:
We have noticed that some migrations stay in the queue for too long which causes the Signed URL to be invalid. This pr looks at increasing TTL of the url a bit longer to prevent issues from our users due to queued times.

- [ ] Issue linked


dylan-smith:
My fear with doing this is that it may hide other problems. E.g. the recent migration that got stuck in the queue for 37 hours, if this timeout was 48hours we probably would have never noticed that we had a problem.

Interested in @timrogers opinion on this

brianaj (approved):

timrogers:
> My fear with doing this is that it may hide other problems. E.g. the recent migration that got stuck in the queue for 37 hours, if this timeout was 48hours we probably would have never noticed that we had a problem.
> 
> Interested in @timrogers opinion on this

All in all, I think this is the right change to make. 

The `generate-script` floe encourages people to queue up many migrations at once and then wait for us to burn through the queue. At the moment, that isn't a viable strategy beyond a certain size because your archive URLs will expire. I think we should extend the validity

That said, we should also improve our instrumentation to address the concern you raised and give us a better view of how long migrations are getting queued for.

I suspect that some customers will be concerned by the security implications of longer signed URL validity. We may at some point want to make this configurable, but I don't think it's necessary now.

timrogers (approved):

dylan-smith (changes requested):
can we add a bullet to the release notes please to let customers know

RELEASENOTES.md
```diff
@@ -1,3 +1,4 @@
 - Improve error messages when the specified target organization or enterprise cannot be found
 - Mask the value for `AWS_ACCESS_KEY` and `AWS_SECRET_KEY` parameters in log output
-- Fix log output so we don't say we've finished upload to Azure Blob Storage when you're actually using Amazon S3
\ No newline at end of file
+- Fix log output so we don't say we've finished upload to Azure Blob Storage when you're actually using Amazon S3
+- Extend the validity of blob storage signed URLs so migration can still be successful even if there is a long queue of migrations
```
dylan-smith, line 4:
```suggestion
- Extend the expiration of blob storage signed URLs from 24hrs to 48hrs so migration can still be successful even if there is a long queue of migrations
```

dylan-smith:
FYI looks like the failed integration test was due to this known bug: https://github.com/github/octoshift/issues/5688

dylan-smith (approved):

PR #783, opened 2023-01-24
Don't validate bbs user and pass if kerberos is set

### LINKED ISSUE #736: `bbs2gh migrate-repo` trying to validate user/password when `--kerberos` specified
## Description

image

When specifying `--kerberos` you shouldn't need to provide a username and password. The way it's working now if you provide a dummy value for username/password it will work and properly use kerberos ignoring whatever username/password you provided.

ArinGhazarian:
Fixes #736 

The PR fixes a bug where the arg validations fails if `--kerberos` is set without providing BBS username and password. 

- [x] Release notes updated (if appropriate) --> Not required until public beta
- [x] Issue linked


src/bbs2gh/Handlers/MigrateRepoCommandHandler.cs
```diff
@@ -344,14 +344,24 @@ private void ValidateOptions(MigrateRepoCommandArgs args)
 
         if (ShouldGenerateArchive(args))
         {
-            if (GetBbsUsername(args).IsNullOrWhiteSpace())
+            if (args.Kerberos)
             {
-                throw new OctoshiftCliException("BBS username must be either set as BBS_USERNAME environment variable or passed as --bbs-username.");
+                if (args.BbsUsername.HasValue() || args.BbsPassword.HasValue())
+                {
+                    throw new OctoshiftCliException("--bbs-username and --bbs-password cannot be provided with --kerberos.");
```
brianaj, line 351: Should we thrown and exception or just log that we are ignoring and you no longer need to provide username and password here?

brianaj (commented):
Little confused if we want to throw an exception here as the original issue said it should just be ignored.

src/bbs2gh/Handlers/MigrateRepoCommandHandler.cs
ArinGhazarian: I guess if they provide wrong input we need to let them know to fix it.

brianaj: This is similar to my changes when if they provide cloud storage info, we just ignore and log it is not needed for their GHES version. I think you can get product's input on how to handle extra info cc @timrogers.

ArinGhazarian: I still think we need to be explicit about garbage/invalid input and so far it's been our approach so even if we want to change it we need to change it everywhere to be consistent. That also means that we probably need to change the cloud storage stuff too if the PR has already been merged. 
@timrogers what do you think?

dylan-smith: I could go either way. I'd probably lean towards a warning instead of an error, as we can still successfully migrate and ignore extraneous args. Feels like a slightly better UX. But having more strict validations encourages users to do the right things.

dylan-smith (approved):

src/bbs2gh/Handlers/MigrateRepoCommandHandler.cs
timrogers: Sorry for being slow to jump on this thread. I'd prefer to go with the explicit approach where we error if the input params don't make sense together. I realise that's just a subjective preference, but it's the call I'd make 😊

PR #786, opened 2023-01-26
Rename archiveId to migrationId for GitHub Migrations API execution

pnsk:
- [ ] Issue linked

Hi 👋 

Currently if we run into an error when running GitHub Migrations API to generate an archive file, we see the following messages in the octoshift log:

```
[15:35] [INFO] Waiting for archive with id <id> generation to finish. Current status: failed
[15:35] [ERROR] OctoshiftCLI.OctoshiftCliException: Archive generation failed for id: <id>
```

I think renaming `id` to `Migration Id` in the log should be helpful for the investigation.

Additionally, in the [src/gei/Handlers/MigrateRepoCommandHandler.cs](https://github.com/github/gh-gei/blob/724c11854fd9fa012f5f4b018c11feb411563b5c/src/gei/Handlers/MigrateRepoCommandHandler.cs), the id is defined as "archiveId", but it's actually migration id for ["Organization migrations" API](https://docs.github.com/en/rest/migrations/orgs?apiVersion=2022-11-28). To clarify, this PR will rename the "archiveId"  to "migrationId".

Thanks!


dylan-smith:
Isn't the archive ID different than the migration ID? When we say migration ID in GEI we usually mean the RM_123456... ID. I thought that this ID is something different, as in a single migration we will have 2 different archive ID's (one for the git archive and one for the metadata archive), but only one overall migration ID (e.g. RM_123456...)

pnsk:
@dylan-smith Thank you for your comment 😄 

> Isn't the archive ID different than the migration ID? 

I'm not sure where the "archive ID" here is defined, but the ID is "migration id" for [the GitHub REST API –Migrations API](https://docs.github.com/en/rest/migrations/orgs?apiVersion=2022-11-28).

https://github.com/github/gh-gei/blob/724c11854fd9fa012f5f4b018c11feb411563b5c/src/gei/Handlers/MigrateRepoCommandHandler.cs#L168

https://github.com/github/gh-gei/blob/724c11854fd9fa012f5f4b018c11feb411563b5c/src/Octoshift/GithubApi.cs#L573-L593

However, as you mentioned, for GEI CLI, we already used the "Migration Id" for Id starting with `RM_` which is generated when running GitHub GraphQL API for migrations on target GitHub Enterprise Cloud.

So, I changed to "Source Migrations API Id". 

Could you please let me know how you feel with the name?

Thanks!

dylan-smith:
IMO I don't think this is an improvement (and it only changes the variable names for the gitArchive not the metadataArchive so we'd probably have to change that too for consistency).

I'm leaning towards leaving the current naming as-is. @timrogers or @ArinGhazarian do you want to weigh in?

pnsk:
@dylan-smith thank you for your comment!

Could you please let me know where archive id is from?

I'm not stick on the name of "Migrations API Id" so I'm happy to update it to better naming but I can't find a good reason why it's called 'archive id'. I feel like the current name of "archive id" points to the archive file itself. 

If renaming the name of "archive id" will be rejected, at least could you please update the following error message because the name of only 'id' is not helpful for troubleshooting:

> [15:35] [ERROR] OctoshiftCLI.OctoshiftCliException: Archive generation failed for id: <id>

Actually when I ran into this error message, I was not sure what the 'id' means. I would like to change it to more intuitive one.

Thanks!

dylan-smith:
> Could you please let me know where archive id is from?
> 
> I'm not stick on the name of "Migrations API Id" so I'm happy to update it to better naming but I can't find a good reason why it's called 'archive id'. I feel like the current name of "archive id" points to the archive file itself.


We have functions `StartGitArchiveGeneration` and `StartMetadataArchiveGeneration` in the CLI codebase, the ID here is the ID returned from those functions, which ultimately is the ID returned from the API that we call to start the generation of archives `{_apiUrl}/orgs/{org}/migrations`

ArinGhazarian:
👋  @pnsk ,
The `Id` is the archive id which is the same as a the migration id. In all of the legacy migration logs (AFAIK) in Splunk the archive id is actually the migration id but it is being logged as archive id, so IMO it's better to leave it as is and not change it to be consistent.

pnsk:
@dylan-smith
Thank you for your reply.
I wanted to see where the id GitHub Migrations API returns is defined officially 😅


> We have functions StartGitArchiveGeneration and StartMetadataArchiveGeneration in the CLI codebase, the ID here is the ID returned from those functions, which ultimately is the ID returned from the API that we call to start the generation of archives {_apiUrl}/orgs/{org}/migrations

[Yeah I see it and in the same file the id named 'migration_id' is used to run GitHub Migrations Id. If you all prefer 'archive id', I feel it would be better to change them to 'archive id' though...

pnsk:
@ArinGhazarian thank you for letting me know your opinion!!

> The Id is the archive id which is the same as a the migration id. In all of the legacy migration logs (AFAIK) in Splunk the archive id is actually the migration id but it is being logged as archive id,

As GEI CLI is open sourced, I feel it would be better to use the official name so that developers can do troubleshooting easier. 

If you need the name of 'archive id' for your internal debugging, how do you feel to output both 'Migrations API Id (archive id) in the log?

ArinGhazarian:
> As GEI CLI is open sourced, I feel it would be better to use the official name so that developers can do troubleshooting easier.

I don't feel like it's necessary since it's fairly obvious from the `StartGitArchiveGeneration` and `StartMetadataArchiveGeneration` methods what `id` means. Also the official name is `id` as returned by the [migrations API](https://docs.github.com/en/rest/migrations/orgs?apiVersion=2022-11-28#start-an-organization-migration).

> Archive generation failed for Source Migration ID (archived id)

To me it's too verbose, at the end of the day we are going to debug this by looking at Splunk/Sentry logs and we all know what archive id is. So both of the following would wrok:
- Archive generation failed for id: <id> --> this clearly says archive for id which implies archive id
- Archive generation failed for migration id: <id> --> even though this is more verbose but I can live with it 😃 

Now any thing other than these would be misleading, for example `sourceMigrationApiId` in my mind means something different that `sourceMigrationId`. API Id usually is an id that only the API returns (i.e. a GUID) that can be different that the actual id of the underlying resource.

src/gei/Handlers/MigrateRepoCommandHandler.cs
```diff
@@ -165,16 +165,16 @@ private string ExtractGhesBaseUrl(string ghesApiUrl)
       bool lockSourceRepo,
       bool blobCredentialsRequired)
     {
-        var gitDataArchiveId = await _sourceGithubApi.StartGitArchiveGeneration(githubSourceOrg, sourceRepo);
-        _log.LogInformation($"Archive generation of git data started with id: {gitDataArchiveId}");
+        var gitSourceMigrationsApiId = await _sourceGithubApi.StartGitArchiveGeneration(githubSourceOrg, sourceRepo);
```
ArinGhazarian, line 168: as I mentioned [here](https://github.com/github/gh-gei/pull/786#issuecomment-1407258606) using the term `ApiId` is misleading. I really think the variable name was good as is.

src/gei/Handlers/MigrateRepoCommandHandler.cs
```diff
@@ -165,16 +165,16 @@ private string ExtractGhesBaseUrl(string ghesApiUrl)
...
+        _log.LogInformation($"Archive generation of git data started with id: {gitSourceMigrationsApiId}");
```
ArinGhazarian, line 169: same

src/gei/Handlers/MigrateRepoCommandHandler.cs
```diff
@@ -216,20 +216,20 @@ private async Task<(string, string)> UploadArchivesToAws(string bucketName, stri
         return (authenticatedGitArchiveUri.ToString(), authenticatedMetadataArchiveUri.ToString());
     }
 
-    private async Task<string> WaitForArchiveGeneration(GithubApi githubApi, string githubSourceOrg, int archiveId)
+    private async Task<string> WaitForArchiveGeneration(GithubApi githubApi, string githubSourceOrg, int sourceMigrationsApiId)
```
ArinGhazarian, line 219: same

src/gei/Handlers/MigrateRepoCommandHandler.cs
```diff
@@ -216,20 +216,20 @@ private async Task<(string, string)> UploadArchivesToAws(string bucketName, stri
...
     {
         var timeout = DateTime.Now.AddHours(ARCHIVE_GENERATION_TIMEOUT_IN_HOURS);
         while (DateTime.Now < timeout)
         {
-            var archiveStatus = await githubApi.GetArchiveMigrationStatus(githubSourceOrg, archiveId);
-            _log.LogInformation($"Waiting for archive with id {archiveId} generation to finish. Current status: {archiveStatus}");
+            var archiveStatus = await githubApi.GetArchiveMigrationStatus(githubSourceOrg, sourceMigrationsApiId);
+            _log.LogInformation($"Waiting for archive with Source Migrations API ID {sourceMigrationsApiId} generation to finish. Current status: {archiveStatus}");
```
ArinGhazarian, line 225: same here too.

ArinGhazarian (changes requested):

timrogers:
@pnsk Thanks for taking the time to create this PR! ❤️ 

I'm really grateful for your efforts to improve our logs and error messages - these things can have a big impact on the overall GEI user experience.

But in this specific case, I don't think that the change you're proposing would be a meaningful improvement to the messaging that we have now. I'm not sure how the changes would make it easier for someone to debug a failed migration. What do you think?

pnsk:
Thank you so much @dylan-smith @ArinGhazarian @timrogers for the conversation.

It's likely to better to keep the name as 'archive id' as it was. Thanks you all for explaining the background why you use the name of `archive id` for the `id` of GitHub REST Migrations API.

> I also agree with @dylan-smith here, using the term migration id in Octoshift points to a migration on the Octoshift side which is totally different than a legacy migration id on the GitHub side. On both sides they are called migration id though, so it's better to distinguish between them.

It does make sense 👍 Thank you!

From our conversation, I reverted the name to `archive id` in https://github.com/github/gh-gei/blob/main/src/gei/Handlers/MigrateRepoCommandHandler.cs 🙇 

 Instead, I updated the name of `migrationId` which used for `id` as returned by GitHub REST Migrations APIs in [the src/Octoshift/GithubApi.cs](https://github.com/github/gh-gei/blob/main/src/Octoshift/GithubApi.cs)  to `archiveId`  for consistency.  How do you feel with it?

src/Octoshift/GithubApi.cs
```diff
@@ -375,14 +375,14 @@ public virtual async Task<string> StartOrganizationMigration(string sourceOrgUrl
             return (string)data["data"]["startOrganizationMigration"]["orgMigration"]["id"];
         }
 
-        public virtual async Task<(string State, string SourceOrgUrl, string TargetOrgName, string FailureReason, int? RemainingRepositoriesCount, int? TotalRepositoriesCount)> GetOrganizationMigration(string migrationId)
```
ArinGhazarian, line 378: This is not the archive Id, here it is the `migrationId` on the Octoshift side.

src/Octoshift/GithubApi.cs
```diff
@@ -419,14 +419,14 @@ public virtual async Task<string> StartBbsMigration(string migrationSourceId, st
             );
         }
 
-        public virtual async Task<(string State, string RepositoryName, string FailureReason)> GetMigration(string migrationId)
```
ArinGhazarian, line 422: same here, this is an Octoshift migration id.

ArinGhazarian (changes requested):

pnsk:
@ArinGhazarian thanks! I've updated based on your comments!

RELEASENOTES.md
```diff
@@ -1,2 +1,2 @@
+- Renamed the parameter name from migrationId to archiveId for GitHub REST Migrations APIs
```
timrogers, line 1: If we merge this change, I think we can leave out the release notes as it's a refactor that doesn't affect the customer.

pnsk: @timrogers thanks! I've remove the release notes!

pnsk:
@ArinGhazarian could you please review this PR again? Thanks!

timrogers (commented):
I don't mind either way on this change - I think the variable names before and after are both okay. I'll let @ArinGhazarian be the decider since I'm not really an engineer ❤️

timrogers (approved):

timrogers:
@dylan-smith How can I get the integration test to run? It seems to be skipping 🤔

dylan-smith:
this doesn't currently compile. If you give me access to push to the fork/branch I can fix it and merge this. There should be a checkbox when @pnsk is lookin at the PR to grant permissions as per this doc: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

pnsk:
@dylan-smith thanks! I've sent an invitation for you to join my repo!

dylan-smith (approved):

PR #787, opened 2023-01-26
626 check for target repo existance

### LINKED ISSUE #626: Don't generate GHES archives if target repo already exists
One of our goals is to have our generated scripts be idempotent. Meaning if you run a script and it dies halfway through, once you fix whatever caused it to fail you can re-run the script unchanged, and it will basically do noops for the first half that have already been migrated then pickup where it failed out the first time.

In order to achieve this we have tried to make all our commands idempotent. For `migrate-repo` the way we did that is check if the target repo exists, and if so spit out a warning and return.

Recently - in the interest of minimizing the number of REST API calls the CLI makes - we changed the implementation to eliminate the REST call to check the target repo existence, and instead just kicked off the migration, but if it fails we inspect the error returned and specifically look to see if it failed because the target repo already existed, and if so we do the same behavior (spit out a warning and exit without error).

HOWEVER, if we are migrating from GHES, in the case where the target repo already exists we wont' discover that until after we've already asked GHES to generate the migration archives, downloaded them, uploaded them to azure, then tried to migrate them. That's alot of unnecessary work, and wasted time.

In GHES migrations specifically we should re-introduce the REST API call to check for target repo existence, before trying to generate the migration archive.

hfishback01:
Closes #626 

Makes an API call to check if the target repo already exists; if it does, the GHES archives are not generated.

- [x] Issue linked


src/gei/Handlers/MigrateRepoCommandHandler.cs
```diff
@@ -242,7 +249,8 @@ private async Task<bool> DetermineIfBlobCredentialsRequired(MigrateRepoCommandAr
         if (args.GhesApiUrl.HasValue())
         {
             _log.LogInformation("Using GitHub Enterprise Server - verifying server version");
-            var ghesVersion = await _sourceGithubApi.GetEnterpriseServerVersion();
+            // var ghesVersion = await _sourceGithubApi.GetEnterpriseServerVersion();
+            var ghesVersion = "3.0.4";
```
dylan-smith, line 253:
```suggestion
            var ghesVersion = await _sourceGithubApi.GetEnterpriseServerVersion();
```

src/Octoshift/GithubApi.cs
```diff
@@ -105,6 +106,23 @@ public virtual async Task RemoveTeamMember(string org, string teamSlug, string m
...
+                await _client.GetNonSuccessAsync(url, HttpStatusCode.NotFound);
+            }
+            catch (HttpRequestException ex)
+            {
+                if (ex.StatusCode == HttpStatusCode.OK)
+                {
+                    return true;
+                }
+            }
+            return false;
```
ArinGhazarian, line 123: 1. It's a bit strange to always rely on an exception to determine the result of a method, instead I would naturally return either `false` or `true` and then check only the exceptional case in the `catch` block:
```c#
try
{
    await _client.GetAsync(url);
    return true;
}
catch (HttpRequestException ex) when (ex.StatusCode == HttpStatusCode.NotFound)
{
    return false;
}
```
2. The catch block actually swallows all `HttpRequestExceptions`! so for example if there is a rate limit exception it swallows it and returns `false` instead. The way I re-wrote it above won't have this problem because the catch block is only going to get executed if the condition is true.

ArinGhazarian (changes requested):

src/Octoshift/GithubApi.cs
```diff
@@ -4,6 +4,7 @@
 using System.Net;
 using System.Net.Http;
 using System.Threading.Tasks;
+using Amazon.S3.Model;
```
ArinGhazarian, line 7: We don't need this using here, it's probably a mistake.

ArinGhazarian (changes requested):

src/Octoshift/GithubApi.cs
dylan-smith: we originally had the code very similar to this, the problem is in the normal case (repo does not exist) the GetAsync will retry 5 times, which will affect perf.

ArinGhazarian: then I will reverse the logic like this. We still shouldn't swallow all exceptions:

```c#
try
{
    await _client. GetNonSuccessAsync(url, HttpStatusCode.NotFound);
    return false;
}
catch (HttpRequestException ex) when (ex.StatusCode == HttpStatusCode.OK)
{
    return true;
}
```

RELEASENOTES.md
```diff
@@ -1 +1,8 @@
+- Improve error messages when the specified target organization or enterprise cannot be found
+- Mask the value for `AWS_ACCESS_KEY` and `AWS_SECRET_KEY` parameters in log output
+- Fix log output so we don't say we've finished upload to Azure Blob Storage when you're actually using Amazon S3
+- Extend the expiration of blob storage signed URLs from 24hrs to 48hrs so migration can still be successful even if there is a long queue of migrations
+- Skip the upload to Azure/AWS blob storage when migrating from GHES 3.8+, as GHES will now handle putting the archives into blob storage
+- Fixed a bug where bad credentials were incorrectly being treated as rate-limit errors
```
ArinGhazarian, line 6: Should this all be here? I guess this happened because of a merge conflict.

src/Octoshift/GithubApi.cs
```diff
@@ -105,6 +105,20 @@ public virtual async Task RemoveTeamMember(string org, string teamSlug, string m
             await _retryPolicy.HttpRetry(() => _client.DeleteAsync(url), _ => true);
         }
 
+        public virtual async Task<bool> DoesRepoExist(string org, string repo)
```
ArinGhazarian, line 108: I would also like to see a unit tests for this.

ArinGhazarian (commented):
It looks good now, I just added a comment about the release notes and adding a couple of unit tests for the added `GithubAPI` method.

src/OctoshiftCLI.Tests/gei/Handlers/MigrateRepoCommanHandlerTests.cs
```diff
@@ -47,6 +47,30 @@ public MigrateRepoCommandHandlerTests()
                 _mockHttpDownloadService.Object);
         }
 
+        // TO DO: Reformat to actually work for our scenario
```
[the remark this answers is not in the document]
hfishback01: Addressed in https://github.com/github/gh-gei/pull/787/commits/2afeca3056b0bd970e90cee0a09daedb5ec0c9c5! 🎉

src/gei/Handlers/MigrateRepoCommandHandler.cs
hfishback01: Addressed in https://github.com/github/gh-gei/pull/787/commits/2afeca3056b0bd970e90cee0a09daedb5ec0c9c5! 🎉

src/Octoshift/GithubApi.cs
hfishback01: Addressed in https://github.com/github/gh-gei/pull/787/commits/2afeca3056b0bd970e90cee0a09daedb5ec0c9c5! 🎉

RELEASENOTES.md
hfishback01: Addressed in https://github.com/github/gh-gei/pull/787/commits/ed65a969f41f9ab96f6811fc9b8ff9410000f87a :tada:

src/Octoshift/GithubApi.cs
hfishback01: Addressed in https://github.com/github/gh-gei/pull/787/commits/f0def9ecd76d60ade5b2c4eddfae9913e85493b2! :tada:

src/Octoshift/GithubApi.cs
dylan-smith: Addressed in [`eb2601b`](https://github.com/github/gh-gei/pull/787/commits/eb2601b00ae9bb68e82fe03f158415332a9200e0)

dylan-smith (approved):

ArinGhazarian (approved):
🎉

PR #788, opened 2023-01-27
Add support of downloading BBS archive using SMB

### LINKED ISSUE #526: Transfer the archive off BBS Server to CLI machine (using SMB)
Leverage the output of [this spike](https://github.com/github/octoshift/issues/5181) to implement the SMB archive transfer.

ArinGhazarian:
Closes #526 

- [x] Release notes updated (if appropriate) --> no release notes for BBS
- [x] Issue linked

This PR adds support of downloading the BBS export archive off of Windows machines using the SMB protocol. In general there are two parts in this PR:

1. The new `BbsSmbArchiveDownlaoder` service class to encapsulate all we need for SMB download.
2. The wire-up code that leverages the service class and the SMB credentials.


src/OctoshiftCLI.Tests/bbs2gh/Handlers/MigrateRepoCommandHandlerTests.cs
```diff
@@ -534,6 +534,28 @@ public async Task Invoke_With_Bbs_Server_Url_Throws_When_Smb_User_Is_Provided_An
             await _handler.Invoking(x => x.Handle(args)).Should().ThrowExactlyAsync<OctoshiftCliException>();
         }
 
+        [Fact]
+        public async Task Invoke_With_Bbs_Server_Url_Should_Not_Throw_When_Smb_User_Is_Provided_And_Smb_Password_Is_Provided_Via_Environment_Variable()
```
timrogers, line 538: What would happen if you specified SMB configuration and SSH configuration together? That might be a good thing to document through a test.

src/bbs2gh/Commands/GenerateScriptCommand.cs
```diff
@@ -50,7 +52,8 @@ public class GenerateScriptCommand : CommandBase<GenerateScriptCommandArgs, Gene
 
     public Option<string> BbsSharedHome { get; } = new(
         name: "--bbs-shared-home",
-        description: "Bitbucket server's shared home directory. If not provided \"/var/atlassian/application-data/bitbucket/shared\" will be used.");
+        description: "Bitbucket server's shared home directory. If not provided \"/var/atlassian/application-data/bitbucket/shared\" will be used when using SSH to download the export archive " +
```
timrogers, line 55: What do you think of rephrasing this more like this? I think it might be easier to understand.

> Bitbucket Server's shared home directory, where migration archives are stored. Defaults to `/var/atlassian/application-data/bitbucket/shared\` if downloading the archive from the server using SSH, and `c$\\atlassian\\applicationdata\\bitbucket\\shared\` if downloading using SMB.

src/bbs2gh/Commands/GenerateScriptCommand.cs
```diff
@@ -65,6 +68,16 @@ public class GenerateScriptCommand : CommandBase<GenerateScriptCommandArgs, Gene
         description: "The SSH port (default: 22).",
         getDefaultValue: () => 22);
 
+    public Option<string> SmbUser { get; } = new(
+        name: "--smb-user",
+        description: "The SMB user to be used for downloading the export archive off of the Bitbucket server." +
+                     $"{Environment.NewLine}" +
+                     "Note: The SMB password will also be required in order to connect to a SMB share which has to be set as an environment variable before running the script.");
+
+    public Option<string> Domain { get; } = new(
```
timrogers, line 77: What do you think of prefixing this with `Smb` to match the offer SMB-specific options?

```suggestion
    public Option<string> SmbDomain { get; } = new(
```

src/bbs2gh/Commands/GenerateScriptCommand.cs
```diff
@@ -65,6 +68,16 @@ public class GenerateScriptCommand : CommandBase<GenerateScriptCommandArgs, Gene
         description: "The SSH port (default: 22).",
         getDefaultValue: () => 22);
 
+    public Option<string> SmbUser { get; } = new(
+        name: "--smb-user",
+        description: "The SMB user to be used for downloading the export archive off of the Bitbucket server." +
```
timrogers, line 73: I think this makes it a bit clearer what this does:

```suggestion
        description: "The SMB user used for authentication when downloading the export archive from the Bitbucket Server instance." +
```

src/bbs2gh/Commands/GenerateScriptCommand.cs
```diff
@@ -65,6 +68,16 @@ public class GenerateScriptCommand : CommandBase<GenerateScriptCommandArgs, Gene
...
+                     $"{Environment.NewLine}" +
+                     "Note: The SMB password will also be required in order to connect to a SMB share which has to be set as an environment variable before running the script.");
```
timrogers, line 75: I think this is a bit simpler and makes it clear exactly what you need to do.

```suggestion
                     "You must also specify the SMB password using the `SMB_PASSWORD` environment variable.");
```

src/bbs2gh/Commands/GenerateScriptCommand.cs
```diff
@@ -65,6 +68,16 @@ public class GenerateScriptCommand : CommandBase<GenerateScriptCommandArgs, Gene
...
+
+    public Option<string> Domain { get; } = new(
+        name: "--domain",
+        description: "The optional domain name when using SMB for downloading the export archive.");
```
timrogers, line 79: How does the user know if they need to specify a domain name? Will it be obvious to them? I don't know too much about Windows server administration.

src/bbs2gh/Commands/GenerateScriptCommand.cs
timrogers: For my suggestions in this file, if you accept them, we should apply them to `src/bbs2gh/Commands/MigrateRepoCommand.cs` too.

src/bbs2gh/Handlers/MigrateRepoCommandHandler.cs
```diff
@@ -405,9 +410,9 @@ private void ValidateDownloadOptions(MigrateRepoCommandArgs args)
             throw new OctoshiftCliException("Both --ssh-user and --ssh-private-key must be specified for SSH download.");
         }
 
-        if (args.SmbUser.HasValue() ^ args.SmbPassword.HasValue())
+        if ((args.SmbUser.HasValue() && GetSmbPassword(args).IsNullOrWhiteSpace()) || (args.SmbPassword.HasValue() && args.SmbUser.IsNullOrWhiteSpace()))
         {
-            throw new OctoshiftCliException("Both --smb-user and --smb-password must be specified for SMB download.");
+            throw new OctoshiftCliException("Both --smb-user and --smb-password (or SMB_PASSWORD env. variable) must be specified for SMB download.");
```
timrogers, line 415:
```suggestion
            throw new OctoshiftCliException("Both --smb-user and --smb-password (or `SMB_PASSWORD` environment variable) must be specified for SMB download.");
```

src/bbs2gh/Services/BbsSmbArchiveDownloader.cs
```diff
@@ -0,0 +1,227 @@
...
+    {
+        if (DateTime.Now < _nextProgressReport)
+        {
+            return;
+        }
+
+        var totalProgressMessage = totalBytes.HasValue
+            ? $" out of {GetLogFriendlySize(totalBytes.Value)} ({GetPercentage(downloadedBytes, totalBytes.Value)})"
+            : "";
+        _log.LogInformation($"Download archive in progress, {GetLogFriendlySize(downloadedBytes)}{totalProgressMessage} completed...");
```
timrogers, line 125:
```suggestion
        _log.LogInformation($"Archive download in progress, {GetLogFriendlySize(downloadedBytes)}{totalProgressMessage} completed...");
```

src/bbs2gh/Services/BbsSmbArchiveDownloader.cs
```diff
@@ -0,0 +1,227 @@
...
+
+    private void LogProgress(long downloadedBytes, long? totalBytes)
+    {
+        if (DateTime.Now < _nextProgressReport)
+        {
+            return;
+        }
+
+        var totalProgressMessage = totalBytes.HasValue
+            ? $" out of {GetLogFriendlySize(totalBytes.Value)} ({GetPercentage(downloadedBytes, totalBytes.Value)})"
```
timrogers, line 123:
```suggestion
            ? $"/{GetLogFriendlySize(totalBytes.Value)} ({GetPercentage(downloadedBytes, totalBytes.Value)})"
```

src/bbs2gh/Services/BbsSmbArchiveDownloader.cs
```diff
@@ -0,0 +1,227 @@
...
+
+    private void ConnectToHost()
+    {
+        var isConnected = IPAddress.TryParse(_host, out var ipAddress)
+            ? _smbClient.Connect(ipAddress, SMBTransportType.DirectTCPTransport)
+            : _smbClient.Connect(_host, SMBTransportType.DirectTCPTransport);
+
+        if (!isConnected)
+        {
+            throw new OctoshiftCliException($"Unable to connect to host \"{_host}\".");
```
timrogers, line 175: For an error like this, do you think we should make it clear what stage in the process we've reached (i.e. we're trying to download using SMB)? Or is that going to be obvious from the existing log output?

src/bbs2gh/Services/BbsSmbArchiveDownloader.cs
```diff
@@ -0,0 +1,227 @@
...
+        }
+    }
+
+    private void Login()
+    {
+        var status = _smbClient.Login(_domainName ?? "", _smbUser, _smbPassword);
+
+        if (!IsSuccessStatus(status))
+        {
+            throw new OctoshiftCliException($"Unable to login with provided credentials (Status Code: {status}).");
```
timrogers, line 185: Same question here - should we make it clear that it's the SMB login that has failed rather than, for example, the login to the BBS API?

timrogers (commented):
Great to see this coming together ✨ I haven't reviewed the code, but I have taken a look at the argument naming, docs and error messages. If you have any questions or disagree with any of my thoughts, do comment back ❤️

src/OctoshiftCLI.Tests/bbs2gh/Handlers/MigrateRepoCommandHandlerTests.cs
ArinGhazarian: All of the possible (and reachable) configs are already covered in the following tests:

- Invoke_With_Bbs_Server_Url_Throws_When_Both_Ssh_User_And_Smb_User_Are_Provided
- Invoke_With_Bbs_Server_Url_Throws_When_Both_Ssh_User_And_Smb_Password_Are_Provided
- Invoke_With_Bbs_Server_Url_Throws_When_Both_Ssh_Private_Key_And_Smb_User_Are_Provided
- Invoke_With_Bbs_Server_Url_Throws_When_Both_Ssh_Private_Key_And_Smb_Password_Are_Provided

src/bbs2gh/Commands/GenerateScriptCommand.cs
ArinGhazarian: I never actually tested the SMB download with a domain, I guess if there is a domain then they probably need to specify it.

src/bbs2gh/Handlers/MigrateRepoCommandHandler.cs
ArinGhazarian: will apply the change but two things for consistency:
1. We didn't use back ticks in any of our command docs as far as I know since the help text is a plain text and back-ticks won't do anything special. That being said if you think it'd be better to use them, then we need to apply it to all other help docs too which can be done in a separate PR with a new ticket. 
2. If environment is preferred over env. then we need to change every other occurrence too. Probably we're going to need an issue for it 😃

src/bbs2gh/Services/BbsSmbArchiveDownloader.cs
ArinGhazarian: again for consistency, this should also be applied to SSH downloader too.

src/bbs2gh/Services/BbsSmbArchiveDownloader.cs
ArinGhazarian: this also needs to be applied to SSH downloader too.

src/bbs2gh/Services/BbsSmbArchiveDownloader.cs
ArinGhazarian: I guess it's going to be obvious since it only happens in the SMB flow.

src/bbs2gh/Services/BbsSmbArchiveDownloader.cs
ArinGhazarian: Same thing here too, it would be perfectly clear that this happened during the SMB download.

src/bbs2gh/Services/IBbsArchiveDownloader.cs
```diff
@@ -1,11 +1,11 @@
 using System.IO;
 using System.Threading.Tasks;
+using OctoshiftCLI.Extensions;
 
 namespace OctoshiftCLI.BbsToGithub.Services;
 
 public interface IBbsArchiveDownloader
 {
-    const string DEFAULT_BBS_SHARED_HOME_DIRECTORY = "/var/atlassian/application-data/bitbucket/shared";
     const string EXPORT_ARCHIVE_SOURCE_DIRECTORY = "data/migration/export";
```
synthead, line 9: What are your thoughts on supporting a Samba location that isn't hard-coded to inspect a subdirectory of the "shared Atlassian" directory?  For example, the `export/` directory could be shared directly (instead of the root shared directory).

I'm certainly good on shipping this as-is and possibly adding support for this in a future issue :slightly_smiling_face:

synthead (approved):
Wow, this looks great!  I left a nit and a comment that could incur scope creep :smile: Feel free to ship as-is if you'd like :+1:

src/bbs2gh/Services/IBbsArchiveDownloader.cs
ArinGhazarian: It's totally possible with my changes, all they need to do is to then set the optional `--bbs-home` directory. SMB is supposed to be used only in windows and by default the root directory is always shared for admins.

src/bbs2gh/Commands/GenerateScriptCommand.cs
ArinGhazarian: addressed in 236d419ec115ff9bb55d83c20a0c22d7ebb34d94

src/bbs2gh/Commands/GenerateScriptCommand.cs
ArinGhazarian: I am going to modify the copy text but gonna skip `where migration archives are stored` b/c there are a lot more things stored in the shared home, basically the entire Bitbucket data is stored there including the migration archive. This should be a known term to a Bitbucket admin so I don't think we really need to add more explanation :)

ArinGhazarian: addressed in 40c97a3e9cbeed8a169e3cd1252911c167dd261b

src/bbs2gh/Commands/GenerateScriptCommand.cs
ArinGhazarian: Addressed in d24bdf4f6625c50603738913d1e984396644becc

src/bbs2gh/Handlers/MigrateRepoCommandHandler.cs
ArinGhazarian: I am going to skip this b/c everywhere else we used `env.` instead of `environment` and no back ticks.

src/bbs2gh/Services/BbsSmbArchiveDownloader.cs
ArinGhazarian: addressed in 28ca1022e7e2599f3e374aa83643ce8dd535f1e2

src/bbs2gh/Services/BbsSmbArchiveDownloader.cs
ArinGhazarian: addressed in 28ca1022e7e2599f3e374aa83643ce8dd535f1e2

src/OctoshiftCLI.Tests/bbs2gh/Handlers/GenerateScriptCommandHandlerTests.cs
```diff
@@ -152,6 +154,7 @@ public async Task Two_Projects_Two_Repos_Each_All_Options()
         _mockEnvironmentVariableProvider.Verify(m => m.BbsPassword(It.IsAny<bool>()), Times.Never);
     }
 
+
```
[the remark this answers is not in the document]
ArinGhazarian: addressed in fa2b371b7cd19865c7246e0a37b4033524cbdb13

src/bbs2gh/Commands/GenerateScriptCommand.cs
ArinGhazarian: addressed in 00ba81e9e3a38610372fbdb52f77fea5d1b21662

PR #789, opened 2023-01-28
Creates check to determine if target org exists

### LINKED ISSUE #782: Don't generate GHES archives if target organization does not exist
When performing a migration from GHES, the CLI does not check that the provided `--github-target-org` exists before talking to the GHES instance and generating an archive.

This creates a lot of unnecessary work and wasted time in the case where the target organization doesn't exist so the migration is impossible.

We should validate the existence of the target organization upfront before starting the GHES archive generation.

See similar issue #626, which instead looks at the case where the target organization does exist and the target repo exists - in which case we can't migrate to that destination.

hfishback01:
Closes #782 

Creates logic to ensure the target org exists before generating GHES archives.

- [x] Issue linked

### DIFF
--- a/src/gei/Handlers/MigrateRepoCommandHandler.cs
+++ b/src/gei/Handlers/MigrateRepoCommandHandler.cs
@@ -65,7 +65,7 @@ public class MigrateRepoCommandHandler : ICommandHandler<MigrateRepoCommandArgs>
 
             if (!targetOrgExists)
             {
-                throw new OctoshiftCliException($"The target or, {args.GithubTargetOrg}, does not exist. Migration unable to continue at this time.");
+                throw new OctoshiftCliException($"The target org, {args.GithubTargetOrg}, does not exist. Migration unable to continue at this time.");
             }
 
             (args.GitArchiveUrl, args.MetadataArchiveUrl) = await GenerateAndUploadArchive(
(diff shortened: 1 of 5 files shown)


src/gei/Handlers/MigrateRepoCommandHandler.cs
```diff
@@ -55,13 +55,19 @@ public async Task Handle(MigrateRepoCommandArgs args)
...
 
-            if (targetExists)
+            if (targetRepoExists)
             {
                 throw new OctoshiftCliException($"A repository called {args.GithubTargetOrg}/{args.TargetRepo} already exists");
             }
 
+            if (!targetOrgExists)
+            {
+                throw new OctoshiftCliException($"The target org, {args.GithubTargetOrg}, does not exist. Migration unable to continue at this time.");
```
ArinGhazarian, line 68: IMO the `migration is unable to continue...` is a bit redundant since we're already throwing an exception and telling the user that we cannot continue 😃  

```suggestion
                throw new OctoshiftCliException($"The target org \"{args.GithubTargetOrg}\" does not exist.");
```

ArinGhazarian (commented):
Just added a small comment but the rest is good 🎉

src/gei/Handlers/MigrateRepoCommandHandler.cs
hfishback01: Fair point! Fixed here: c091ce9ac66d3dbfbdfa36db43f670651c5fc37a

ArinGhazarian (approved):
💯

PR #798, opened 2023-02-02
Unit Test: Generates archive if target repo doesn't exist

hfishback01:
Adds a unit test to specifically test if archive generation is attempted if the target repo doesn't exist.

Related to #782 and PR #[789](https://github.com/github/gh-gei/pull/789)

- [x] Issue linked

### DIFF
--- a/src/OctoshiftCLI.Tests/gei/Handlers/MigrateRepoCommanHandlerTests.cs
+++ b/src/OctoshiftCLI.Tests/gei/Handlers/MigrateRepoCommanHandlerTests.cs
@@ -61,6 +61,7 @@ namespace OctoshiftCLI.Tests.GithubEnterpriseImporter.Commands
                 GithubTargetOrg = TARGET_ORG,
                 TargetRepo = TARGET_REPO,
                 GhesApiUrl = GHES_API_URL,
+                AzureStorageConnectionString = AZURE_CONNECTION_STRING,
                 Wait = false,
             };
             await FluentActions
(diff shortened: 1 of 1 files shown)


hfishback01:
There are already a few tests that ensure successful migration for GHES... I tried to make this as bare bones as possible to highlight the fact that the target repo is absent.

src/OctoshiftCLI.Tests/gei/Handlers/MigrateRepoCommanHandlerTests.cs
```diff
@@ -70,6 +71,51 @@ await FluentActions
...
+                TargetRepo = TARGET_REPO,
+                GhesApiUrl = GHES_API_URL,
+                AzureStorageConnectionString = AZURE_CONNECTION_STRING,
+                Wait = false,
+            };
+            await handler.Handle(args);
+
+            // Assert
+            _mockSourceGithubApi.Verify(x => x.StartGitArchiveGeneration(It.IsAny<string>(), It.IsAny<string>()), Times.Once);
+        }
```
ArinGhazarian, line 117: It looks like this test is redundant since it's testing a natural happy path which should be already covered. The only thing that needs to be tested in the exceptional case (here if the repo exists) which I think has already been covered.

ArinGhazarian (changes requested):
I think we can simply close the PR unless there is something that I am missing.

src/OctoshiftCLI.Tests/gei/Handlers/MigrateRepoCommanHandlerTests.cs
hfishback01: I kinda thought the same but it had been asked for so I wanted to make sure that at least a PR was up for it. Will close :)

PR #800, opened 2023-02-03
Run Bitbucket integration tests against BBS 5.14.0 instance

ArinGhazarian:
Closes #https://github.com/github/gh-gei/issues/570

- [x] Issue linked

### Description
This PR added support of running the BBS's current INT test against a BBS 5.14.0 instance too. The following changes were made as part of this PR:
1. A new Bitbucket 5.14.0 server instance was created in Azure.
2. The BBS url was added as `InlineData` to support running the test against the new BBS instance.
3. Because now we have two BBS instances (8.5.0 and 5.14.0), just having one `SSH_KEY` secret wouldn't suffice, so two SSH keys were added to the repo secrets with `SSH_KEY_BBS_{VERSION}` format (eg. `SSH_KEY_BBS_5_14_0`). After merging this PR, the old `SSH_KEY` secret can be safely removed.
4. All git-ops used to be done with `LibGit2Sharp` were replaced by shell-out commands and the library was removed from the INT tests project. The reason is the newly added test was consistently failing with `unknown 500` error raised by `LibGit2Sharp`, after chasing the error down the rabbit hole, it turned out that it was a bug in `LibGit2Sharp` library!


synthead (approved):
Looks great!

PR #801, opened 2023-02-03
611 archive generation retry logic

### LINKED ISSUE #611: Retry GHES archive generation
We've been seeing a bunch of integration test failures when generating GHES archives. Usually the error is something like the connection was closed, or other networking sounding errors.

We could probably do a retry on archive generation if it fails for any reason.

hfishback01:
Closes #611 

Add retry logic to archive generation in case of transient failure. Two tests, one to confirm retry is attempted on failure while the second confirms retry is not attempted when export is successful.

- [x] Issue linked


hfishback01:
Thinking about adding more detailed logging information. Might add additional test to ensure logging is done correctly but will need to check if that's redundant.

src/OctoshiftCLI.Tests/gei/Handlers/MigrateRepoCommandHandlerTests.cs
```diff
@@ -47,6 +47,100 @@ public MigrateRepoCommandHandlerTests()
...
+            };
+            await FluentActions
+                .Invoking(async () => await handler.Handle(args)).Should().ThrowExactlyAsync<OctoshiftCliException>();
+
+            // Assert
+            _mockSourceGithubApi.Verify(x => x.GetArchiveMigrationStatus(SOURCE_ORG, It.IsAny<int>()), Times.Exactly(7)); // 6 times for git archive, once for metadata
+        }
+
+        [Fact]
+        public async Task Doesnt_Retry_On_Exported_Status()
```
ArinGhazarian, line 101: generally speaking these kind of tests are going to be covered in a happy path, now if you want to test that a method hasn't been called, it can be done with a simply verify in the happy path test.

ArinGhazarian (changes requested):
I believe this PR doesn't cover what #611 asks:

> We have already added retries on HTTP failures. But sometime the archive generation might fail, but GHES returns an HTTP 200 with a message telling us the archive generation failed. In that case we should restart the archive generation again.

this is an edge case an in order to do it we need to know what message is being returned and add proper unit tests for that too.

src/gei/Handlers/MigrateRepoCommandHandler.cs
```diff
@@ -236,13 +236,19 @@ private async Task<string> WaitForArchiveGeneration(GithubApi githubApi, string
         {
             var archiveStatus = await githubApi.GetArchiveMigrationStatus(githubSourceOrg, archiveId);
             _log.LogInformation($"Waiting for archive with id {archiveId} generation to finish. Current status: {archiveStatus}");
-            if (archiveStatus == ArchiveMigrationStatus.Exported)
+            if (archiveStatus == ArchiveMigrationStatus.Failed)
             {
-                return await githubApi.GetArchiveMigrationUrl(githubSourceOrg, archiveId);
+                var retryPolicy = new RetryPolicy(_log);
+
+                var result = await retryPolicy.RetryOnResult(async () => await githubApi.GetArchiveMigrationStatus(githubSourceOrg, archiveId), ArchiveMigrationStatus.Failed);
```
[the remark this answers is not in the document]
hfishback01: But doesn't the client just check for non-200 errors?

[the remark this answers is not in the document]
ArinGhazarian: Oh I see, so in this case the response status in 200 and we're retrying for `ArchiveMigrationStatus.Failed` 👍

src/gei/Handlers/MigrateRepoCommandHandler.cs
```diff
@@ -236,13 +238,17 @@ private async Task<string> WaitForArchiveGeneration(GithubApi githubApi, string
         {
             var archiveStatus = await githubApi.GetArchiveMigrationStatus(githubSourceOrg, archiveId);
             _log.LogInformation($"Waiting for archive with id {archiveId} generation to finish. Current status: {archiveStatus}");
-            if (archiveStatus == ArchiveMigrationStatus.Exported)
+            if (archiveStatus == ArchiveMigrationStatus.Failed)
             {
-                return await githubApi.GetArchiveMigrationUrl(githubSourceOrg, archiveId);
+                var result = await _retryPolicy.RetryOnResult(async () => await githubApi.GetArchiveMigrationStatus(githubSourceOrg, archiveId), ArchiveMigrationStatus.Failed);
```
ArinGhazarian, line 243: this probably applies to all `GetArchiveMigrationStatus` calls, so it'd be better to move this retry logic into the `GithubApi` itself and eventually return the migration status. It would also be better separation of concerns.

hfishback01: I moved the retry logic into GithubApi here: 01a9f99107172c274bf39daf8c6ea6fa447e76cd

src/OctoshiftCLI.Tests/gei/Handlers/MigrateRepoCommandHandlerTests.cs
hfishback01: Dropped this test in favor of the happy path. 01a9f99107172c274bf39daf8c6ea6fa447e76cd

hfishback01:
> I believe this PR doesn't cover what #611 asks:
> 
> > We have already added retries on HTTP failures. But sometime the archive generation might fail, but GHES returns an HTTP 200 with a message telling us the archive generation failed. In that case we should restart the archive generation again.
> 
> this is an edge case an in order to do it we need to know what message is being returned and add proper unit tests for that too.

But, in case of non-HTTP failure (200 codes), won't the state be all that we need to worry about?

Per this [doc](https://docs.github.com/en/rest/migrations/orgs?apiVersion=2022-11-28#get-an-organization-migration-status), I figured we would just need to check for the failure state and retry based on that? Or am I misunderstanding?

ArinGhazarian:
> > I believe this PR doesn't cover what #611 asks:
> > > We have already added retries on HTTP failures. But sometime the archive generation might fail, but GHES returns an HTTP 200 with a message telling us the archive generation failed. In that case we should restart the archive generation again.
> > 
> > 
> > this is an edge case an in order to do it we need to know what message is being returned and add proper unit tests for that too.
> 
> @ArinGhazarian But, in case of non-HTTP failure (200 codes), won't the state be all that we need to worry about?
> 
> Per this [doc](https://docs.github.com/en/rest/migrations/orgs?apiVersion=2022-11-28#get-an-organization-migration-status), I figured we would just need to check for the failure state and retry based on that? Or am I misunderstanding?

I guess you're right, I judged solely based on the ticket and wasn't aware of the `state`.

src/Octoshift/GithubApi.cs
```diff
@@ -644,8 +644,18 @@ public virtual async Task<string> GetArchiveMigrationStatus(string org, int migr
...
 
-            return (string)data["state"];
+            if (state == ArchiveMigrationStatus.Failed)
+            {
+                var result = await _retryPolicy.RetryOnResult(async () => await GetArchiveMigrationStatus(org, migrationId), ArchiveMigrationStatus.Failed);
+
+                state = result.Outcome == Polly.OutcomeType.Failure
+                    ? throw new OctoshiftCliException($"Archive generation failed for id: {migrationId}")
+                    : result.Result;
+            }
```
ArinGhazarian, line 656: This won't work as expected and will cause an infinite loop probably because we're recursively calling the method! It would work in the handler but here since it's a lower lever we need to also retry the lower level calls like this:
```c#
public virtual async Task<string> GetArchiveMigrationStatus(string org, int migrationId)
{
    var url = $"{_apiUrl}/orgs/{org}/migrations/{migrationId}";

    var response = await _retryPolicy.RetryOnResult(async () =>
    {
        var httpResponse = await _client.GetAsync(url);
        var data = JObject.Parse(httpResponse);

        return (string)data["state"];
    }, ArchiveMigrationStatus.Failed);

    return response.Outcome == OutcomeType.Failure
        ? throw new OctoshiftCliException($"Archive generation failed for id: {migrationId}")
        : response.Result;
}
```

ArinGhazarian (changes requested):

src/Octoshift/GithubApi.cs
hfishback01: Oh! Derp- good point! Committed these changes in: 431fffdc5085540df5df70e6eaab633d6d5f89fd

src/OctoshiftCLI.Tests/gei/Handlers/MigrateRepoCommandHandlerTests.cs
```diff
@@ -51,6 +51,56 @@ public MigrateRepoCommandHandlerTests()
...
+                GhesApiUrl = GHES_API_URL,
+                AzureStorageConnectionString = AZURE_CONNECTION_STRING,
+                Wait = true
+            };
+            await FluentActions
+                .Invoking(async () => await handler.Handle(args)).Should().ThrowExactlyAsync<OctoshiftCliException>();
+
+            // Assert
+            _mockSourceGithubApi.Verify(x => x.GetArchiveMigrationStatus(SOURCE_ORG, It.IsAny<int>()), Times.Exactly(7)); // 6 times for git archive, once for metadata
+        }
```
ArinGhazarian, line 102: Now that the retry logic is moved from the handler to the api, you can just remove this test and write a brand new one for the api. It's going to be way simpler that this. There are plenty of samples in `GithubApiTests`.

src/OctoshiftCLI.Tests/gei/Handlers/MigrateRepoCommandHandlerTests.cs
```diff
@@ -122,7 +172,6 @@ public async Task Happy_Path_Without_Wait()
             _mockTargetGithubApi.Verify(m => m.GetOrganizationId(TARGET_ORG));
             _mockTargetGithubApi.Verify(m => m.CreateGhecMigrationSource(githubOrgId));
             _mockTargetGithubApi.Verify(m => m.StartMigration(migrationSourceId, githubRepoUrl, githubOrgId, TARGET_REPO, sourceGithubPat, targetGithubPat, null, null, false, false));
-
```
ArinGhazarian, line 125: nit: this is probably accidental.

src/gei/Commands/MigrateRepoCommand.cs
```diff
@@ -163,7 +163,6 @@ public override MigrateRepoCommandHandler BuildHandler(MigrateRepoCommandArgs ar
             var log = sp.GetRequiredService<OctoLogger>();
             var environmentVariableProvider = sp.GetRequiredService<EnvironmentVariableProvider>();
             var httpDownloadService = sp.GetRequiredService<HttpDownloadService>();
-
```
ArinGhazarian, line 166: nit: same

ArinGhazarian (changes requested):
The main logic looks good now the only thing is to add a retry for the api method.

src/OctoshiftCLI.Tests/GithubApiTests.cs
```diff
@@ -877,6 +877,26 @@ public async Task StartMigration_Returns_New_Repository_Migration_Id()
...
+
+            _githubClientMock.SetupSequence(x => x.GetAsync(url, null).Result)
+                .Returns(failedResponse)
+                .Returns(failedResponse)
+                .Returns(successfulResponse);
+            // Act
+            _ = await _githubApi.GetArchiveMigrationStatus(GITHUB_ORG, 1);
+
+            // Assert
+            _githubClientMock.Verify(x => x.GetAsync(url, null), Times.Exactly(3));
```
ArinGhazarian, line 897: I would also assert than we're getting the expected response:
```suggestion
            
            // Act
            var migrationStatus = await _githubApi.GetArchiveMigrationStatus(GITHUB_ORG, 1);

            // Assert
            migrationStatus.Should().Be("exported");
            _githubClientMock.Verify(x => x.GetAsync(url, null), Times.Exactly(3));
```

hfishback01: Resolved in 2e3d1a26dd8b0241b6f6e204181c0af5058c5ffd :tada:

Issue #781, opened 2023-01-16
502 Bad Gateway can cause the migration to halt


testworksau:
## Description

We attempted to run a trial migration over the weekend via the CLI (v0.31); it got ~4 hours into queuing repos and failed with:

```
[9:16 pm] [INFO] You are running the latest version of the gei CLI [v0.31]
[9:16 pm] [INFO] Migrating Repo...
[9:16 pm] [INFO] GITHUB SOURCE ORG:<ourorg>
[9:16 pm] [INFO] SOURCE REPO: <repo>
[9:16 pm] [INFO] GITHUB TARGET ORG: <ourorg>-target-trial-2
[9:16 pm] [INFO] TARGET REPO: <repo>
[9:16 pm] [ERROR] An unexpected error happened. Please see the logs for details.
```

The verbose logs show:

```
9:16 pm] [DEBUG] HTTP GET: https://raw.githubusercontent.com/github/gh-gei/main/LATEST-VERSION.txt
[9:16 pm] [DEBUG] RESPONSE (OK): v0.31

[9:16 pm] [INFO] You are running the latest version of the gei CLI [v0.31]
[9:16 pm] [INFO] Migrating Repo...
[9:16 pm] [INFO] GITHUB SOURCE ORG: <ourorg>
[9:16 pm] [INFO] SOURCE REPO: <repo>
[9:16 pm] [INFO] GITHUB TARGET ORG: <ourorg>-target-trial-2
[9:16 pm] [INFO] TARGET REPO: <repo>
[9:16 pm] [DEBUG] HTTP POST: https://api.github.com/graphql
[9:16 pm] [DEBUG] HTTP BODY: {"query":"query($login: String!) {organization(login: $login) { login, id, name } }","variables":{"login":"<ourorg>-target-trial-2"}}
[9:16 pm] [DEBUG] GITHUB REQUEST ID: F87D:52D2:159A400:16FBA90:63C12F78
[9:16 pm] [DEBUG] RESPONSE (OK): {"data":{"organization":{"login":"<ourorg>-target-trial-2","id":"O_kgDOB06WWA","name":"<ourorg>-target-trial-2"}}}
[9:16 pm] [DEBUG] HTTP POST: https://api.github.com/graphql
[...]

PR #811, opened 2023-02-10
Make generated migration scripts more robust if queueing one migration fails

timrogers:
Customers often use migration scripts generated with `gh gei generate-script` to queue up many migrations.

If queueing up one migration fails, then at the moment, the entire script explodes. This most often happens due to a transient network error when talking to GitHub.com or because archive generation fails in GitHub Enterprise Server.

This is a pretty painful experience, since customers will often leave the script running unattended (e.g. overnight), so they won't know when it terminates and they won't be able to intervene.

This changes the behaviour of our generated scripts so that, if one migration cannot be queued, we print out an error and move on to the next one.

If we have failed to queue up a migration, if applicable, it will be included in the "failed migrations" count printed at the end of the script and we will not attempt to download logs if the `--download-migration-logs` option was selected.

_I haven't written tests for this at the moment. Before I do that, I just wanted to get a review of my approach, since I wrote this code in a bit of a vacuum on a plane._

Fixes https://github.com/github/gh-gei/issues/781.

- [ ] Issue linked

### DIFF
--- a/src/gei/Handlers/GenerateScriptCommandHandler.cs
+++ b/src/gei/Handlers/GenerateScriptCommandHandler.cs
@@ -248,16 +248,19 @@ public class GenerateScriptCommandHandler : ICommandHandler<GenerateScriptComman
         content.AppendLine();
         content.AppendLine(VersionComment);
         content.AppendLine(EXEC_FUNCTION_BLOCK);
+        content.AppendLine(EXEC_AND_GET_MIGRATION_ID_FUNCTION_BLOCK);
+        content.AppendLine("$RepoMigrations = [ordered]@{}");
 
         content.AppendLine($"# =========== Organization: {githubSourceOrg} ===========");
 
         foreach (var repo in repos)
         {
-            content.AppendLine(Exec(MigrateGithubRepoScript(githubSourceOrg, githubTargetOrg, repo, ghesApiUrl, awsBucketName, noSslVerify, true, skipReleases, lockSourceRepo)));
+            content.AppendLine($"$MigrationID = {ExecAndGetMigrationId(MigrateGithubRepoScript(githubSourceOrg, githubTargetOrg, repo, ghesApiUrl, awsBucketName, noSslVerify, true, skipReleases, lockSourceRepo))}");
+            content.AppendLine($"$RepoMigrations[\"{repo}\"] = $MigrationID");
 
             if (downloadMigrationLogs)
             {
-                content.AppendLine(Exec(DownloadMigrationLogScript(githubTargetOrg, repo)));
+                content.AppendLine(DownloadMigrationLogScript(githubTargetOrg, repo));
             }
         }
 
@@ -276,7 +279,7 @@ public class GenerateScriptCommandHandler : ICommandHandler<GenerateScriptComman
 
         content.AppendLine();
         content.AppendLine("$Succeeded = 0");
-        content.AppendLine("$Failed = 0");
+        content.AppendLine("$global:failed = 0");
         content.AppendLine("$RepoMigrations = [ordered]@{}");
 
         content.AppendLine();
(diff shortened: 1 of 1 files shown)


src/gei/Handlers/GenerateScriptCommandHandler.cs
```diff
@@ -338,6 +340,8 @@ private string GenerateSequentialAdoScript(IDictionary<string, IEnumerable<strin
         content.AppendLine(VersionComment);
         content.AppendLine(EXEC_FUNCTION_BLOCK);
 
+        content.AppendLine("$RepoMigrations = [ordered]@{}");
```
timrogers, line 343: Recording this allows us to know what migrations generated an ID, and therefore what migrations we should wait for.

src/gei/Handlers/GenerateScriptCommandHandler.cs
```diff
@@ -481,11 +485,13 @@ private string GetGhesRepoOptions(string ghesApiUrl, string awsBucketName, bool
         return $"--ghes-api-url \"{ghesApiUrl}\"{(awsBucketName.HasValue() ? $" --aws-bucket-name \"{awsBucketName}\"" : "")}{(noSslVerify ? " --no-ssl-verify" : string.Empty)}";
     }
 
-    private string WaitForMigrationScript(string repoMigrationKey = null) => $"gh gei wait-for-migration --migration-id $RepoMigrations[\"{repoMigrationKey}\"]";
+    private string WaitForMigrationScript(string repoMigrationKey = null) {
+        return $"if ($RepoMigrations[\"{repoMigrationKey}\"]) {{ gh gei wait-for-migration --migration-id $RepoMigrations[\"{repoMigrationKey}\"]\nif ($lastexitcode -eq 0) {{ $Succeeded++ }} else {{ $global:failed++ }} }}";
```
timrogers, line 489: There is probably a better way to write this as a multi-line string

src/gei/Handlers/GenerateScriptCommandHandler.cs
```diff
@@ -481,11 +485,13 @@ private string GetGhesRepoOptions(string ghesApiUrl, string awsBucketName, bool
...
+    }
 
     private string DownloadMigrationLogScript(string githubTargetOrg, string targetRepo)
     {
-        return $"gh gei download-logs --github-target-org \"{githubTargetOrg}\" --target-repo \"{targetRepo}\"";
+        return $"if ($RepoMigrations[\"{targetRepo}\"]) {{ gh gei download-logs --github-target-org \"{githubTargetOrg}\" --target-repo \"{targetRepo}\" }}";
```
timrogers, line 494: We only want to download the logs if a migration ID was recorded for the repo. If no migration started, we can skip it.

src/gei/Handlers/GenerateScriptCommandHandler.cs
```diff
@@ -422,7 +427,6 @@ private string GenerateParallelAdoScript(IDictionary<string, IEnumerable<string>
             foreach (var repo in repos[teamProject].Select(r => GetGithubRepoName(teamProject, r)))
             {
                 content.AppendLine(WaitForMigrationScript(repo));
-                content.AppendLine("if ($lastexitcode -eq 0) { $Succeeded++ } else { $Failed++ }");
```
timrogers, line 425: This has now moved inside `WaitForMigrationScript`.

dylan-smith:
Can somebody include an example of a script before and after this change, easier to understand if we can look at the actual script instead of trying to mentally model the string concat logic.

src/gei/Handlers/GenerateScriptCommandHandler.cs
```diff
@@ -276,7 +279,7 @@ private string GenerateParallelGithubScript(IEnumerable<string> repos, string gi
 
         content.AppendLine();
         content.AppendLine("$Succeeded = 0");
-        content.AppendLine("$Failed = 0");
+        content.AppendLine("$global:failed = 0");
```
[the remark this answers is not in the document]
ArinGhazarian: based on what I already mentioned for `Exec` we cannot do this.

ArinGhazarian (changes requested):

ArinGhazarian:
@timrogers I am working on a new PR to address this issue for all `generate-script` commands.

timrogers:
Closing this in favour of @ArinGhazarian's upcoming implementation.  I just wrote this during a flight for a bit of fun, so I'm not at all precious about it ❤️

Issue #692, opened 2022-10-20
Add support for Security Token Service (STS)-provided temporary security credentials for S3 blob storage


timrogers:
We now allow users to use Amazon S3 for blob storage when migrating from GitHub Enterprise Server.

When calling the CLI, we accept an AWS access key, secret key and bucket name. This works well if using standard AWS IAM credentials tied to a user. 

But it doesn't work if you want to use temporary security credentials provided by AWS Security Token Service (STS) for security reasons.

We should allow these to be used with the CLI. A basic solution would just be to accept [the session token that goes with the access key and secret key in the STS case](https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/prog-services-sts.html). A more complex solution could conceivably generate credentials for you.

mattbryce93:
I've created this pull request that I believe would address this, I'm quite unfamiliar with c# so any assistance with tests and checking my code would be appreciated: https://github.com/github/gh-gei/pull/725

dylan-smith:
Hi @mattbryce93, thanks for trying to contribute! Unfortunately, it looks like your PR needs a lot more work before we could consider accepting it (e.g. it doesn't compile, doesn't pass the linter, I suspect it breaks existing scenarios, and is missing a bunch of things needed to support the STS scenario such as new args).

My best suggestion would be to find a friend that knows a bit more c# to help out, or wait until this backlog item raises to the top of our teams priority list and we can help tackle it.

timrogers:
Fixed by #844.

PR #824, opened 2023-02-19
Use AWS credential provider chain to load credentials

leonli-xero:
- Use AWS credential provider chain to load credentials that follow the order of AWS session credentials, basic credentials, and webIdentity credentials.
- [x] Issue linked
 - https://github.com/github/gh-gei/issues/692

### DIFF
--- a/src/Octoshift/AwsArgs.cs
+++ b/src/Octoshift/AwsArgs.cs
@@ -1,0 +1,10 @@
+﻿namespace OctoshiftCLI;
+
+public class AWSArgs
+{
+    public string AwsBucketName { get; set; }
+    public string AwsAccessKey { get; set; }
+    public string AwsSecretKey { get; set; }
+    public string AwsSessionToken { get; set; }
+    public string AwsCredentialType { get; set; }
+}
--- a/src/Octoshift/AwsCredentialProvider.cs
+++ b/src/Octoshift/AwsCredentialProvider.cs
@@ -1,0 +1,69 @@
+﻿using System;
+using Amazon.Runtime;
+
+namespace OctoshiftCLI;
+
+public class AwsCredentialProvider
+{
+    private readonly AWSArgs _awsArgs;
+    public AwsCredentialProvider(AWSArgs awsArgs)
+    {
+        _awsArgs = awsArgs;
+    }
+
+    public AWSCredentials GetCredentials()
+    {
+        if (string.IsNullOrEmpty(_awsArgs.AwsCredentialType)) //load credentials from provider chain
+        {
+            AWSCredentials result = SessionCredentials();
+            if (result != null)
+            {
+                return result;
+            }
+            result = BasicCredentials();
+            if (result != null)
+            {
+                return result;
+            }
+            result = WebIdentityCredentials();
+            return result ?? throw new OctoshiftCliException("No AWS credentials could be found");
+        }
+        else
+        { //use specific way to load credentials
+            return _awsArgs.AwsCredentialType == "basic"
+                ? BasicCredentials()
+                : _awsArgs.AwsCredentialType == "session"
+                    ? SessionCredentials()
+                    : _awsArgs.AwsCredentialType == "webIdentity"
+                                    ? WebIdentityCredentials()
+                                    : throw new ArgumentException("Unknown AwsCredentialType:" + _awsArgs.AwsCredentialType);
+        }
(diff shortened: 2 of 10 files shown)


src/Octoshift/AwsArgs.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace OctoshiftCLI;
+
+public class AWSArgs
+{
+    public string AwsBucketName { get; set; }
+    public string AwsAccessKey { get; set; }
+    public string AwsSecretKey { get; set; }
+    public string AwsSessionToken { get; set; }
+    public string AwsCredentialType { get; set; }
```
timrogers, line 9: How does this ever get set? I can't see any sign of it in the code - but I'm probably missing something.

src/Octoshift/AwsCredentialProvider.cs
```diff
@@ -0,0 +1,69 @@
...
+        var awsSecretKey = _awsArgs.AwsSecretKey;
+        var awsSessionToken = _awsArgs.AwsSessionToken;
+        return string.IsNullOrEmpty(awsAccessKey) || string.IsNullOrEmpty(awsSecretKey) || string.IsNullOrEmpty(awsSessionToken)
+            ? null
+            : new SessionAWSCredentials(awsAccessKey, awsSecretKey, awsSessionToken);
+    }
+
+    private AssumeRoleWithWebIdentityCredentials WebIdentityCredentials()
+    {
+        return string.IsNullOrEmpty(Environment.GetEnvironmentVariable("AWS_WEB_IDENTITY_TOKEN_FILE"))
```
timrogers, line 55: If you use this approach to authentication, do you also need to specify the role to assume? Or is that done before you use the `gh gei` CLI somehow? I'd love to understand the workflow a little better.

src/Octoshift/AwsCredentialProvider.cs
```diff
@@ -0,0 +1,69 @@
...
+    public AwsCredentialProvider(AWSArgs awsArgs)
+    {
+        _awsArgs = awsArgs;
+    }
+
+    public AWSCredentials GetCredentials()
+    {
+        if (string.IsNullOrEmpty(_awsArgs.AwsCredentialType)) //load credentials from provider chain
+        {
+            AWSCredentials result = SessionCredentials();
```
timrogers, line 18: I am guessing you would use another tool before you call `gh gei` to generate the session credentials, and then you would pass them to the `gh gei` CLI? How would you normally generate the session credentials in practice?

timrogers (commented):
Thanks for creating this separate PR! 🧡 We're not AWS experts here on the Migration Tools at GitHub, so I'm going to ask some questions about how that works. Otherwise, I'll ask the team to review this and get back to you.

src/Octoshift/AwsArgs.cs
leonli-xero: I want to use an AWSArgs object instead of passing too many args.
Setting the value in [src/gei/Commands/MigrateRepoCommand.cs ](https://github.com/leonli-xero/gh-gei/blob/Add-AWS-credential-provider/src/gei/Commands/MigrateRepoCommand.cs#L182-L189)
Use the values in the file [src/Octoshift/AwsApi.cs](https://github.com/leonli-xero/gh-gei/blob/Add-AWS-credential-provider/src/Octoshift/AwsApi.cs#L81)

src/Octoshift/AwsCredentialProvider.cs
leonli-xero: You need to assume the role before you use the` gh gei` CLI.  
We are using the EKS cluster, we want to run GEI as a pod/job in our EKS cluster with a service account. The service account binds with an AWS IAM role. When we start the pod/job EKS will help us to assume the role and prepare the credentials files and env variables in the pod, so we don't need to assume it.

src/Octoshift/AwsCredentialProvider.cs
leonli-xero: When I do some tests on my local laptop, I can use `aws cli` to get session credentials, such as `aws sts get-session-token`. And then I can store credentials as ENV variables and call `gh gei` CLI.

ArinGhazarian:
Hi @leonli-xero 👋 ,

I am closing this PR in favor of #844. Thanks for your contribution again.

PR #826, opened 2023-02-20
Support implicit AWS SDK credential resolution

### LINKED ISSUE #692: Add support for Security Token Service (STS)-provided temporary security credentials for S3 blob storage

### LINKED ISSUE #738: Align AWS cli args and environment vars to aws cli
## Description

Currently the `migrate-repo` command supports the following cli args:
`--aws-access-key <aws-access-key>`
`--aws-secret-key <aws-secret-key>`

... or the following environment variables:
`AWS_ACCESS_KEY`
`AWS_SECRET_KEY`

AWS cli uses the following environment variables:
`AWS_ACCESS_KEY_ID`
`AWS_SECRET_ACCESS_KEY`

To make this more familiar to AWS users, it would be best to align these cli arg and environment variables to the environment variables that the aws cli uses.

Please add support for the following cli arg:
`--aws-access-key-id <aws-access-key-id>`
`--aws-secret-access-key <aws-secret-access-key>`

Please also add support for the following environment variables:
`AWS_ACCESS_KEY_ID`
`AWS_SECRET_ACCESS_KEY`

Please update the gei cli help message for the `migrate-repo` command to reflect these new preferred args and environment variables. Please also update the early access documentation.

** Please do not remove support for the existing cli args and environment variables because these might be in use by customers who are migrating now. Best to add support for the new stuff, update the docs, and wait until a time when most users who might be migrating now are done. Only then, make the breaking change to remove support for the existing cli args and environment variables.

derwasp:
## Background

AWS .NET SDK provides a way to automatically get the credentials based on conventions defined in [this document](https://docs.aws.amazon.com/sdk-for-net/v3/developer-guide/creds-assign.html).
This is how it works when you are creating any entity that works with the API.

This is meant to "just work" in AWS environments: be it lambda, ECS, EC2 or anything else. It works just as well with the local profiles and environment variables.

Basically, using this approach would be interoperable with everything else that usually works with AWS CLI.

For example, this should also allow to interop with solutions like `aws-vault` where these credentials would be basically injected into your application whenever you run this.

I tried to provide compatibility with the old approach, too though I would just deprecate this over this more standard way of doing things.

One bit that is still missing is the `AWS_REGION`: as far as I remember (and it's been a while since I used it) this is the variable that you can rely on to get the region if it's set. (Just checked: [yeah, here it is](https://github.com/aws/aws-sdk-net/blob/d8218832bfe590f858fe195faaf57eeceb8a1ee6/sdk/src/Core/Amazon.Runtime/AWSRegion.cs#L74)).

I am not making this change here and I am not changing the behavior of just using the default Virginia region here, though I would recommend doing it.

## Proposed changes

* Use FallbackCredentialsFactory instead of always relying on basic credentials: this code can be found [here](https://github.com/aws/aws-sdk-net/blob/master/sdk/src/Services/S3/Generated/_bcl45/AmazonS3Client.cs#L80)
* Kind of keep backward compatibility with checking the parameters: I am just explicitly creating the credentials and if an exception is thrown, I fail
* Add a deprecation note to for the old variables

## Notes

More information on credential resolution here:
https://docs.aws.amazon.com/sdk-for-net/v3/developer-guide/creds-assign.html

Closes #738
Closes #692

Supersedes #824 #731

- [x] Issue linked

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -1,1 +1,2 @@
 - Create shared access signature (SAS) with read-only permissions - not read-write - when generating Azure Blob Storage URL
+- Add support for Implicit Credentials Resolution for the S3 bucket: https://docs.aws.amazon.com/sdk-for-net/v3/developer-guide/creds-assign.html
--- a/src/Octoshift/AwsApi.cs
+++ b/src/Octoshift/AwsApi.cs
@@ -17,9 +17,9 @@ public class AwsApi : IDisposable
     private readonly ITransferUtility _transferUtility;
 
 #pragma warning disable CA2000
-    public AwsApi(AWSCredentials credentials) 
+    public AwsApi(AWSCredentials credentials)
         : this(new AmazonS3Client(credentials, RegionEndpoint))
-            
+
 #pragma warning restore CA2000
     {
     }
--- a/src/Octoshift/AwsApi.cs
+++ b/src/Octoshift/AwsApi.cs
@@ -2,6 +2,7 @@
 using System.IO;
 using System.Threading.Tasks;
 using Amazon;
+using Amazon.Runtime;
 using Amazon.S3;
 using Amazon.S3.Model;
 using Amazon.S3.Transfer;
@@ -16,11 +17,14 @@ public class AwsApi : IDisposable
     private readonly ITransferUtility _transferUtility;
 
 #pragma warning disable CA2000
-    public AwsApi(string awsAccessKey, string awsSecretKey) : this(new TransferUtility(new AmazonS3Client(awsAccessKey, awsSecretKey, RegionEndpoint)))
+    public AwsApi(AWSCredentials credentials) 
+        : this(new AmazonS3Client(credentials, RegionEndpoint))
+            
 #pragma warning restore CA2000
     {
     }
 
+    private AwsApi(IAmazonS3 s3Client) : this(new TransferUtility(s3Client)) { }
     internal AwsApi(ITransferUtility transferUtility) => _transferUtility = transferUtility;
 
     public virtual async Task<string> UploadToBucket(string bucketName, string fileName, string keyName)
--- a/src/Octoshift/EnvironmentVariableProvider.cs
+++ b/src/Octoshift/EnvironmentVariableProvider.cs
@@ -8,8 +8,6 @@ public class EnvironmentVariableProvider
     private const string TARGET_GH_PAT = "GH_PAT";
     private const string ADO_PAT = "ADO_PAT";
     private const string AZURE_STORAGE_CONNECTION_STRING = "AZURE_STORAGE_CONNECTION_STRING";
-    private const string AWS_ACCESS_KEY = "AWS_ACCESS_KEY";
-    private const string AWS_SECRET_KEY = "AWS_SECRET_KEY";
     private const string BBS_USERNAME = "BBS_USERNAME";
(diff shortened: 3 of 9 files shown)


derwasp:
It seems like there's some kind of a default profile on the boxes that gets picked up by the AWS SDK. This means the credentials are not really empty :( Hence the current tests are failing because of that.
I am not sure how to proceed with this in case you would like me to continue on this proposal.

derwasp:
Okay, it seems like this is the normal behavior for the sdk 🤔 

[Here](https://github.com/aws/aws-sdk-net/blob/c9817f68151aae00173ade0b71b9b589902ad5af/sdk/src/Core/Amazon.Runtime/Credentials/FallbackCredentialsFactory.cs#L116) it seems to be always falling back to the `DefaultInstanceProfileAWSCredentials`.
This means it's not throwing while I expected it to.

Do we want to just rely on the SDK auth and not expect this to crash? This means we would have to remove the tests that expect a failure.

derwasp:
I am not sure why CodeQL is failing: if anyone can let me know how to fix it, I am happy to do so :)

derwasp:
tagging @timrogers
the PR is kind of ready for review now folks :)

derwasp:
I found the codeql issue, should be fixed now.

ArinGhazarian:
Hi @derwasp 👋,

Thanks for your great contribution and the very useful info. I very much enjoyed reading the PR writeup 😃. After lots of internal discussions we decided to take the explicit approach and just add the support for AWS session token auth since we believe that it will cover almost all auth scenarios.

derwasp:
@ArinGhazarian thank you for your message! It's great that you've added support for it!

Yeah, it would cover some scenarios.

I just want to put this message out there for people reading up on this later: the session token parameter would still not work with tools like `aws-vault` out of the box because your are using non-standard environment variables for the other variables.
This would also not work in the EC2/ECS environments as well as this won't work with the default profiles.

So this [resolution strategy](https://docs.aws.amazon.com/sdk-for-net/v3/developer-guide/creds-assign.html) won't be applicable.

But this would indeed allow to pass the temporary credentials in.

timrogers:
> @ArinGhazarian thank you for your message! It's great that you've added support for it!
> 
> Yeah, it would cover some scenarios.
> 
> I just want to put this message out there for people reading up on this later: the session token parameter would still not work with tools like `aws-vault` out of the box because your are using non-standard environment variables for the other variables. This would also not work in the EC2/ECS environments as well as this won't work with the default profiles.
> 
> So this [resolution strategy](https://docs.aws.amazon.com/sdk-for-net/v3/developer-guide/creds-assign.html) won't be applicable.
> 
> But this would indeed allow to pass the temporary credentials in.

Just so you know, we're planning to address the names of those arguments next - so hopefully after that, your vault will just work!

PR #830, opened 2023-02-21
Use read-only permissions when creating a SAS to generate Azure Blob Storage SAS URL

timrogers:
After uploading an archive to Azure Blob Storage, we need to generate a temporary URL for the file to send to the GitHub API. In the world of Azure, this is called a "SAS URL".

To be able to create a SAS URL, we create a shared access signature (SAS). 

At the moment, the SAS is created with read-write permissions, which is unnecessary, given that we only need to be able to read the file.

This changes it to use read-only permissions.

- [ ] Issue linked

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -1,1 +1,1 @@
-
+- Create shared access signature (SAS) with read-only permissions - not read-write - when generating Azure Blob Storage URL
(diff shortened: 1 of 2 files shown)


dylan-smith (approved):

timrogers:
Tested locally and this works ✅

PR #833, opened 2023-02-22
Fix NullReferenceException when running on Bitbucket server

### LINKED ISSUE #745: `gh bbs2gh migrate-repo` explodes without the `--ssh-user` and `--ssh-private-key` arguments
## Description

With `gh bbs2gh`, the intent is that you can run the `migrate-repo` command on a BBS server itself and then the file will be immediately available, instead of having to use SSH to download it.

However, if you run `gh bbs2gh migrate-repo` without the `--ssh-user` and `--ssh-private-key` arguments, you get an error like this, rather than the migration continuing:

```
[11:24] [INFO] Export completed. Your migration archive should be ready on your instance at $BITBUCKET_SHARED_HOME/data/migration/export/Bitbucket_export_14.tar
[11:24] [ERROR] System.NullReferenceException: Object reference not set to an instance of an object.
```

## Reproduction Steps

Use `gh bbs2gh migrate-repo` without `--ssh-user` and/or `--ssh-private-key`, e.g.:

```
gh bbs2gh migrate-repo --bbs-server-url "http://bbs.eastus.cloudapp.azure.com:7990" --bbs-project "OS" --bbs-repo "airports" --ssh-port 22 --github-org "Deep-Avocado-9" --github-repo "airports" --verbose --wait --aws-bucket-name "timrogers-bbs2gh-ie"
```

ArinGhazarian:
Fixes #745

We support a scenario where the CLI can be run on the Bitbucket server itself. In that case there will be naturally no download options and as a result the BBS downloader service won't be instantiated and will be injected into the handler as `null`. This was causing a `NullReferenceException` in the upload phase since we were trying to retrieve the source archive's path from an `IBbsArchiveDownloader` instance which was indeed `null`. To fix it, I basically moved the `GetSourceExportArchiveAbsolutePath()` method from a concrete implementation into the abstraction (`IBbsArchiveDownloader`) and added the `bbsSharedHomeDirectory` as an additional param to it. The result is a new static method in the interface the doesn't depend on a concrete implementation. 

Now in order to get the default Bitbucket's shared home directory (when `--bbs-shared-home` is `null`) in the handler, I detect the server's OS version and then get the default BBS home from the `BbsSettings` class.  

- [x] Issue linked

### DIFF
--- a/src/bbs2gh/Handlers/MigrateRepoCommandHandler.cs
+++ b/src/bbs2gh/Handlers/MigrateRepoCommandHandler.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Linq;
+using System.Runtime.InteropServices;
 using System.Threading.Tasks;
 using OctoshiftCLI.BbsToGithub.Commands;
 using OctoshiftCLI.BbsToGithub.Services;
@@ -74,7 +75,7 @@ public class MigrateRepoCommandHandler : ICommandHandler<MigrateRepoCommandArgs>
             // This is for the case where the CLI is being run on the BBS server itself
             if (args.ArchivePath.IsNullOrWhiteSpace())
             {
-                args.ArchivePath = _bbsArchiveDownloader.GetSourceExportArchiveAbsolutePath(exportId);
+                args.ArchivePath = GetSourceExportArchiveAbsolutePath(args.BbsSharedHome, exportId);
             }
 
             args.ArchiveUrl = args.AwsBucketName.HasValue()
@@ -86,6 +87,15 @@ public class MigrateRepoCommandHandler : ICommandHandler<MigrateRepoCommandArgs>
         {
             await ImportArchive(args, args.ArchiveUrl);
         }
+    }
+
+    private string GetSourceExportArchiveAbsolutePath(string bbsSharedHomeDirectory, long exportId)
+    {
+        static string GetDefaultBbsSharedHomeDirectory() => RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
+            ? BbsSmbArchiveDownloader.DEFAULT_BBS_SHARED_HOME_DIRECTORY
+            : BbsSshArchiveDownloader.DEFAULT_BBS_SHARED_HOME_DIRECTORY;
+
+        return IBbsArchiveDownloader.GetSourceExportArchiveAbsolutePath(bbsSharedHomeDirectory ?? GetDefaultBbsSharedHomeDirectory(), exportId);
     }
 
     private bool ShouldGenerateArchive(MigrateRepoCommandArgs args)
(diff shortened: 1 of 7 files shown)


src/bbs2gh/Handlers/MigrateRepoCommandHandler.cs
```diff
@@ -88,6 +89,15 @@ public async Task Handle(MigrateRepoCommandArgs args)
         }
     }
 
+    private string GetSourceExportArchiveAbsolutePath(string bbsSharedHomeDirectory, long exportId)
+    {
+        static string GetDefaultBbsSharedHomeDirectory() => RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
+            ? BbsSmbArchiveDownloader.DEFAULT_BBS_SHARED_HOME_DIRECTORY
+            : BbsSshArchiveDownloader.DEFAULT_BBS_SHARED_HOME_DIRECTORY;
+
```
dylan-smith, line 97: Wouldn't this function be better living in IBbsArchiveDownloader?

ArinGhazarian: Well actually a good idea, I can create two constant one for windows and one for linux

dylan-smith: Might be a good argument for having this as a base class instead of an interface. Then you could have a virtual `DEFAULT_BBS_SHARED_HOME_DIRECTORY` that you overload in each subclass.

Having an interface (or base class) have to "know" about every subclass/implementation seems bad.

ArinGhazarian: I still like an interface better. The problem is not overriding the default shared directory since it can be easily done with a property in the interface too. The problem is that I need to access the constants outside of the concrete implementations too, in our case in the handler class. Basically depending on the OS BBS has two default home directories.
I originally took the constants out of the interface and put them into the concrete implementations because as you also mentioned I didn't want the interface to know about the implementations. But the ultimate problem is that I want to have these constants defined only in one place. So I have the following options now:
1. Leave everything and access the default shared home through a concrete downloader implementation. 
2. Create a class to hold these constants (e.g. `BbsSettings`, `Defaults`, ...).
3. Move the constants back into the interface and rename them to indicate one is for linux and the other for windows.
4. Create a service (e.g. `BbsDefaultHomeDirectoryProvider`) --> this is an overkill

Now with everything being said, I am personally leaning toward option 1 and 3. What do you think ?

ArinGhazarian: Also as a quick side note, using ssh doesn't necessarily mean that the BBS's OS is linux! Ppl can opt-in to use ssh on a windows server too.

timrogers:
I'm not going to review this one - y'all are the experts on the architecture of the CLI, and there isn't any product-y stuff in the code 🧡

dylan-smith (approved):

Issue #843, opened 2023-02-28
Add AWS region and session token args


ArinGhazarian:
Currently CLI only supports [basic AWS credentials](https://docs.aws.amazon.com/sdk-for-net/v3/developer-guide/creds-assign.html) to connect to a AWS S3 bucket. In order to support a wider range of authentication scenarios we need to also support session AWS credentials. Additionally we are going to switch to V4 signature and add AWS region arg. If the AWS region is not specified we are not going to use the V4 signature and will spit out a warning message. 

## Todo
- [x] Add optional `--aws-session-token` arg to `migrate-repo` commands for both gei and bbs2gh.
- [x] Default to signature V4.
- [x] Add (optional for now) `--aws-region` to both `migrate-repo` and `generate-script` commands. 
- [x] If `--aws-region` is not specified, don't use signature V4 and log a warning message that  an AWS region will be required in a future release (let's say from June 1st, giving 3 months). 

## Related
- #744

Issue #744, opened 2022-11-29
Switch to using V2 signing for AWS pre-signed URLs, which is available in all AWS regions


timrogers:
## Description

When migrating from GitHub Enterprise Server to GitHub Enterprise Cloud using the CLI, we:

1. Generate migration archives on the GHES instance
2. Download the migration archives to the machine where the CLI is running
3. Upload the migration archives to a blob storage provider (AWS or Azure)
4. Generate pre-signed URLs for the archive files that GHEC can use to download them
5. Trigger the migration in GHEC, passing in the pre-signed URLs for the archives

When using AWS S3 as the blob storage provider, we seem to use a URL signing algorithm, the so-called V2, that is [only supported in certain AWS regions](https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-authenticating-requests.html).

If your bucket is located in a region where V2 is not supported (e.g. `eu-west-2`), when GitHub tries to access the generated URL to download the archive, it gets a `400 Bad Request`.

This can be reproduced by trying to access the URL in your own browser, which responses a response like this:

```
HTTP/2 400 Bad Request

<Error>
<Code>InvalidRequest</Code>
<Message>The authorization mechanism you have provided is not supported. Please use AWS4-HMAC-SHA256.</Message>
<RequestId>MGRND6804N48D125</RequestId>
<HostId>qOTkJDLbVc7Pre+D2AoELHSXzpoRj4rKntkMo4pC+j58GJlCVpBA2towOZAjIHOykHXHWgh2770=</HostId>
</Error>
```

We should switch to using the V4 signing algorithm which is supported in all AWS regions.

## Reproduction Steps

Create an AWS S3 bucket in the London (`eu-west-2`) region and try to use it for a GitHub Enterprise Server to GitHub Enterprise Cloud migration. Your migration will fail with a `Git source migration failed. Error message: 400 Bad Request` error.

timrogers:
This issue contains a potential solution: https://github.com/aws/aws-sdk-net/issues/1826. 

It's very odd that this is so unobvious in AWS's docs and that the defaults don't work in all regions.

timrogers:
It's suggested here that any new bucket as of 2010 will exhibit this problem, but I guess that must have rowed back on that, as I created a bucket yesterday that seemed to allow V2: https://aws.amazon.com/blogs/aws/amazon-s3-update-sigv2-deprecation-period-extended-modified/

timrogers:
[...]

PR #844, opened 2023-03-01
Add support of AWS session token

ArinGhazarian:
Closes: #843 
Fixes: #744 

- [x] Issue linked

### Description

This PR adds support of the long awaited AWS session token authentication to `bbs2gh migrate-repo` and `gei migrate-repo` commands. Additionally the AWS region can also be passed into both `migrate-repo` and `generate-script` for `bbs2gh` and `gei`. 

This PR naturally obsoletes #826, #731 and #824. Thanks @leonli-xero and @derwasp for your invaluable contributions.


RELEASENOTES.md
```diff
@@ -1 +1,2 @@
-- Create shared access signature (SAS) with read-only permissions - not read-write - when generating Azure Blob Storage URL
\ No newline at end of file
+- Create shared access signature (SAS) with read-only permissions - not read-write - when generating Azure Blob Storage URL
+- Add support of AWS session token and region when using AWS S3 for archive upload in `gh gei migrate-repo` and `gh bbs2gh migrate-repo`. AWS region can also be passed into `gh gei generate-script` and `gh bbs2gh generate-script` commands.
```
timrogers, line 2:
```suggestion
- Add support for authenticating with AWS session tokens when using AWS S3 for archive upload in `gh gei` and `gh bbs2gh`. When specifying a session, the AWS region must also be specified.
```

src/OctoshiftCLI.Tests/bbs2gh/Handlers/GenerateScriptCommandHandlerTests.cs
```diff
@@ -344,7 +345,7 @@ public async Task Generated_Script_Contains_Exec_Function_Block()
     }
 
     [Fact]
-    public async Task One_Repo_With_Aws_Bucket_Name()
+    public async Task One_Repo_With_Aws_Bucket_Name_And_Region()
```
timrogers, line 348: Should we have a test which covers what happens if you don't submit a region? (I think we said that the region would be required for `generate-script`, but not `migrate-repo`.)

src/OctoshiftCLI.Tests/gei/Handlers/GenerateScriptCommandHandlerTests.cs
```diff
@@ -1408,7 +1409,7 @@ public async Task Sequential_Ghes_Single_Repo_Aws_S3()
                 .Setup(m => m.GetRepos(SOURCE_ORG))
                 .ReturnsAsync(new[] { REPO });
 
-            var expected = $"Exec {{ gh gei migrate-repo --github-source-org \"{SOURCE_ORG}\" --source-repo \"{REPO}\" --github-target-org \"{TARGET_ORG}\" --target-repo \"{REPO}\" --ghes-api-url \"{ghesApiUrl}\" --aws-bucket-name \"{AWS_BUCKET_NAME}\" --wait }}";
+            var expected = $"Exec {{ gh gei migrate-repo --github-source-org \"{SOURCE_ORG}\" --source-repo \"{REPO}\" --github-target-org \"{TARGET_ORG}\" --target-repo \"{REPO}\" --ghes-api-url \"{ghesApiUrl}\" --aws-bucket-name \"{AWS_BUCKET_NAME}\" --aws-region \"{AWS_REGION}\" --wait }}";
```
timrogers, line 1412: Should we have a test which covers what happens if you don't submit a region? (I think we said that the region would be required for `generate-script`, but not `migrate-repo`.)

src/bbs2gh/Handlers/MigrateRepoCommandHandler.cs
```diff
@@ -495,10 +513,15 @@ private void ValidateUploadOptions(MigrateRepoCommandArgs args)
             {
                 throw new OctoshiftCliException("Either --aws-secret-key or AWS_SECRET_KEY environment variable must be set.");
             }
+
+            if (!GetAwsRegion(args).HasValue())
+            {
+                _log.LogWarning("Please consider providing --aws-region or AWS_REGION environment variable. It will be required in future releases.");
```
timrogers, line 519: > Specifying an AWS region with the --aws-region argument or AWS_REGION environment variable is currently not required, but will be required in a future release. Defaulting to us-east-1.

src/bbs2gh/Handlers/MigrateRepoCommandHandler.cs
```diff
@@ -495,10 +513,15 @@ private void ValidateUploadOptions(MigrateRepoCommandArgs args)
...
+            }
         }
-        else if (args.AwsAccessKey.HasValue() || args.AwsSecretKey.HasValue())
+        else if (new[] { args.AwsAccessKey, args.AwsSecretKey, args.AwsSessionToken, args.AwsRegion }.Any(x => x.HasValue()))
         {
-            throw new OctoshiftCliException("--aws-access-key and --aws-secret-key can only be provided with --aws-bucket-name.");
+            throw new OctoshiftCliException("Amazon S3 upload options can only be provided with --aws-bucket-name.");
```
timrogers, line 524:
```suggestion
            throw new OctoshiftCliException("The AWS S3 bucket name must be provided with --aws-bucket-name if other AWS S3 upload options are set.");
```

src/gei/Handlers/MigrateRepoCommandHandler.cs
```diff
@@ -464,6 +483,11 @@ private void ValidateGHESOptions(MigrateRepoCommandArgs args, bool cloudCredenti
                 {
                     throw new OctoshiftCliException("Either --aws-secret-key or AWS_SECRET_KEY environment variable must be set.");
                 }
+
+                if (!GetAwsRegion(args).HasValue())
+                {
+                    _log.LogWarning("Please consider providing --aws-region or AWS_REGION environment variable. It will be required in future releases.");
```
timrogers, line 489: > Specifying an AWS region with the --aws-region argument or AWS_REGION environment variable is currently not required, but will be required in a future release. Defaulting to us-east-1.

src/gei/Commands/MigrateRepoCommand.cs
```diff
@@ -100,6 +102,14 @@ public class MigrateRepoCommand : CommandBase<MigrateRepoCommandArgs, MigrateRep
         {
             Description = "If uploading to S3, the AWS secret key. If not provided, it will be read from AWS_SECRET_KEY environment variable. Not required if migrating from GitHub Enterprise Server 3.8.0 or later."
         };
+        public Option<string> AwsSessionToken { get; } = new("--aws-session-token")
+        {
+            Description = "If using AWS, the AWS session token. If not provided, it will be read from AWS_SESSION_TOKEN environment variable."
+        };
+        public Option<string> AwsRegion { get; } = new("--aws-region")
+        {
+            Description = "If using AWS, the AWS region. If not provided, it will be read from AWS_REGION environment variable. Defaults to us-east-1."
```
timrogers, line 111:
```suggestion
            Description = "If using AWS, the AWS region. If not provided, it will be read from AWS_REGION environment variable. Defaults to us-east-1 if neither the argument nor the environment variable is set. In a future release, you will be required to set an AWS region if using AWS S3 as your blob storage provider."
```

src/gei/Commands/GenerateScriptCommand.cs
```diff
@@ -101,6 +102,11 @@ public class GenerateScriptCommand : CommandBase<GenerateScriptCommandArgs, Gene
             Description = "If using AWS, the name of the S3 bucket to upload the BBS archive to."
         };
 
+        public Option<string> AwsRegion { get; } = new("--aws-region")
+        {
+            Description = "If using AWS, the AWS region. If not provided, it will be read from AWS_REGION environment variable. Defaults to us-east-1."
```
timrogers, line 107:
```suggestion
            Description = "If using AWS, the AWS region. If not provided, it will be read from AWS_REGION environment variable. Defaults to us-east-1 if neither the argument nor the environment variable is set. In a future release, you will be required to set an AWS region if using AWS S3 as your blob storage provider."
```

src/bbs2gh/Handlers/MigrateRepoCommandHandler.cs
```diff
@@ -495,10 +513,15 @@ private void ValidateUploadOptions(MigrateRepoCommandArgs args)
             {
                 throw new OctoshiftCliException("Either --aws-secret-key or AWS_SECRET_KEY environment variable must be set.");
             }
+
+            if (!GetAwsRegion(args).HasValue())
```
timrogers, line 517: Can we add a validation that requires the user to set their AWS region if they provide a session token? We're only trying to preserve backwards compatibility for old scripts really, and if you specify a session token, then it's a good time to force the region too I think.

src/OctoshiftCLI.Tests/bbs2gh/Handlers/MigrateRepoCommandHandlerTests.cs
```diff
@@ -891,7 +893,39 @@ public async Task It_Throws_When_Aws_Bucket_Name_Not_Provided_But_Aws_Secret_Key
             }))
                 .Should()
                 .ThrowAsync<OctoshiftCliException>()
-                .WithMessage("*--aws-access-key*--aws-secret-key*");
+                .WithMessage("*Amazon S3*--aws-bucket-name*");
+        }
+
+        [Fact]
+        public async Task It_Throws_When_Aws_Bucket_Name_Not_Provided_But_Aws_Session_Token_Provided()
```
timrogers, line 900: Should we also add a test for the case where a session token is provided but a region is not provided?

timrogers:
@ArinGhazarian Thanks for smashing this out ❤️

dylan-smith (changes requested):
I gave it a look over and I don't have anything to add beyond what @timrogers suggested. I agree that we should make region required on migrate-repo if session-token is provided. And that region should be required on generate-script (if using AWS). And that we should have tests for both with and without region for these scenarios.

src/OctoshiftCLI.Tests/bbs2gh/Handlers/GenerateScriptCommandHandlerTests.cs
ArinGhazarian: In `generate-script` so far we have done very minimal validations the actual validations are going to happen in `migrate-repo` so there is no need to repeat them in `generate-script` too unless it's crucial to `generate-script` itself.

src/OctoshiftCLI.Tests/gei/Handlers/GenerateScriptCommandHandlerTests.cs
ArinGhazarian: see [this](https://github.com/github/gh-gei/pull/844#discussion_r1123632495)

src/OctoshiftCLI.Tests/bbs2gh/Handlers/MigrateRepoCommandHandlerTests.cs
ArinGhazarian: Not now, because the region is actually optional in our code so it's totally valid to provide the session token but not the region. that being said I can add the test and make sure that we're printing the log message and it future once we made the region required we can just update the test to check for the validation.

src/OctoshiftCLI.Tests/bbs2gh/Handlers/GenerateScriptCommandHandlerTests.cs
ArinGhazarian: Oh I see, it makes perfect sense to make region required if session token is passed! I will make that change in `migrate-repo` and add the validation. I still don't think we need to add extra validations in `generate-script` though.

RELEASENOTES.md
ArinGhazarian: adressed in 98de164b93e6c5ad0a9f24c52673cd80b4f9a7fc

src/OctoshiftCLI.Tests/bbs2gh/Handlers/GenerateScriptCommandHandlerTests.cs
ArinGhazarian: Addressed in 98de164b93e6c5ad0a9f24c52673cd80b4f9a7fc

src/bbs2gh/Handlers/MigrateRepoCommandHandler.cs
ArinGhazarian: Addressed in 98de164b93e6c5ad0a9f24c52673cd80b4f9a7fc

src/bbs2gh/Handlers/MigrateRepoCommandHandler.cs
ArinGhazarian: Addressed in 98de164b93e6c5ad0a9f24c52673cd80b4f9a7fc

src/gei/Handlers/MigrateRepoCommandHandler.cs
ArinGhazarian: Addressed in 98de164b93e6c5ad0a9f24c52673cd80b4f9a7fc

src/gei/Commands/MigrateRepoCommand.cs
ArinGhazarian: Addressed in 98de164b93e6c5ad0a9f24c52673cd80b4f9a7fc

src/gei/Commands/GenerateScriptCommand.cs
ArinGhazarian: Addressed in 98de164b93e6c5ad0a9f24c52673cd80b4f9a7fc

src/bbs2gh/Handlers/MigrateRepoCommandHandler.cs
ArinGhazarian: Addressed in 98de164b93e6c5ad0a9f24c52673cd80b4f9a7fc

ArinGhazarian:
@dylan-smith and @timrogers, I addressed all your comments with one exception. For `generate-script` command we haven't added exhaustive validations as we did for `migrate-repo` command unless required for the `generate-script`'s functionality itself. 
I don't think we need to duplicate the `migrate-repo` validations for `generate-script` but If you guys think that we should, then it'd be better to address it in a separate issue because the validations will be almost identical to `migrate-repo` and we probably need to extract them into its own class to be DRY.

src/Octoshift/AwsApi.cs
```diff
@@ -5,24 +5,40 @@
...
     }
 
     internal AwsApi(ITransferUtility transferUtility) => _transferUtility = transferUtility;
 
+    private static AmazonS3Client BuildAmazonS3Client(string awsAccessKey, string awsSecretKey, string awsRegion, string awsSessionToken)
+    {
+        var regionEndpoint = DefaultRegionEndpoint;
+        if (awsRegion.HasValue())
+        {
+            regionEndpoint = RegionEndpoint.GetBySystemName(awsRegion);
```
timrogers, line 33: Will this throw an exception if an invalid region name is specified? I just want to make sure I understand our failure modes!

src/Octoshift/AwsApi.cs
```diff
@@ -5,24 +5,40 @@
...
+            AWSConfigsS3.UseSignatureVersion4 = true;
```
timrogers, line 34: ❤️

timrogers (approved):
Looking good to me ❇️ But let's wait for a review from Dylan too.

timrogers:
> @dylan-smith and @timrogers, I addressed all your comments with one exception. For `generate-script` command we haven't added exhaustive validations as we did for `migrate-repo` command unless required for the `generate-script`'s functionality itself. I don't think we need to duplicate the `migrate-repo` validations for `generate-script` but If you guys think that we should, then it'd be better to address it in a separate issue because the validations will be almost identical to `migrate-repo` and we probably need to extract them into its own class to be DRY.

My personal view is that we should have more exhaustive validations at the script generation stage - but maybe I'm being too conservative, so I'd love to hear Dylan's thoughts. In any event, it can be a follow-up PR!

dylan-smith (approved):

dylan-smith:
> > @dylan-smith and @timrogers, I addressed all your comments with one exception. For `generate-script` command we haven't added exhaustive validations as we did for `migrate-repo` command unless required for the `generate-script`'s functionality itself. I don't think we need to duplicate the `migrate-repo` validations for `generate-script` but If you guys think that we should, then it'd be better to address it in a separate issue because the validations will be almost identical to `migrate-repo` and we probably need to extract them into its own class to be DRY.
> 
> My personal view is that we should have more exhaustive validations at the script generation stage - but maybe I'm being too conservative, so I'd love to hear Dylan's thoughts. In any event, it can be a follow-up PR!

I agree, but I'm not sure what we're missing. `generate-script` only takes `--aws-bucket-name` and `--aws-region` and region is optional so there's not a whole lot of validation needed. What am I missing?

timrogers:
> > > @dylan-smith and @timrogers, I addressed all your comments with one exception. For `generate-script` command we haven't added exhaustive validations as we did for `migrate-repo` command unless required for the `generate-script`'s functionality itself. I don't think we need to duplicate the `migrate-repo` validations for `generate-script` but If you guys think that we should, then it'd be better to address it in a separate issue because the validations will be almost identical to `migrate-repo` and we probably need to extract them into its own class to be DRY.
> > 
> > 
> > My personal view is that we should have more exhaustive validations at the script generation stage - but maybe I'm being too conservative, so I'd love to hear Dylan's thoughts. In any event, it can be a follow-up PR!
> 
> I agree, but I'm not sure what we're missing. `generate-script` only takes `--aws-bucket-name` and `--aws-region` and region is optional so there's not a whole lot of validation needed. What am I missing?

That's a good point. I think that probably is everything, actually. I'll do some more testing to confirm, and we can always have a follow-up PR.

ArinGhazarian:
> > > @dylan-smith and @timrogers, I addressed all your comments with one exception. For `generate-script` command we haven't added exhaustive validations as we did for `migrate-repo` command unless required for the `generate-script`'s functionality itself. I don't think we need to duplicate the `migrate-repo` validations for `generate-script` but If you guys think that we should, then it'd be better to address it in a separate issue because the validations will be almost identical to `migrate-repo` and we probably need to extract them into its own class to be DRY.
> > 
> > 
> > My personal view is that we should have more exhaustive validations at the script generation stage - but maybe I'm being too conservative, so I'd love to hear Dylan's thoughts. In any event, it can be a follow-up PR!
> 
> I agree, but I'm not sure what we're missing. `generate-script` only takes `--aws-bucket-name` and `--aws-region` and region is optional so there's not a whole lot of validation needed. What am I missing?

Well for AWS it only takes those two but for validating them we need to do extra stuff, for example we don't pass secrets via args in `generate-script` so we need to read the env. variables to do the validations. Also there are other args that we haven't validated and if we decide to go this route we need to add validations for them too that eventually leads to basically duplicating the validation for `migrate-repo`. Also there's always a possibility that they modify the script and add/remove stuff.

dylan-smith:
> 
> Well for AWS it only takes those two but for validating them we need to do extra stuff, for example we don't pass secrets via args in `generate-script` so we need to read the env. variables to do the validations. Also there are other args that we haven't validated and if we decide to go this route we need to add validations for them too that eventually leads to basically duplicating the validation for `migrate-repo`. Also there's always a possibility that they modify the script and add/remove stuff.

I would want generate-script to work without requiring those env vars to be set. They have to be set before the script can be run, but it should be able to be generated without credentials set.

What other args haven't we validated?

ArinGhazarian:
> I would want generate-script to work without requiring those env vars to be set. They have to be set before the script can be run, but it should be able to be generated without credentials set.
> 

That's exactly my point (and why I think it'd be overkill to do validations for `generate-script`). Most of those validations need the env. variable to be set. For example `aws-region` is only required (for now) if the `AWS_SESSION_TOKEN` is set. And once we made the aws region required we're gonna mark it so in the command so we'll get the validation for free.

> What other args haven't we validated?
For example the ssh and smb options in `bbs2gh`, etc.

dylan-smith:
In generate-script I thought we made aws-region required if aws-bucket was set, but I see now we didn't. I'm struggling to see what other validations would be necessary in gei generate-script. I wouldn't want the session token to be needed to be provided at script generation time.

For bbs2gh generate-script I just noticed we have ZERO validation. That's a problem we should fix.

src/Octoshift/AwsApi.cs
ArinGhazarian: @timrogers I am not sure how I missed this comment, but this is a very valid question. I was naturally under the impression that `GetBySystemName()` would throw if the region string was invalid but interestingly that's not the case! `GetBySystemName()` always succeeds even for an invalid region name! Not sure why they don't validate it! The migration fails eventually but we should def. validate and throw if the region is invalid. Fail fast, fail early 😄 
I am going to put together a quick fix for this. Thanks for pointing this out 🙏

ArinGhazarian: PR for the fix --> https://github.com/github/gh-gei/pull/872

PR #848, opened 2023-03-02
Update SecretScanningAlertService.cs

kyle-jones:
Adding additional definitions of matched for the secrets, if the target or the source contains the secret string within the other then this is also considered a match.

- [ ] Issue linked


kyle-jones:
Support Ticket 2038460

dylan-smith:
Thanks for the contribution! We'll try to have somebody look into this next week (I would but I'm on vacation!). We'll want to try and repro the issue so we can test the fix.

timrogers:
FYI @peter-murray - I know you built this feature originally, so you might have some useful context.

peter-murray:
It was written to ensure an exact match and avoid false matches as you would not want to remediate a secret that has not actually been remediated in the migrated data.

All testing for the feature was performed on a GHEC to GHEC EMU process, so this potential discrepancy was not present in the testing that I performed when I wrote it.

My only concern would be in avoiding any potential false positive matches, as it is better to not link a remediation over getting a remediation incorrectly applied to a valid open alert.

Looking at the context of the support ticket, there might be other permutations between GHES and GHEC where the alerts from the APIs differ, which probably requires a deeper investigation so that this fix can be made to accommodate.

kyle-jones:
I have 1771 repositories that I have migrated from GHES to GHEC EMU to test the solution with. I would be happy to test the solution for you. Currently false negative matches abound. We activated secret scanning over a year ago on the GHES platform so we have a volume of alerts we a dealing with.

src/Octoshift/SecretScanningAlertService.cs
```diff
@@ -89,7 +89,11 @@ private AlertWithLocations MatchTargetSecret(AlertWithLocations source, List<Ale
             }
 
             if (source.Alert.SecretType == target.Alert.SecretType
-                && source.Alert.Secret == target.Alert.Secret)
+                && ((source.Alert.Secret == target.Alert.Secret) ||
+                    (source.Alert.Secret.Length > target.Alert.Secret.Length && source.Alert.Secret.Contains(target.Alert.Secret)) ||
+                    (source.Alert.Secret.Length < target.Alert.Secret.Length && target.Alert.Secret.Contains(source.Alert.Secret))
+                   )
+                )
```
ArinGhazarian, line 96: While I am not sure if this is correct since I am not super familiar with secret scanning stuff, but in terms of pure code review, the condition can be simplified:
```suggestion
                && (source.Alert.Secret.Contains(target.Alert.Secret) || target.Alert.Secret.Contains(source.Alert.Secret)))
```

A couple of notes:
1. `string.Contains()` is super optimized so no need to do the length check before calling it at all.
2. A == B basically means A.Contains(B) && B.Contains(A) but since we're doing A.Contains(B) || B.Contains(A) later then we can just remove the redundant equality check.

ArinGhazarian (changes requested):
Hi @kyle-jones 👋,

While I am not confident that the change is correct, I dropped a comment regarding the code itself. Also assuming that the change is good at least two unit tests needs to be added before we can approve this work.

kyle-jones:
@ArinGhazarian I am a newbie with this process, is there anything else I need to do for this change to be considered?

mbadgett:
@kyle-jones 
Hi, I'm from the Secret Scanning team. I don't have any immediate concerns with the code changes but I am also not certain of what type of gaps this change is trying to close. Do you have any examples of alerts that would match with the before code logic that will match with the change you have made? If so I would like to see some examples of the secrets and Alert types that would be affected by this change.

This is the kind of thing where unit tests in the code would definitely help for a person like me who is not familiar with all of the inner workings of this repo.

dylan-smith:
> @kyle-jones Hi, I'm from the Secret Scanning team. I don't have any immediate concerns with the code changes but I am also not certain of what type of gaps this change is trying to close. Do you have any examples of alerts that would match with the before code logic that will match with the change you have made? If so I would like to see some examples of the secrets and Alert types that would be affected by this change.
> 
> This is the kind of thing where unit tests in the code would definitely help for a person like me who is not familiar with all of the inner workings of this repo.

@mbadgett There's some more info in this issue: https://github.com/github/secret-scanning/issues/4493

Although my understanding is we haven't been able to repro it yet. Certainly there would need to be unit tests before we'd consider merging this change.

timrogers:
@kyle-jones Based on @mbadgett's feedback above, I think this logic is probably "sound" from a secret scanning perspective. Would you be able to (a) add unit tests and (b) provide an example of the case where this change is needed?

mbadgett:
Just confirming, I have looked into the actual pattern that was found for this customer. At some point between GHES 3.6 and now we changed the Azure account storage key pattern to include the `AccountName=[0-9a-z]*;AccountKey=` as a delimiter for the secret whereas before, it was part of the secret. This is an appropriate fix with the caveat that adding an if statement targeting the specific alert type would be the safest thing to do, but it would only be an added layer of redundancy as far as I'm concerned given that this code already checks for secret location info matching.

mbadgett (approved):
Representing the Secret Scanning Team

timrogers:
@mbadgett Thanks! 🥳 

@kyle-jones Thanks again for your contribution here. We have confirmation from the Secret Scanning team that this makes sense to them, and we would be happy to merge and release once we have unit tests.

dylan-smith:
I was just working on tossing a few unit tests to go along with this change so we could merge it in, but I discovered a problem. I don't think this change as-is is enough to fix the problem. Because the current code also does matching based on the location, which includes looking for a matching: StartLine/EndLine/StartColumn/EndColumn.

And it's not immediately obvious to me the best way to solve for this since the columns and possibly lines (for multi-line secrets) will differ in the scenario this PR is intended to fix.

If somebody wanted to take a crack at that logic, I'd be happy to look at it again later and help with unit tests if necessary.

dylan-smith:
Closing this until we can come up with a solution to the problem mentioned above.

PR #855, opened 2023-03-03
Rename AWS env variables to align with AWS CLI

ArinGhazarian:
Related to #738 

- [x] Issue linked

### Description

AWS CLI uses slightly [different environment variable names](https://docs.aws.amazon.com/sdk-for-net/v3/developer-guide/creds-assign.html) than we use. In order to avoid confusion and be able to reuse already existing AWS CLI variables, we need to also rename our env. variables to align with AWS CLI ones. This PR only changes the environment variable names not the actual arg names for backward compatibility. We will tackle the actual arg name changes later when timing is right in a separate PR.
The old environment variables are still supported but we spit out a warning to notify the users that they will be removed in future. 

Note:
- Once this PR is merged, the old action secrets for INT tests should be removed.


timrogers (approved):
Nice work! 🥳

synthead (approved):
Looks great!  Love the backward compatibility! :ship:

Issue #867, opened 2023-03-09
Receiving a GraphQL error from `createMigrationSource` mutation triggers a `System.InvalidOperationException` with unhelpful error message


timrogers:
## Description

When running `gh gei migrate-repo`, if the GraphQL `createMigrationSource` mutation responds with a `200 OK` with a GraphQL error, e.g.:

```json
{"data":{"createMigrationSource":null},"errors":[{"type":"FORBIDDEN","path":["createMigrationSource"],"extensions":{"saml_failure":true},"locations":[{"line":1,"column":109}],"message":"Resource protected by organization SAML enforcement. You must grant your Personal Access token access to this organization."}]}
```

...the CLI throws a `System.InvalidOperationException: Cannot access child value on Newtonsoft.Json.Linq.JValue.` and exits.

The underlying error message (`Resource protected by organization SAML enforcement. You must grant your Personal Access token access to this organization.`) is not visible unless they either (a) use the `--verbose` option or (b) manually check the verbose log file.

We should handle this case - and indeed other cases where the GraphQL API returns an error - by throwing an exception with the relevant error message.

## Reproduction Steps

Call the `gh gei migrate-repo` command with a target organization that requires the personal access token (PAT) to be enabled for SSO, but where the token hasn't been authorized for SSO.

tgnanakumar:
We integrated [GitHub App Token](https://github.com/marketplace/actions/github-app-token) as part of our GitHub Action workflow automated using the IssueOps way (instead of using individual user personal access token) for migrating repositories from Azure DevOps to GitHub. However, we're facing a similar FORBIDDEN error as @timrogers but with a different message **Resource not accessible by integration** in the log. Are both of these the same issue or different? When I explicitly use a personal access token instead that is enabled and authorized for SSO, it works fine though. What are we missing here?

Here is the error printed in the log,
```
{
  "data": {
    "createMigrationSource": null
  },
  "errors": [
    {
      "type": "FORBIDDEN",
      "path": [
        "createMigrationSource"
      ],
      "extensions": {
        "saml_failure": false
      },
      "locations": [
        {
          "line": 1,
          "column": 109
        }
      ],
      "message": "Resource not accessible by integration"
    }
  ],
  "extensions": {
    "warnings": [
      {
        "type": "DEPRECATION",
        "message": "The id xxxxxxxxxxxxxx is deprecated. Update your cache to use the next_global_id from the data payload.",
        "data": {
[...]

PR #868, opened 2023-03-09
Consistently convert GraphQL errors to `OctoshiftCliException`s with helpful error messages across `GithubApi`

timrogers:
In `GithubApi`, we have the helpful `EnsureSuccessGraphQLResponse` helper which checks the GraphQL response for errors, and if an error is found, it raises a `OctoshiftCliException`.

Currently, this helper is not used consistently across GraphQL API calls, which means that we have some cases where the GraphQL API returns an error and the CLI crashes with an unhelpful `System.InvalidOperationException: Cannot access child value on Newtonsoft.Json.Linq.JValue.`.

To find out what actually went wrong, the user has to either:

* (a) use the `--verbose` option or
* (b) manually check the verbose log file

This PR creates a new `PostGraphQLAsync` method on `GitHubClient` which includes the error handling logic from `EnsureSuccessGraphQLResponse`.

We then use the `PostGraphQLAsync` method instead of `PostAsync` for GraphQL requests, which means that we get this consistent error handling for free.

I do not apply this to `ReclaimMannequin` (which has its own error handling) and `GetMannequins` (which has some complex pagination logic which I didn't want to mess with right now).

Fixes https://github.com/github/gh-gei/issues/867.

- [x] Issue linked

### DIFF
--- a/src/Octoshift/GithubApi.cs
+++ b/src/Octoshift/GithubApi.cs
@@ -167,10 +167,7 @@ namespace OctoshiftCLI
 
             var response = await _retryPolicy.Retry(async () =>
             {
-                var httpResponse = await _client.PostAsync(url, payload);
-                var data = JObject.Parse(httpResponse);
-
-                EnsureSuccessGraphQLResponse(data);
+                var data = await _client.PostGraphQLAsync(url, payload);
 
                 return (string)data["data"]["organization"]["id"];
             });
@@ -192,10 +189,7 @@ namespace OctoshiftCLI
 
             var response = await _retryPolicy.Retry(async () =>
             {
-                var httpResponse = await _client.PostAsync(url, payload);
-                var data = JObject.Parse(httpResponse);
-
-                EnsureSuccessGraphQLResponse(data);
+                var data = await _client.PostGraphQLAsync(url, payload);
 
                 return (string)data["data"]["enterprise"]["id"];
             });
@@ -227,10 +221,7 @@ namespace OctoshiftCLI
                 operationName = "createMigrationSource"
             };
 
-            var response = await _client.PostAsync(url, payload);
-            var data = JObject.Parse(response);
-
-            EnsureSuccessGraphQLResponse(data);
+            var data = await _client.PostGraphQLAsync(url, payload);
 
             return (string)data["data"]["createMigrationSource"]["migrationSource"]["id"];
         }
@@ -255,10 +246,7 @@ namespace OctoshiftCLI
                 operationName = "createMigrationSource"
             };
 
-            var response = await _client.PostAsync(url, payload);
-            var data = JObject.Parse(response);
-
-            EnsureSuccessGraphQLResponse(data);
+            var data = await _client.PostGraphQLAsync(url, payload);
 
             return (string)data["data"]["createMigrationSource"]["migrationSource"]["id"];
         }
@@ -283,10 +271,7 @@ namespace OctoshiftCLI
                 operationName = "createMigrationSource"
             };
 
-            var response = await _client.PostAsync(url, payload);
-            var data = JObject.Parse(response);
-
-            EnsureSuccessGraphQLResponse(data);
+            var data = await _client.PostGraphQLAsync(url, payload);
 
             return (string)data["data"]["createMigrationSource"]["migrationSource"]["id"];
         }
@@ -357,10 +342,7 @@ namespace OctoshiftCLI
                 operationName = "startRepositoryMigration"
             };
 
-            var response = await _client.PostAsync(url, payload);
-            var data = JObject.Parse(response);
-
-            EnsureSuccessGraphQLResponse(data);
+            var data = await _client.PostGraphQLAsync(url, payload);
 
             return (string)data["data"]["startRepositoryMigration"]["repositoryMigration"]["id"];
         }
@@ -401,10 +383,7 @@ namespace OctoshiftCLI
                 operationName = "startOrganizationMigration"
             };
 
-            var response = await _client.PostAsync(url, payload);
-            var data = JObject.Parse(response);
-
-            EnsureSuccessGraphQLResponse(data);
+            var data = await _client.PostGraphQLAsync(url, payload);
 
             return (string)data["data"]["startOrganizationMigration"]["orgMigration"]["id"];
         }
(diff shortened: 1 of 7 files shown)


src/Octoshift/ReclaimService.cs
```diff
@@ -99,10 +99,6 @@ public virtual async Task ReclaimMannequin(string mannequinUser, string mannequi
             }
 
             var targetUserId = await _githubApi.GetUserId(targetUser);
-            if (targetUserId == null)
```
timrogers, line 102: This now throws an error instead

src/OctoshiftCLI.Tests/GithubApiTests.cs
```diff
@@ -1832,45 +2030,41 @@ public async Task GetUserId_Returns_The_User_Id()
         }
 
         [Fact]
-        public async Task GetUserId_For_No_Existant_User_Returns_Null()
```
timrogers, line 1835: We now just cover the generic "there is an error" case rather than the "user not found" case - although the test proves how we handle a missing user in particular

ArinGhazarian (changes requested):
A more DRY way to handle this is to add a method just for posting GQL requests, like `PostGraphQLAsync()`  in `GithubClient`, and then move the ensure success logic there. We then need to update all GQL api methods to call `PostGraphQLAsync` instead of `PostAsync` and of course add/update tests.

timrogers:
> A more DRY way to handle this is to add a method just for posting GQL requests, like `PostGraphQLAsync()` in `GithubClient`, and then move the ensure success logic there. We then need to update all GQL api methods to call `PostGraphQLAsync` instead of `PostAsync` and of course add/update tests.

That's a good point. Happy to do that.

src/Octoshift/GithubApi.cs
```diff
@@ -696,8 +670,7 @@ public virtual async Task<string> GetUserId(string login)
             };
 
             // TODO: Add retry logic here, but need to inspect the actual error message and differentiate between transient failure vs user doesn't exist (only retry on failure)
-            var response = await _client.PostAsync(url, payload);
-            var data = JObject.Parse(response);
+            var data = await _client.PostGraphQLAsync(url, payload);
 
             return data["data"]["user"].Any() ? (string)data["data"]["user"]["id"] : null;
```
timrogers: A missing user triggers a GraphQL error, which now causes `PostGraphQLAsync` to throw, so this ternary is no longer needed. If we get to this point, there is a user.

timrogers:

src/Octoshift/GithubClient.cs
```diff
@@ -73,6 +73,27 @@ public virtual async IAsyncEnumerable<JToken> GetAllAsync(string url, Dictionary
...
+
+            if (response.TryGetValue("errors", out var jErrors) && jErrors is JArray { Count: > 0 } errors)
+            {
+                var error = (JObject)errors[0];
+                var errorMessage = error.TryGetValue("message", out var jMessage) ? (string)jMessage : null;
+                throw new OctoshiftCliException($"{errorMessage ?? "UNKNOWN"}");
+            }
+
+            return response;
+        }
```
ArinGhazarian, line 95: This method basically needs to call the `PostAsync` and then additionally ensures a success GQL response. So no need to re-set the `variables` or do anything extra (it does so in the paginated version for a reason), it should be as simple as passing in an object body and then re-passing it to `PostAsync`. I would also recommend keeping the error checking logic in its separate method since it also needs to be called in the `PostGraphQLWithPaginationAsync()` method. With all these your method should look like this (Note that the method return a `JToken` which is more generic that `JObject`):
```suggestion
        public virtual async Task<JToken> PostGraphQLAsync(
            string url,
            object body,
            Dictionary<string, string> customHeaders = null)
        {
            var response = await PostAsync(url, body, customHeaders);
            var data = JObject.Parse(response);

            EnsureSuccessGraphQLResponse(data);
            
            return data;
        }
```

---
### The following can also be tackled in a separate PR

Also the `PostGraphQLWithPaginationAsync()` needs to be refactored with the above changes, just two lines of changes (use `PostAsync` instead of `SendAsync` and call `EnsureSuccessGraphQLResponse()`):
```c#
public virtual async IAsyncEnumerable<JToken> PostGraphQLWithPaginationAsync(...)
{
    ...

    var response = await PostAsync(url, body, customHeaders);
    var jContent = JObject.Parse(content);
    EnsureSuccessGraphQLResponse(jContent);

    ...
}
```
of course a test also needs to be added for `PostGraphQLWithPaginationAsync()` to make sure that it throws in case of a GQL error.

ArinGhazarian (changes requested):

src/Octoshift/GithubClient.cs
timrogers: Thanks for the feedback! Totally makes sense - and interesting to know that JToken is more generic, I didn't know that 😊

I'll update this tomorrow - and can look at the paginated method in a separate PR.

src/OctoshiftCLI.Tests/GithubClientTests.cs
```diff
@@ -619,6 +629,138 @@ await FluentActions
...
+        {
+            // Arrange
+            using var httpClient = new HttpClient(MockHttpHandlerForGraphQLPost().Object);
+            var githubClient = new GithubClient(_mockOctoLogger.Object, httpClient, null, _retryPolicy, _dateTimeProvider.Object, PERSONAL_ACCESS_TOKEN);
+
+            // Act
+            var response = await githubClient.PostGraphQLAsync("http://example.com", _rawRequestBody);
+
+            // Assert
+            Assert.True(JToken.DeepEquals(response, JObject.Parse(EXPECTED_GRAPHQL_JSON_RESPONSE_BODY)));
```
ArinGhazarian, line 643: We use fluent assertions, so either use `Should().BeTrue()` or IMO even better the following:
```suggestion
            response.ToJson().Should().Be(EXPECTED_JSON_REQUEST_BODY);
```

ArinGhazarian (commented):
Looks good just left a small comment also would you create an issue to refactor the paginated GQL call to also throw GQL errors.

src/OctoshiftCLI.Tests/GithubClientTests.cs
timrogers: Thanks for flagigng! Let me try this. I was trying to use an assertion like that but it was failing :(

timrogers: That works perfectly! Let me add it.

timrogers:
> Looks good just left a small comment also would you create an issue to refactor the paginated GQL call to also throw GQL errors.

Follow up issue created: https://github.com/github/gh-gei/issues/883

timrogers:
@ArinGhazarian Ready for review!

ArinGhazarian (approved):
approved_1

PR #872, opened 2023-03-10
Validate AWS region

### LINKED ISSUE #871: Validate AWS Region in AwsApi
## Description

If the AWS region is provided by passing the `--aws-region` option or setting the `AWS_REGION` endpoint, we try to get an AWS `RegionEndpoint` object by calling `RegionEndpoint.GetBySystemName()`. Unlike many familiar APIs, this method doesn't validate  and throw if the region is invalid! The AWS S3 upload fails in case of an invalid region but we want to fail fast and early for a better user experience. 

## Todo

- [x] Validation the provided AWS region in AwsApi.

ArinGhazarian:
Closes #871 

- [x] Issue linked

### Description
This PR adds validation for AWS region in order to fail early.

### DIFF
--- a/src/Octoshift/AwsApi.cs
+++ b/src/Octoshift/AwsApi.cs
@@ -30,13 +30,22 @@ public class AwsApi : IDisposable
         var regionEndpoint = DefaultRegionEndpoint;
         if (awsRegion.HasValue())
         {
-            regionEndpoint = RegionEndpoint.GetBySystemName(awsRegion);
+            regionEndpoint = GetRegionEndpoint(awsRegion);
             AWSConfigsS3.UseSignatureVersion4 = true;
         }
 
         return awsSessionToken.HasValue()
             ? new AmazonS3Client(awsAccessKeyId, awsSecretAccessKey, awsSessionToken, regionEndpoint)
             : new AmazonS3Client(awsAccessKeyId, awsSecretAccessKey, regionEndpoint);
+    }
+
+    private static RegionEndpoint GetRegionEndpoint(string awsRegion)
+    {
+        var regionEndpoint = RegionEndpoint.GetBySystemName(awsRegion);
+
+        return regionEndpoint.DisplayName.ToLower() == "unknown"
+            ? throw new OctoshiftCliException($"Invalid AWS region \"{awsRegion}\".")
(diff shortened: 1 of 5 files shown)


src/Octoshift/AwsApi.cs
```diff
@@ -39,6 +39,10 @@ private static AmazonS3Client BuildAmazonS3Client(string awsAccessKeyId, string
             : new AmazonS3Client(awsAccessKeyId, awsSecretAccessKey, regionEndpoint);
     }
 
+    private static RegionEndpoint GetRegionEndpoint(string awsRegion) => RegionEndpoint.GetBySystemName(awsRegion) is { DisplayName: not "Unknown" } regionEndpoint
+        ? regionEndpoint
+        : throw new OctoshiftCliException($"Invalid AWS region \"{awsRegion}\".");
```
timrogers, line 44: I guess that this change means that we will have to update `gh-gei` and bump the version of the AWS SDK every time AWS launch a new region. Otherwise, that region won't be known and we will throw an exception. 

I wonder if it's a good idea to couple the CLI so tightly to the SDK. I wonder if it would be a better option to log a warning for an unrecognised region instead. What do you think?

ArinGhazarian: This may be a valid point but I am not sure what to do. The AWS library is calling the AWS web api under the hood so probably if we use an older library, passing in a new region shouldn't break the api call but I wonder how frequently it happens. Also there are two scenarios involved here:
1. the region is totally invalid like `blahbalh`.
2. the region is valid but not present in the AWS library version we use. 
This validation is trying to prevent 1, but also in case 2 even if we spit out a warning it will be a false negative and still be somewhat confusing. So I am wondering that in any case whether we error or log a warning, we'll be probably forced to upgrade the AWS library. 
If we don't validate (AWS also doesn't validate) the upload phase is going to fail eventually but it's not gonna be the best UX probably.
What do you think @timrogers?

timrogers: I think we should stick with this now, and we can see whether we get any issues. Overall, I do think a stricter validation this is probably better.

timrogers (approved):

Issue #873, opened 2023-03-13
Allow the host for SSH connections to be configured in `gh bbs2gh`


timrogers:
`gh bbs2gh` can automatically connect to a Bitbucket Server instance using SSH to download the migration archive generated using the Bitbucket Server REST API.

Currently, our code assumes that the same hostname should be used for the SSH connection as the HTTP(S) connection to the REST API. This may not be the case - for example, the Bitbucket Server's UI and REST API may be exposed through a reverse proxy.

We should add an explicit `--ssh-host` option to allow the SSH host to be configured directly - though we would continue to default to the current behaviour of using the host from the `--bbs-server-url`.

This issue was reported by a customer who was unable to use the CLI due to this limitation.

## Todo

- [ ] ...

## Dependencies

- ...

bzanin-wdc:
I would love for this option to be unhidden.  Would have saved us a few hours of unnecessary working-around if we had been able to find the option!  (Delighted to learn about it now, though.)

timrogers:
See https://github.com/github/gh-gei/pull/874#issuecomment-1536200359 for my review to the above comment 👼🏻

PR #874, opened 2023-03-13
Add support for specifying custom host for SSH/SMB downloads with hidden `--archive-download-host` argument in `gh bbs2gh`

timrogers:
`gh bbs2gh` can automatically connect to a Bitbucket Server instance using SSH or SMB to download the migration archive generated using the Bitbucket Server REST API.

Currently, our code assumes that the same hostname should be used for the SSH/SMB connection as the HTTP(S) connection to the REST API. This may not be the case - for example, the Bitbucket Server's UI and REST API may be exposed through a reverse proxy.

This adds an explicit `--archive-download-host` option to allow the host for the SSH/SMB connection to be configured directly - though we will continue to default to the current behaviour of using the host from the `--bbs-server-url`. 

This argument is hidden from the help docs for now, and we can unhide it in future if it turns out that lots of customers ask about it, rather than just one.

This issue was reported by a customer who was unable to use the CLI due to this limitation.

Fixes https://github.com/github/gh-gei/issues/873.

- [x] Issue linked

### DIFF
--- a/src/OctoshiftCLI.Tests/bbs2gh/Commands/MigrateRepoCommandTests.cs
+++ b/src/OctoshiftCLI.Tests/bbs2gh/Commands/MigrateRepoCommandTests.cs
@@ -14,7 +14,7 @@ public class MigrateRepoCommandTests
     private const string SSH_PRIVATE_KEY = "ssh-private-key";
     private const int SSH_PORT = 1234;
     private const string BBS_SHARED_HOME = "shared-home";
-    private const string BBS_HOST = "bbs-server-url-host";
+    private const string BBS_HOST = "bbs-host";
     private const string BBS_SERVER_URL = $"https://{BBS_HOST}";
     private const string GITHUB_ORG = "github-org";
     private const string GITHUB_PAT = "github-pat";
--- a/src/OctoshiftCLI.Tests/bbs2gh/Commands/MigrateRepoCommandTests.cs
+++ b/src/OctoshiftCLI.Tests/bbs2gh/Commands/MigrateRepoCommandTests.cs
@@ -9,12 +9,12 @@ namespace OctoshiftCLI.Tests.BbsToGithub.Commands;
 
 public class MigrateRepoCommandTests
 {
+    private const string SSH_HOST = "ssh-host";
     private const string SSH_USER = "ssh-user";
     private const string SSH_PRIVATE_KEY = "ssh-private-key";
     private const int SSH_PORT = 1234;
     private const string BBS_SHARED_HOME = "shared-home";
-    private const string BBS_HOST = "bbs-host";
-    private const string BBS_SERVER_URL = $"https://{BBS_HOST}";
+    private const string BBS_SERVER_URL = $"https://bbs-server-url-host";
     private const string GITHUB_ORG = "github-org";
     private const string GITHUB_PAT = "github-pat";
     private const string BBS_USERNAME = "bbs-username";
@@ -52,7 +52,7 @@ public class MigrateRepoCommandTests
         var command = new MigrateRepoCommand();
         command.Should().NotBeNull();
         command.Name.Should().Be("migrate-repo");
-        command.Options.Count.Should().Be(28);
+        command.Options.Count.Should().Be(29);
 
         TestHelpers.VerifyCommandOption(command.Options, "bbs-server-url", false);
         TestHelpers.VerifyCommandOption(command.Options, "bbs-project", false);
@@ -70,6 +70,7 @@ public class MigrateRepoCommandTests
         TestHelpers.VerifyCommandOption(command.Options, "github-org", false);
         TestHelpers.VerifyCommandOption(command.Options, "github-repo", false);
         TestHelpers.VerifyCommandOption(command.Options, "github-pat", false);
+        TestHelpers.VerifyCommandOption(command.Options, "ssh-host", false);
         TestHelpers.VerifyCommandOption(command.Options, "ssh-user", false);
         TestHelpers.VerifyCommandOption(command.Options, "ssh-private-key", false);
         TestHelpers.VerifyCommandOption(command.Options, "ssh-port", false);
@@ -84,7 +85,7 @@ public class MigrateRepoCommandTests
     }
 
     [Fact]
-    public void BuildHandler_Creates_Bbs_Ssh_Archive_Downloader_When_Ssh_User_Is_Provided()
+    public void BuildHandler_Creates_Bbs_Ssh_Archive_Downloader_Based_On_Server_Url_When_Ssh_User_Is_Provided()
     {
         // Arrange
         var args = new MigrateRepoCommandArgs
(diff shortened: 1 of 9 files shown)


src/bbs2gh/Commands/MigrateRepoCommand.cs
```diff
@@ -208,10 +213,11 @@ public override MigrateRepoCommandHandler BuildHandler(MigrateRepoCommandArgs ar
...
         {
             var bbsArchiveDownloaderFactory = sp.GetRequiredService<BbsArchiveDownloaderFactory>();
-            var bbsHost = new Uri(args.BbsServerUrl).Host;
+            var sshHost = args.SshHost.HasValue() ? args.SshHost : new Uri(args.BbsServerUrl).Host;
+
             bbsArchiveDownloader = args.SshUser.HasValue()
-                ? bbsArchiveDownloaderFactory.CreateSshDownloader(bbsHost, args.SshUser, args.SshPrivateKey, args.SshPort, args.BbsSharedHome)
-                : bbsArchiveDownloaderFactory.CreateSmbDownloader(bbsHost, args.SmbUser, args.SmbPassword, args.SmbDomain, args.BbsSharedHome);
+                ? bbsArchiveDownloaderFactory.CreateSshDownloader(sshHost, args.SshUser, args.SshPrivateKey, args.SshPort, args.BbsSharedHome)
+                : bbsArchiveDownloaderFactory.CreateSmbDownloader(sshHost, args.SmbUser, args.SmbPassword, args.SmbDomain, args.BbsSharedHome);
```
ArinGhazarian, line 220: This is going to be very confusing since `ssh-host` is being used both for SMB and SSH. Either we need to drop the ssh to indicate it's for both SMB and SSM or create yet another one for SMB (e.g. `--smb-host`).

src/bbs2gh/Handlers/MigrateRepoCommandHandler.cs
```diff
@@ -136,7 +136,7 @@ private bool ShouldGenerateArchive(MigrateRepoCommandArgs args)
 
     private bool ShouldDownloadArchive(MigrateRepoCommandArgs args)
     {
-        return args.SshUser.HasValue() || args.SmbUser.HasValue();
+        return args.SshUser.HasValue() || args.SmbUser.HasValue() || args.SshHost.HasValue();
```
ArinGhazarian, line 139: No need to add the ssh host here, our indicators can still be the ssh user and smb user.

ArinGhazarian (changes requested):
My main problem is that we reused the `--ssh-host` both for SMB and SSH which is very confusing.

src/bbs2gh/Commands/MigrateRepoCommand.cs
timrogers: Oops! I didn't mean to do that. Would it be okay with you if I just add the `--ssh-user` option for now, which only works for SSH, and then we can add a similar feature for SMB later?

src/OctoshiftCLI.Tests/bbs2gh/Commands/MigrateRepoCommandTests.cs
```diff
@@ -9,12 +9,12 @@ namespace OctoshiftCLI.Tests.BbsToGithub.Commands;
...
 public class MigrateRepoCommandTests
 {
+    private const string SSH_HOST = "ssh-host";
     private const string SSH_USER = "ssh-user";
     private const string SSH_PRIVATE_KEY = "ssh-private-key";
     private const int SSH_PORT = 1234;
     private const string BBS_SHARED_HOME = "shared-home";
-    private const string BBS_HOST = "bbs-host";
-    private const string BBS_SERVER_URL = $"https://{BBS_HOST}";
+    private const string BBS_SERVER_URL = $"https://bbs-server-url-host";
```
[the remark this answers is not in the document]
timrogers: How about if I rename it to `BBS_API_HOST`, just so it's meaning is clear?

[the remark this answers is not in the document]
ArinGhazarian: What's wrong with just calling it what it is? it is technically the BBS [host](https://developer.mozilla.org/en-US/docs/Web/API/URL/host) part of the url. it's not the api host or anything else.

src/bbs2gh/Commands/MigrateRepoCommand.cs
ArinGhazarian: is it worth creating two args? can't we just create one arg that can be reused for both? here are a couple of examples:
- `--bbs-host`
- `--download-host`

src/OctoshiftCLI.Tests/bbs2gh/Commands/MigrateRepoCommandTests.cs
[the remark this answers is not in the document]
timrogers: I just thought it was a helpful name to make it clear that there are different hosts used for different purposes. 

Happy to leave it if you don't think it's confusing 😊

src/bbs2gh/Commands/MigrateRepoCommand.cs
timrogers: I agree with you that it would probably be nice to have a single argument that covers SMB and SSH - but I want to think about that a bit more and consider what name I like best. 

I'll come back to you tomorrow!

src/OctoshiftCLI.Tests/bbs2gh/Commands/MigrateRepoCommandTests.cs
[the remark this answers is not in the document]
ArinGhazarian: I don't think it's gonna be confusing especially if we use the `--bbs-host` as the arg.

src/bbs2gh/Commands/MigrateRepoCommand.cs
ArinGhazarian: it sounds great 🎉

timrogers:

dylan-smith:
What do you think about making this a hidden arg? I'm thinking this will be an extremely niche scenario (but I may be wrong).

timrogers:
> What do you think about making this a hidden arg? I'm thinking this will be an extremely niche scenario (but I may be wrong).

I'm up for that - and we can always unhide it if it turns out that lots of customers ask about it.

timrogers:
@dylan-smith Done!

src/bbs2gh/Handlers/MigrateRepoCommandHandler.cs
```diff
@@ -471,7 +476,7 @@ private void ValidateOptions(MigrateRepoCommandArgs args)
 
     private void ValidateDownloadOptions(MigrateRepoCommandArgs args)
     {
-        var sshArgs = new[] { args.SshUser, args.SshPrivateKey };
+        var sshArgs = new[] { args.ArchiveDownloadHost, args.SshUser, args.SshPrivateKey };
```
ArinGhazarian, line 479: Since this is an optional arg that can belong to both ssh and smb I would just remove it from ssh args. Instead one extra validation can be added:
```c#
if (args.ArchiveDownloadHost.HasValue() && !shouldUseShh && !shouldUseSmb)
{
    throw new OctoshiftCliException("--archive-download-host should be provided either with SSH or SMB download options.");
}
```

ArinGhazarian (changes requested):

timrogers:
@ArinGhazarian Thanks for the review! I'll address this tomorrow.

src/bbs2gh/Handlers/MigrateRepoCommandHandler.cs
timrogers: Good point! I wrote this before we decided to use the same argument for SSH and SMB.

src/bbs2gh/Handlers/MigrateRepoCommandHandler.cs
```diff
@@ -437,10 +442,7 @@ private void ValidateOptions(MigrateRepoCommandArgs args)
                 }
             }
 
-            if (ShouldDownloadArchive(args))
-            {
-                ValidateDownloadOptions(args);
-            }
+            ValidateDownloadOptions(args);
```
timrogers, line 445: We can actually run this validation method even if we're not downloading - the method is safe to run. This allows us to include the new `--archive-download-host can only be provided if SSH or SMB download options are provided.` validation inside the method too.

ArinGhazarian (approved):
🎉

bzanin-wdc:
For whatever it's worth, we had to deal with this specific issue too.  If this option had not been hidden it would have saved us a good two or three hours of troubleshooting during a pretty intense working session!

timrogers:
> For whatever it's worth, we had to deal with this specific issue too. If this option had not been hidden it would have saved us a good two or three hours of troubleshooting during a pretty intense working session!

@bzanin-wdc Thanks for taking the time to comment here. Your comment is what I needed to persuade me to make this option public - see #972 for the change that's now in review. Sorry for the pain you had tracking this option down!

Issue #831, opened 2023-02-21
Add `--keep-archive` option for GHES migrations to keep the archive on the user's machine


timrogers:
When running a migration from GitHub Enterprise Server, we use the GHES API to generate an archive and then we upload that archive to the user's blob storage provider.

The API is not stored on the user's disk - which is great! - but it would be helpful to have an option to optionally keep the archive for the case where the customer wants to share it with us for debugging reasons.

We should add an optional `--keep-archive` option for GHES migrations which leaves the archive stored on user's machine. By default, we should continue to not store the archive.

spindev:
@timrogers See #875 please.

PR #875, opened 2023-03-14
Added --keep-archive for GHES migrations

spindev:
Solved #831

- [x] Issue linked

### DIFF
--- a/src/OctoshiftCLI.Tests/gei/Handlers/MigrateRepoCommandHandlerTests.cs
+++ b/src/OctoshiftCLI.Tests/gei/Handlers/MigrateRepoCommandHandlerTests.cs
@@ -1724,6 +1724,22 @@ namespace OctoshiftCLI.Tests.GithubEnterpriseImporter.Commands
         }
 
         [Fact]
+        public async Task Keep_Archive_Without_Ghes_Api_Url_Throws()
+        {
+            await _handler.Invoking(async x => await x.Handle(new MigrateRepoCommandArgs
+            {
+                SourceRepo = SOURCE_REPO,
+                GithubSourceOrg = SOURCE_ORG,
+                GithubTargetOrg = TARGET_ORG,
+                TargetRepo = TARGET_REPO,
+                KeepArchive = true
+            }))
+                .Should()
+                .ThrowAsync<OctoshiftCliException>()
+                .WithMessage("*--keep-archive*");
+        }
+
+        [Fact]
         public async Task Keep_Archive_Does_Not_Call_DeleteIfExists()
         {
             var githubOrgId = Guid.NewGuid().ToString();
--- a/src/OctoshiftCLI.Tests/gei/Handlers/MigrateRepoCommandHandlerTests.cs
+++ b/src/OctoshiftCLI.Tests/gei/Handlers/MigrateRepoCommandHandlerTests.cs
@@ -1722,5 +1722,80 @@ namespace OctoshiftCLI.Tests.GithubEnterpriseImporter.Commands
                 .ThrowAsync<OctoshiftCliException>()
                 .WithMessage("*--no-ssl-verify*");
         }
+
+        [Fact]
+        public async Task Keep_Archive_Does_Not_Call_DeleteIfExists()
+        {
+            var githubOrgId = Guid.NewGuid().ToString();
+            var migrationSourceId = Guid.NewGuid().ToString();
+            var sourceGithubPat = Guid.NewGuid().ToString();
+            var targetGithubPat = Guid.NewGuid().ToString();
+            var githubRepoUrl = $"https://myghes/{SOURCE_ORG}/{SOURCE_REPO}";
+            var migrationId = Guid.NewGuid().ToString();
+            var gitArchiveId = 1;
+            var metadataArchiveId = 2;
+            var gitArchiveFilePath = "path/to/git_archive";
+            var metadataArchiveFilePath = "path/to/metadata_archive";
+
+            var gitArchiveUrl = $"https://example.com/{gitArchiveId}";
+            var metadataArchiveUrl = $"https://example.com/{metadataArchiveId}";
+            var authenticatedGitArchiveUrl = new Uri($"https://example.com/{gitArchiveId}/authenticated");
+            var authenticatedMetadataArchiveUrl = new Uri($"https://example.com/{metadataArchiveId}/authenticated");
+
+            _mockSourceGithubApi.Setup(x => x.GetEnterpriseServerVersion()).ReturnsAsync("GitHub AE");
+            _mockTargetGithubApi.Setup(x => x.GetOrganizationId(TARGET_ORG).Result).Returns(githubOrgId);
+            _mockTargetGithubApi.Setup(x => x.CreateGhecMigrationSource(githubOrgId).Result).Returns(migrationSourceId);
+            _mockTargetGithubApi
+                .Setup(x => x.StartMigration(
+                    migrationSourceId,
+                    githubRepoUrl,
+                    githubOrgId,
+                    TARGET_REPO,
+                    sourceGithubPat,
+                    targetGithubPat,
+                    authenticatedGitArchiveUrl.ToString(),
+                    authenticatedMetadataArchiveUrl.ToString(),
+                    false,
+                    false).Result)
+                .Returns(migrationId);
+            _mockTargetGithubApi.Setup(x => x.GetMigration(migrationId).Result).Returns((State: RepositoryMigrationStatus.Succeeded, TARGET_REPO, null));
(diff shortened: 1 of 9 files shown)


spindev:
#831

timrogers:
@spindev Thanks for your contribution! We'll review this and get back to you.

src/OctoshiftCLI.Tests/gei/Handlers/MigrateRepoCommandHandlerTests.cs
```diff
@@ -1722,5 +1722,80 @@ public async Task No_Ssl_Verify_Without_Ghes_Api_Url_Throws()
...
+                GhesApiUrl = GHES_API_URL,
+                AzureStorageConnectionString = AZURE_CONNECTION_STRING,
+                Wait = true,
+                KeepArchive = true
+            };
+            await _handler.Handle(args);
+
+            _mockTargetGithubApi.Verify(x => x.GetMigration(migrationId));
+            _mockOctoLogger.Verify(x => x.LogInformation($"GHES API URL: {GHES_API_URL}"), Times.Once);
+            _mockOctoLogger.Verify(x => x.LogInformation("AZURE STORAGE CONNECTION STRING: ***"), Times.Once);
```
ArinGhazarian, line 1796: nit: no need to assert these here since other tests are covering these. Just asserting that it doesn't delete should be sufficient enough.

ArinGhazarian (changes requested):
Thanks for the contribution @spindev, 

Your implementation is good but with a few shortcoming that need to be addressed:
1. The `--keep-archive` option needs to be added to `gei generate-script` command too.
2. The `--keep-archvie` flag is only useful for migration from GHES before 3.8. So if it was provided for non GHES < 3.8 migrations, we would like to log a warning message and let the user know that there is no archive to keep basically. It can be added in the `ValidateGHESOptions` method.
3. You need to log the option if it's provided with its value (in our case it'd be `true`).

timrogers:
> 3. You need to log the option if it's provided with its value (in our case it'd be `true`).

@spindev FYI, for #3, @ArinGhazarian is referring to a log statement like this one: https://github.com/github/gh-gei/blob/05d9dd5746c5eeda93fb8ef46b2457c3714c7b4f/src/gei/Handlers/MigrateRepoCommandHandler.cs#L402-L405.

src/OctoshiftCLI.Tests/gei/Handlers/MigrateRepoCommandHandlerTests.cs
spindev: Thx. Fixed.

spindev:
Thanks for your feedback @ArinGhazarian and @timrogers. Review again please. Hopefully I got everything :)

src/gei/Commands/GenerateScriptCommand.cs
```diff
@@ -37,6 +37,8 @@ public class GenerateScriptCommand : CommandBase<GenerateScriptCommandArgs, Gene
             AddOption(GithubSourcePat);
             AddOption(AdoPat);
             AddOption(Verbose);
+
```
ArinGhazarian, line 40: nit: remove the new line

src/gei/Handlers/GenerateScriptCommandHandler.cs
```diff
@@ -171,9 +171,14 @@ private void LogArgs(GenerateScriptCommandArgs args)
         {
             _log.LogInformation($"AWS REGION: {args.AwsRegion}");
         }
+
+        if (args.KeepArchive)
+        {
+            _log.LogInformation($"KEEP ARCHIVE: true");
```
ArinGhazarian, line 177: nit
```suggestion
            _log.LogInformation("KEEP ARCHIVE: true");
```

src/gei/Commands/MigrateRepoCommand.cs
```diff
@@ -43,6 +43,8 @@ public class MigrateRepoCommand : CommandBase<MigrateRepoCommandArgs, MigrateRep
             AddOption(GithubTargetPat);
             AddOption(AdoPat);
             AddOption(Verbose);
+
```
ArinGhazarian, line 46: nit: remove the new line

src/gei/Handlers/MigrateRepoCommandHandler.cs
```diff
@@ -491,6 +501,12 @@ private void ValidateGHESOptions(MigrateRepoCommandArgs args, bool cloudCredenti
                 {
                     _log.LogInformation("GHES version is 3.8.0 or later, no need to set cloud storage options here, please set in GHES admin UI.");
                 }
+
+                if (args.KeepArchive)
+                {
+                    _log.LogWarning("GHES version is 3.8.0 or later, there is no archive to keep on the user machine.");
+                }
```
ArinGhazarian, line 508: 1. This is good 👍 not sure about the copy, @timrogers can help here 😃 
2. Other than this, another validation can also be added for when `GhesApiUrl` doesn't have a value but `--keep-archive` does. Basically in the else part of `if (args.GhesApiUrl.HasValue())`. We still probably shouldn't throw and just log a warning as you did here with a message like `--keep-archive can only be provided with --ghes-api-url.`

ArinGhazarian (changes requested):
Looks good, just a few extra stuff.

src/gei/Handlers/MigrateRepoCommandHandler.cs
timrogers:
```suggestion
                if (args.KeepArchive)
                {
                    _log.LogWarning("Ignoring --keep-archive option because there is no downloaded archive to keep");
                }
```

timrogers: I think we could error in the case where the `--ghes-api-url` is not set and `--keep-archive` is set, because in this case, we know at the very beginning of the command that the combination of options is invalid, whereas in the above case, we only know once we've checked the GHES version.

spindev:
@ArinGhazarian and @timrogers thanks for the feedback. There you go for the next review :)

src/gei/Commands/GenerateScriptCommand.cs
```diff
@@ -111,6 +112,11 @@ public class GenerateScriptCommand : CommandBase<GenerateScriptCommandArgs, Gene
 
         public Option<bool> Verbose { get; } = new("--verbose");
 
+        public Option<bool> KeepArchive { get; } = new("--keep-archive")
+        {
+            Description = "Keeps the archive on the user machine after uploading to the storage account."
```
timrogers, line 117: One more thing from me - but otherwise I'm good! This would also apply to `MigrateRepoCommand`:

```suggestion
            Description = "Keeps the archive on this machine after uploading to the blob storage account. Only applicable for migrations from GitHub Enterprise Server versions before 3.8.0."
```

timrogers (approved):

src/gei/Commands/GenerateScriptCommand.cs
spindev: Done :)

src/gei/Handlers/MigrateRepoCommandHandler.cs
```diff
@@ -570,6 +586,11 @@ private void ValidateGHESOptions(MigrateRepoCommandArgs args, bool cloudCredenti
             {
                 throw new OctoshiftCliException("--ghes-api-url must be specified when --no-ssl-verify is specified.");
             }
+
+            if (args.KeepArchive)
+            {
+                throw new OctoshiftCliException("--ghes-api-url must be specified when --keep-archive is specified.");
+            }
```
ArinGhazarian, line 593: A unit test needs to be added for this since we're throwing here.

ArinGhazarian (changes requested):
It looks great 🎉 just add a unit test for the new validation that throws and this should be good to go.

ArinGhazarian (approved):
🎉

spindev:
@timrogers and @ArinGhazarian Thank you guys. Do I need to do additional stuff before the PR could be merged?

timrogers:
I think this is good to go, but I'd like to get a review from an engineer before we merge.

dylan-smith:
Can you add a bullet point to RELEASENOTES.md with this text:
`- Added `--keep-archive` flag to `gh gei migrate-repo` and `gh gei generate-script`. When migrating from GHES this will skip the step where we delete the archive from your machine, leaving it around as a local file.`

I tried to do it but I don't have permissions to push to your branch/fork. Once that's in there I'll merge this in.

Thanks!

dylan-smith (changes requested):
looks good, just needs a line in release notes and it's good to merge

spindev:
> looks good, just needs a line in release notes and it's good to merge

Done! :)

dylan-smith (approved):

PR #891, opened 2023-03-21
Improve error message when `CreateMigrationSource` is called with insufficient permissions

timrogers:
When using the `migrate-repo` command, under the hood, we call the `CreateMigrationSource` GraphQL mutation. If your target PAT doesn't have the correct permissions, the API returns an error like this:

> timrogers does not have the correct permissions to execute `CreateMigrationSource`

This error isn't very helpful, as it doesn't tell you what you should do to fix it. It would be great to have a more instructive error message that guides the customer on where they've gone wrong.

It isn't trivial for us to change the actual error returned by the API because it is part of GitHub's shared GraphQL logic. This takes a different approach: catching the specific error and decorating the message with some more helpful guidance on what to do next.

- [ ] Issue linked

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -1,1 +1,1 @@
-
+* Improve error message when `migrate-repo` is used with a target personal access token (PAT) with insufficient permissions
--- a/src/Octoshift/GithubApi.cs
+++ b/src/Octoshift/GithubApi.cs
@@ -221,9 +221,17 @@ namespace OctoshiftCLI
                 operationName = "createMigrationSource"
             };
 
-            var data = await _client.PostGraphQLAsync(url, payload);
+            try
+            {
+                var data = await _client.PostGraphQLAsync(url, payload);
+                return (string)data["data"]["createMigrationSource"]["migrationSource"]["id"];
+            }
+            catch (OctoshiftCliException ex)
+            {
+                CheckForMissingPermissionsError(ex);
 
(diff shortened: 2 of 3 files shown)


dylan-smith:
This all looks pretty good to me, a couple tweaks I would make:

1. You can use pattern matching in the catch block
2. Preserve the original exception in InnerException by passing it into the ctor

something like this

```
catch (OctoshiftCliException ex) when (ex.Message.Contains("blah blah permissions")
{
  throw new OctoshiftCliException(ex.Message + PERMISSIONS_ERROR_MSG, ex);
}
```

timrogers:
> This all looks pretty good to me, a couple tweaks I would make:
> 
> 1. You can use pattern matching in the catch block
> 2. Preserve the original exception in InnerException by passing it into the ctor
> 
> something like this
> 
> ```
> catch (OctoshiftCliException ex) when (ex.Message.Contains("blah blah permissions")
> {
>   throw new OctoshiftCliException(ex.Message + PERMISSIONS_ERROR_MSG, ex);
> }
> ```

Thanks! Two C# features I didn't know about there. Let me make those changes and ping for re-review.

P.S. I know I could have created an issue to do this, but this error message came up in a Bitbucket Server customer issue today, and I figured it would be quicker to fix it myself than write an issue.

timrogers:
@dylan-smith Updated! ✅ I refactored out `CheckForMissingPermissionsError` since it doesn't seem to fit well with the pattern matching, but I'm open to feedback.

dylan-smith (approved):

Issue #714, opened 2022-11-02
Parallelize integration tests


dylan-smith:
We already parallelize the tests between windows/linux/mac, but we could also parallelize between gei/ado2gh/bbs2gh.

The runtime is slowly creeping up, almost at 20 mins now. And as we add more tests that time is only going to increase.

spindev:
@dylan-smith Would like to help here out :) Will prepare a PR for that.

PR #896, opened 2023-03-22
Optimize build for integration tests

dylan-smith:
Related to #714 

Did a couple of things to improve the speed of the build:

1. Instead of compiling in every one of the (now) 15 integration test jobs, we'll have another set of 3 jobs run right before them that will compile for all 3 targets (win/linux/mac) but they will all be compiled ON ubuntu (where the compilation runs WAY faster than trying to compile on macos). We'll store the compiled binaries as build artifacts and the integration test jobs will download the binaries they need.

2. The ADO tests were the slowest of the bunch, and conveniently we have 2 completely independent ADO integration tests that take roughly equal amounts of run time. I sliced them up so it runs each of the 2 ADO tests in parallel. This required setting up a handful of new ADO/GH orgs so each test has it's own environment(s).

I also turned off the option to create PR comments with integration test results. When it was just one comment for all INT tests it was nice, but now it will be 5 separate comments for each "slice" of tests that we run. If we want to know the status of integration tests we can look in the build/checks associated with the PR.

- [x] Issue linked

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -151,7 +151,7 @@ jobs:
       fail-fast: false
       matrix:
         runner-os: [windows-latest, ubuntu-latest, macos-latest]
-        source-vcs: [Ado, Bbs, Ghes, Github]
+        source-vcs: [AdoBasic, AdoCsv, Bbs, Ghes, Github]
     runs-on: ${{ matrix.runner-os }}
     concurrency: integration-test-${{ matrix.source-vcs }}-${{ matrix.runner-os }}
     steps:
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -129,6 +129,9 @@ jobs:
         SKIP_WINDOWS: "true"
         SKIP_LINUX: "true"
 
+    - name: Display structure of downloaded files
+      run: ls -lR
+
     - name: Upload Binaries
       uses: actions/upload-artifact@v2
       with:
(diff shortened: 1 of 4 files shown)


.github/workflows/CI.yml
```diff
@@ -194,7 +227,7 @@ jobs:
       working-directory: ./gh-bbs2gh
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-
+
```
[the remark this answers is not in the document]
dylan-smith: done!

src/OctoshiftCLI.IntegrationTests/AdoCsvToGithub.cs
```diff
@@ -0,0 +1,79 @@
...
+{
+    [Collection("Integration Tests")]
+    public class AdoCsvToGithub : AdoToGithub
+    {
+        public AdoCsvToGithub(ITestOutputHelper output) : base(output)
+        {
+        }
+
+        [Fact]
+        public async Task With_Inventory_Report_Csv()
```
ArinGhazarian, line 15: Couldn't this have been just another method?

ArinGhazarian (approved):

src/OctoshiftCLI.IntegrationTests/AdoCsvToGithub.cs
dylan-smith: not sure what you mean? it is another method

dylan-smith: if you're asking about why I split it up into 2 separate classes (and then introduced a base class to avoid repeating the test setup), it's because "slicing" the tests in the build is easier to do at the class boundary (at least the way it's currently implemented).

PR #906, opened 2023-03-24
Add more helpful error message when BBS download archive is not found

### LINKED ISSUE #903: When the archive can't be found in a `bbs2gh` migration, suggest using the `--bbs-shared-home` option
In `bbs2gh` migrations, we download the generated archive from the Bitbucket Server instance using SSH or SMB.

If we look for the file and it doesn't exist, we output a message like this:

> [16:40] [ERROR] Source export archive (/var/atlassian/application-data/bitbucket/shared/data/migration/export/Bitbucket_export_1.tar) does not exist.

https://github.com/github/gh-gei/blob/a8ddc6960494f5b57a5eb86497bed477a3ebe7ac/src/bbs2gh/Services/BbsSshArchiveDownloader.cs#L94

If this happens, the most likely reason is that Bitbucket's data is not located in the default home directory, `/var/atlassian/application-data/bitbucket`. You can point to a non-default home directory by specifying the `--bbs-shared-home` directory.

We should update this error message to explain where we are looking for the file, how to check the correct `$BITBUCKET_HOME` path and how to use `--bbs-shared-home` to customise the search location.

This change may also apply to SMB, but I'm not 100% sure.

## Todo

- [ ] ...

## Dependencies

- ...

ArinGhazarian:
Closes #903 

- [x] Issue linked

When the BBS download archive is not found (either because of wrong path or missing archive), we used to throw a generic error message. We realized that in most cases the problem is that the default BBS shared home is not the one we expect. In that case, if the user hasn't already specified the `--bbs-shared-home`, with these changes we are now going to display a more helpful error message and let them know that they can set the aforementioned arg.

### DIFF
--- a/src/bbs2gh/Services/BbsSmbArchiveDownloader.cs
+++ b/src/bbs2gh/Services/BbsSmbArchiveDownloader.cs
@@ -208,10 +208,8 @@ public sealed class BbsSmbArchiveDownloader : IBbsArchiveDownloader
             : throw new OctoshiftCliException(
                 $"Couldn't create SMB file handle for \"{sharedFilePath}\" (Status Code: {status})." +
                 (IsObjectPathNotFoundStatus(status) && BbsSharedHomeDirectory is BbsSettings.DEFAULT_BBS_SHARED_HOME_DIRECTORY_WINDOWS
-                    ? " This usually happens because the archive could not be found at the default Bitbucket's shared home " +
-                      $"which is located in \"{BbsSettings.DEFAULT_BBS_SHARED_HOME_DIRECTORY_LINUX}\" if Bitbucket is installed on a Linux machine " +
-                      $"or in \"{BbsSettings.DEFAULT_BBS_SHARED_HOME_DIRECTORY_WINDOWS}\" if installed on a Windows machine. " +
-                      "You can point to a non default shared directory by specifying the --bbs-shared-home option."
+                    ? "This most likely means that your Bitbucket instance uses a non-default Bitbucket shared home directory, so we couldn't find your archive. " +
+                      "You can point the CLI to a non-default shared directory by specifying the --bbs-shared-home option."
                     : ""));
     }
 
--- a/src/bbs2gh/Services/BbsSmbArchiveDownloader.cs
+++ b/src/bbs2gh/Services/BbsSmbArchiveDownloader.cs
@@ -205,7 +205,14 @@ public sealed class BbsSmbArchiveDownloader : IBbsArchiveDownloader
 
         return IsSuccessStatus(status)
             ? sharedFileHandle
-            : throw new OctoshiftCliException($"Couldn't create SMB file handle for \"{sharedFilePath}\" (Status Code: {status}).");
+            : throw new OctoshiftCliException(
+                $"Couldn't create SMB file handle for \"{sharedFilePath}\" (Status Code: {status})." +
+                (IsObjectPathNotFoundStatus(status) && BbsSharedHomeDirectory is BbsSettings.DEFAULT_BBS_SHARED_HOME_DIRECTORY_WINDOWS
+                    ? " This usually happens because the archive could not be found at the default Bitbucket's shared home " +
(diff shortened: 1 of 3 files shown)


src/bbs2gh/Services/BbsSmbArchiveDownloader.cs
```diff
@@ -205,7 +205,14 @@ private object CreateFileHandle(ISMBFileStore fileStore, string sharedFilePath)
 
         return IsSuccessStatus(status)
             ? sharedFileHandle
-            : throw new OctoshiftCliException($"Couldn't create SMB file handle for \"{sharedFilePath}\" (Status Code: {status}).");
+            : throw new OctoshiftCliException(
+                $"Couldn't create SMB file handle for \"{sharedFilePath}\" (Status Code: {status})." +
+                (IsObjectPathNotFoundStatus(status) && BbsSharedHomeDirectory is BbsSettings.DEFAULT_BBS_SHARED_HOME_DIRECTORY_WINDOWS
+                    ? " This usually happens because the archive could not be found at the default Bitbucket's shared home " +
```
timrogers, line 211: Can we change this to?:

> This most likely means that your Bitbucket Server uses a non-default Bitbucket shared home directory, so we couldn't find your archive. You can point the CLI to a non-default shared directory by specifying the --bbs-shared-home option.

src/bbs2gh/Services/BbsSmbArchiveDownloader.cs
```diff
@@ -205,7 +205,14 @@ private object CreateFileHandle(ISMBFileStore fileStore, string sharedFilePath)
 
         return IsSuccessStatus(status)
             ? sharedFileHandle
-            : throw new OctoshiftCliException($"Couldn't create SMB file handle for \"{sharedFilePath}\" (Status Code: {status}).");
+            : throw new OctoshiftCliException(
+                $"Couldn't create SMB file handle for \"{sharedFilePath}\" (Status Code: {status})." +
+                (IsObjectPathNotFoundStatus(status) && BbsSharedHomeDirectory is BbsSettings.DEFAULT_BBS_SHARED_HOME_DIRECTORY_WINDOWS
```
timrogers, line 210: Really nice idea to only append this if the value hasn't been customised!

src/bbs2gh/Services/BbsSshArchiveDownloader.cs
```diff
@@ -91,7 +91,14 @@ public async Task<string> Download(long exportJobId, string targetDirectory = IB
 
         if (!_sftpClient.Exists(sourceExportArchiveFullPath))
         {
-            throw new OctoshiftCliException($"Source export archive ({sourceExportArchiveFullPath}) does not exist.");
+            throw new OctoshiftCliException(
+                $"Source export archive ({sourceExportArchiveFullPath}) does not exist." +
+                (BbsSharedHomeDirectory is BbsSettings.DEFAULT_BBS_SHARED_HOME_DIRECTORY_LINUX
+                    ? " This usually happens because the archive could not be found at the default Bitbucket's shared home " +
```
timrogers, line 97: Can we change this to?:

> This most likely means that your Bitbucket Server uses a non-default Bitbucket shared home directory, so we couldn't find your archive. You can point the CLI to a non-default shared directory by specifying the --bbs-shared-home option.

timrogers (changes requested):
Thanks for the speedy work on this! ❤️

src/bbs2gh/Services/BbsSmbArchiveDownloader.cs
ArinGhazarian: addressed in bf5a5d6dea62301209ea978394089a368aca2ac6

src/bbs2gh/Services/BbsSshArchiveDownloader.cs
ArinGhazarian: addressed in bf5a5d6dea62301209ea978394089a368aca2ac6

timrogers (approved):

Issue #629, opened 2022-09-15
Env Var validation in generated scripts


dylan-smith:
The migration scripts produced by `generate-script` expect certain environment variables to be set in order to work. To improve the customer experience we could have a bit of pwsh at the start of the script that validates the appropriate env vars are set and give the user a useful error message if not.

hfishback01:
I believe this is already addressed [here](https://github.com/github/gh-gei/pull/742).  Shall I close? @dylan-smith @ArinGhazarian 

![91CF3301-799C-480B-AC61-F7044997D3C1](https://user-images.githubusercontent.com/38845272/217091656-8fa83e09-7628-4e89-8516-c28fd7f3a41a.jpeg)

timrogers:
@hfishback01 That's a good spot! I think the issue of this issue was to add a more explicit validation error to the top of the script to make it a bit clearer - it could say, for example, "you must set environment variables X, Y and Z when running a migration script". Maybe it isn't strictly necessary, but I think it'll help users.

PR #924, opened 2023-03-31
Validate gei env vars in script

dylan-smith:
- [x] Issue linked

Implements part of https://github.com/github/gh-gei/issues/629

Added some powershell to the start of the script generated by gei generate-script that validates the proper environment variables are set before executing the rest of the script.  GH_PAT always needs to be set, and if `--ghes-api-url` has a value then either AZURE_STORAGE_CONNECTION_STRING or the AWS env vars need to be set.

I intentionally did not add any validation to the ADO migration scripts as that functionality will be removed from `gei` soon.

With GHES source, whether we need to validate Azure/AWS env vars or not depends on what version of GHES it is. We had an existing function in `MigrateRepoCommandHandler` that would retrieve the GHES version number, parse it, and determine whether we need blob storage creds or not. Instead of copy-pasting that I refactored it out into a new service - `GhesVersionCheckerService` - and injected it into both handlers (and added a few unit tests for the new service).


src/gei/Services/GhesVersionCheckerService.cs
```diff
@@ -0,0 +1,45 @@
+﻿using System;
+using System.Threading.Tasks;
+using OctoshiftCLI.Extensions;
+
+namespace OctoshiftCLI.GithubEnterpriseImporter.Services
+{
```
ArinGhazarian, line 6: nit: convert to file scoped namespace

src/gei/Services/GhesVersionCheckerService.cs
```diff
@@ -0,0 +1,45 @@
...
+    public class GhesVersionCheckerService
```
ArinGhazarian, line 7: nit: I wouldn't suffix `Service`:
```suggestion
    public class GhesVersionChecker
```

src/gei/Services/GhesVersionCheckerService.cs
```diff
@@ -0,0 +1,45 @@
...
+    {
+        private readonly OctoLogger _log;
+
+        public GhesVersionCheckerService(OctoLogger log)
+        {
+            _log = log;
+        }
+
+        public virtual async Task<bool> AreBlobCredentialsRequired(string ghesApiUrl, GithubApi api)
```
ArinGhazarian, line 16: `GithubApi` should definitely be injected into the service.

src/OctoshiftCLI.Tests/gei/Services/GhesVersionCheckerServiceTests.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System.Threading.Tasks;
+using FluentAssertions;
+using Moq;
+using OctoshiftCLI.GithubEnterpriseImporter.Services;
+using Xunit;
+
+namespace OctoshiftCLI.Tests.GithubEnterpriseImporter.Services
+{
```
ArinGhazarian, line 8: nit: convert to file scoped namespace to reduce indentation.

ArinGhazarian (changes requested):
The main thing is to inject the `GithubApi` into the service the rest looks good and will leave the nits up to you.

src/gei/Services/GhesVersionCheckerService.cs
dylan-smith: If we do that we have to create a factory class too, which I wanted to avoid.

src/OctoshiftCLI.Tests/gei/Services/GhesVersionCheckerServiceTests.cs
dylan-smith: done https://github.com/github/gh-gei/pull/924/commits/e39386d888d7451ff3c747fe3a1d0badd7a73c4d

src/gei/Services/GhesVersionCheckerService.cs
dylan-smith: done https://github.com/github/gh-gei/pull/924/commits/e39386d888d7451ff3c747fe3a1d0badd7a73c4d

src/gei/Services/GhesVersionCheckerService.cs
ArinGhazarian: It doesn't matter, because the `GithubApi` is a hard dependency and since we're doing DI we should inject it. I understand the pain of creating a factory but a good design surfaces the actual problems rather than suppressing them. Here our actual problem is the fact that we need a factory to create a `GithubApi` which can be addressed in a tech-debt issue later.

src/gei/Services/GhesVersionCheckerService.cs
dylan-smith: I changed this as you suggested, but I'm thinking maybe I should change it back. Everything else in our code has suffix's. Commands, Handlers, Factories, etc.

src/gei/Services/GhesVersionCheckerService.cs
dylan-smith: as much as it pained me, factory class added

src/gei/Factories/GhesVersionCheckerFactory.cs
```diff
@@ -0,0 +1,16 @@
+﻿using OctoshiftCLI.GithubEnterpriseImporter.Services;
+
+namespace OctoshiftCLI.GithubEnterpriseImporter.Factories
+{
```
ArinGhazarian, line 4: 1. nit: use file scoped namespace
2. nit: so far we've put other factories in the root folder. for consistency my suggestion would be to either put this factory in the root directory (with changing the namespace of course) too or move others into the `Factories` folder.

src/gei/Services/GhesVersionChecker.cs
```diff
@@ -0,0 +1,46 @@
...
+using OctoshiftCLI.Extensions;
+
+namespace OctoshiftCLI.GithubEnterpriseImporter.Services;
+
+public class GhesVersionChecker
+{
+    private readonly OctoLogger _log;
+    private readonly GithubApi _githubApi;
+
+    public GhesVersionChecker(OctoLogger log, GithubApi githubApi)
```
ArinGhazarian, line 12: 🎉

ArinGhazarian (approved):
Everything looks good, just added two non-blocking nits.

src/gei/Factories/GhesVersionCheckerFactory.cs
dylan-smith: there's gotta be a way to make dotnet format catch the file scoped namespace for me. I'll do that, but I was gonna do a separate PR to move the other factories around (doing it immediately after this merges).

ArinGhazarian: I use `Rider` and it lets me know about those kind improvements 😃

PR #647, opened 2022-09-20
Add repo visibility to migrate-repo commands

tambling:
This PR adds the first half of https://github.com/github/gh-gei/issues/5: adding `--target-repo-visibility` to our `migrate-repo` commands, which will then send that setting up to the monolith's `StartRepositoryMigration` mutation (which handles the rest).

- [x] Did you write/update appropriate tests
- [x] Release notes updated (if appropriate)
- [x] Appropriate logging output
- [x] Issue linked
- [ ] Docs updated (or issue created)

PR #953, opened 2023-04-11
Add option to set target repo visibility

dylan-smith:
Implements part of #5 (FINALLY!)

- [x] Issue linked

This was originally implemented by @tambling in #647 . However there was some back and forth on some design questions and it never got merged, and now that PR/branch is so far behind main I figured it would be easier to just redo the changes in a new branch rather than trying to resolve the merge/rebase conflicts with the previous PR/branch.

This adds a `--target-repo-visibility` argument to all 3 `migrate-repo` commands. Octoshift backend already supports this arg in the API's, we just never used it (until now).

I will do a separate PR that updates `generate-script`.


src/Octoshift/Services/GithubApi.cs
```diff
@@ -301,7 +301,7 @@ public virtual async Task<string> CreateGhecMigrationSource(string orgId)
         }
     }
 
-    public virtual async Task<string> StartMigration(string migrationSourceId, string sourceRepoUrl, string orgId, string repo, string sourceToken, string targetToken, string gitArchiveUrl = null, string metadataArchiveUrl = null, bool skipReleases = false, bool lockSource = false)
+    public virtual async Task<string> StartMigration(string migrationSourceId, string sourceRepoUrl, string orgId, string repo, string sourceToken, string targetToken, string gitArchiveUrl = null, string metadataArchiveUrl = null, bool skipReleases = false, string targetRepoVisibility = null, bool lockSource = false)
```
ArinGhazarian, line 304: I definitely recommend using an `enum` for the repo visibility, it's more strongly types and less error prone:
```c#
enum RepoVisiblity
{
    Private,
    Public,
    Internal
}
```

src/OctoshiftCLI.Tests/Octoshift/Services/GithubApiTests.cs
```diff
@@ -888,6 +890,7 @@
                 accessToken = sourceToken,
                 githubPat = targetToken,
                 skipReleases = false,
+                targetRepoVisibility = (string)null,
```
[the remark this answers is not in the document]
ArinGhazarian: The CodeQL warning is legit here, no need to extra cast to `string`.

src/Octoshift/Services/GithubApi.cs
dylan-smith: I'm not sure that's necessary given the validation done by the command line library with this line:
```
AddOption(TargetRepoVisibility.FromAmong("public", "private", "internal"));
```

src/OctoshiftCLI.Tests/Octoshift/Services/GithubApiTests.cs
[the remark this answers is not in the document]
dylan-smith: this is necessary because it's an anonymous type and without the cast it doesn't know what type to use for the null (it actually won't compile without the cast IIRC)

src/Octoshift/Services/GithubApi.cs
ArinGhazarian: This is the in the app level but it's also a good idea to do it in the API level too. IMO in general it's better API design and makes it easier to follow the logic. Despite the fact where it's gonna get used.

ArinGhazarian: This is the in the app level but it's also a good idea to do it in the API level too. IMO in general it's better API design and makes it easier to follow the logic. Despite the fact where it's gonna get used.

ArinGhazarian (approved):

PR #956, opened 2023-04-13
Add target-repo-visibility to generated scripts

### LINKED ISSUE #5: Set repo visibility
Customers have asked if it's possible to add an option to let them set a state other than private for a repo when it's done migrating.   This capability is already implemented in Octoshift, so this issue tracks passing the appropriate data to Octoshift via the CLI to use this feature.

The customer can configure this in the CLI. When they run generate script we will add a `--target-repo-visibility` option to migrate repo that by default matches the state that the source repo was in. 

## Todo
- [ ] Update the CLI generate script to add a `--target-repo-visibility` to the `migrate-repo` command that adds the state that the source repo was in (i.e. private/public/internal). If the field isn't included then the default should be private. 
- [ ] `--target-repo-visibility` should be included in for ado2gh customers. We can just default to private for that scenario.

dylan-smith:
- [x] Issue linked

Closes #5 

Updated all 3 `generate-script` commands to always include the `--target-repo-visibility` arg. For `gei` it will inspect the source-repo and set the value appropriately in the script. For `ado2gh` and `bbs2gh` it will always set the value to `private`.

Originally I added a separate function `GithubApi.GetRepoVisibility()` but this resulted in an API call per repo when generating the script which made it significantly slower for large orgs. The API we already call to get the list of repos also returns the visibility, so I updated `GithubApi.GetRepos()` to return a tuple with repo name and visibility.


src/gei/Handlers/GenerateScriptCommandHandler.cs
```diff
@@ -220,10 +220,9 @@ private async Task<IEnumerable<string>> GetGithubRepos(GithubApi github, string
 
         _log.LogInformation($"GITHUB ORG: {githubOrg}");
         var repos = await github.GetRepos(githubOrg);
-
-        foreach (var repo in repos)
+        foreach (var (Name, _) in repos)
```
ArinGhazarian, line 223: use lowercase variable name here (`name` instead of `Name`). Although properties in value tuples should follow property naming guidelines, in case of deconstruction like this, the properties are going to be used as local variables so they should follow local variable naming guidelines.

src/gei/Handlers/GenerateScriptCommandHandler.cs
```diff
@@ -328,10 +327,10 @@ private async Task<string> GenerateParallelGithubScript(IEnumerable<string> repo
         content.AppendLine("# === Queuing repo migrations ===");
 
         // Queuing migrations
-        foreach (var repo in repos)
+        foreach (var (Name, Visibility) in repos)
```
ArinGhazarian, line 330: same comment about variable naming applies here too.

ArinGhazarian (commented):
looks good, just dropped two comments about naming deconstructed variable naming.

RELEASENOTES.md
```diff
@@ -1 +1,2 @@
 - `migrate-repo` commands now take a `--target-repo-visibility` flag (`"public"`, `"private"`, or `"internal"`; defaults to `"private"`) to set the visibility of the imported repo
+- `gh gei generate-script` will not inspect the source repo visibility and add the appropriate `--target-repo-visibility` flag to the generated script. `ado2gh generate-script` and `bbs2gh generate-script` will include the `--target-repo-visibility` flag in the generated script but it will always be set to private.
```
ArinGhazarian, line 2:
```suggestion
- `gh gei generate-script` will now inspect the source repo visibility and add the appropriate `--target-repo-visibility` flag to the generated script. `ado2gh generate-script` and `bbs2gh generate-script` will include the `--target-repo-visibility` flag in the generated script but it will always be set to private.
```

src/gei/Handlers/GenerateScriptCommandHandler.cs
dylan-smith: good point, dotnet format actually did the refactor to deconstruction for me, so those are the names dotnet format generated

ArinGhazarian: haha! it should be a bug then :)

dylan-smith: fixed

src/gei/Handlers/GenerateScriptCommandHandler.cs
dylan-smith: fixed

src/gei/Handlers/GenerateScriptCommandHandler.cs
ArinGhazarian: unless GH is not showing me the latest changes, I guess you missed this one

dylan-smith: you're right, fixed for real now

ArinGhazarian (approved):
🎉

PR #957, opened 2023-04-14
Improve "insufficient permissions" error message to refer to a specific organization

timrogers:
When we try to call the `createMigrationSource` mutation and we get a permissions error, we already decorate the error message returned from the GraphQL API to be more specific on what you should do to solve it.

Following feedback from customers, this improves that error message to be extremely explicit about what organization you need to have permissions for, so there is no room for doubt.

Ideally, we wouldn't have to make a change like this in the CLI, but the error message we are improving is part of GitHub's core GraphQL API code, so it's non-trivial to change. Improving it in the client layer is the right step to give the best CLI experience possible.

We did experience with alternatives implementations of this change which were more abstract and reusable, but this felt like the right approach with the right level of complexity when we don't know how many other errors we'll want to augment.

- [ ] Issue linked

### DIFF
--- a/src/Octoshift/Services/GithubApi.cs
+++ b/src/Octoshift/Services/GithubApi.cs
@@ -17,8 +17,6 @@ public class GithubApi
     private readonly string _apiUrl;
     private readonly RetryPolicy _retryPolicy;
 
-    private const string INSUFFICIENT_PERMISSIONS_HELP_MESSAGE = ". Please check that (a) you are an organization owner or you have been granted the migrator role and (b) your personal access token has the correct scopes. For more information, see https://docs.github.com/en/migrations/using-github-enterprise-importer/preparing-to-migrate-with-github-enterprise-importer/managing-access-for-github-enterprise-importer.";
-
     public GithubApi(GithubClient client, string apiUrl, RetryPolicy retryPolicy)
     {
         _client = client;
@@ -228,15 +226,8 @@ public class GithubApi
             operationName = "createMigrationSource"
         };
 
-        try
-        {
-            var data = await _client.PostGraphQLAsync(url, payload);
-            return (string)data["data"]["createMigrationSource"]["migrationSource"]["id"];
-        }
-        catch (OctoshiftCliException ex) when (ex.Message.Contains("not have the correct permissions to execute"))
-        {
-            throw new OctoshiftCliException(ex.Message + INSUFFICIENT_PERMISSIONS_HELP_MESSAGE, ex);
-        }
+        var data = await _client.PostGraphQLAsync(url, payload);
+        return (string)data["data"]["createMigrationSource"]["migrationSource"]["id"];
     }
 
     public virtual async Task<string> CreateBbsMigrationSource(string orgId)
@@ -259,15 +250,8 @@ public class GithubApi
             operationName = "createMigrationSource"
         };
 
-        try
-        {
-            var data = await _client.PostGraphQLAsync(url, payload);
-            return (string)data["data"]["createMigrationSource"]["migrationSource"]["id"];
-        }
-        catch (OctoshiftCliException ex) when (ex.Message.Contains("not have the correct permissions to execute"))
-        {
-            throw new OctoshiftCliException(ex.Message + INSUFFICIENT_PERMISSIONS_HELP_MESSAGE, ex);
-        }
+        var data = await _client.PostGraphQLAsync(url, payload);
+        return (string)data["data"]["createMigrationSource"]["migrationSource"]["id"];
     }
 
     public virtual async Task<string> CreateGhecMigrationSource(string orgId)
@@ -290,15 +274,8 @@ public class GithubApi
             operationName = "createMigrationSource"
         };
 
-        try
-        {
-            var data = await _client.PostGraphQLAsync(url, payload);
-            return (string)data["data"]["createMigrationSource"]["migrationSource"]["id"];
-        }
-        catch (OctoshiftCliException ex) when (ex.Message.Contains("not have the correct permissions to execute"))
-        {
-            throw new OctoshiftCliException(ex.Message + INSUFFICIENT_PERMISSIONS_HELP_MESSAGE, ex);
-        }
+        var data = await _client.PostGraphQLAsync(url, payload);
+        return (string)data["data"]["createMigrationSource"]["migrationSource"]["id"];
     }
 
     public virtual async Task<string> StartMigration(string migrationSourceId, string sourceRepoUrl, string orgId, string repo, string sourceToken, string targetToken, string gitArchiveUrl = null, string metadataArchiveUrl = null, bool skipReleases = false, string targetRepoVisibility = null, bool lockSource = false)
--- a/src/OctoshiftCLI.Tests/ado2gh/Commands/MigrateRepo/MigrateRepoCommandHandlerTests.cs
+++ b/src/OctoshiftCLI.Tests/ado2gh/Commands/MigrateRepo/MigrateRepoCommandHandlerTests.cs
@@ -216,7 +216,7 @@ public class MigrateRepoCommandHandlerTests
         }))
             .Should()
             .ThrowAsync<OctoshiftCliException>()
-            .WithMessage($"monalisa does not have the correct permissions to execute `CreateMigrationSource`. Please check that (a) you are a member of the `{GITHUB_ORG}` organization, (b) you are an organization owner or you have been granted the migrator role and (c) your personal access token has the correct scopes. For more information, see https://docs.github.com/en/migrations/using-github-enterprise-importer/preparing-to-migrate-with-github-enterprise-importer/managing-access-for-github-enterprise-importer.");
+            .WithMessage($"monalisa does not have the correct permissions to execute `CreateMigrationSource`. Please check that:\n  (a) you are a member of the `{GITHUB_ORG}` organization,\n  (b) you are an organization owner or you have been granted the migrator role and\n  (c) your personal access token has the correct scopes.\nFor more information, see https://docs.github.com/en/migrations/using-github-enterprise-importer/preparing-to-migrate-with-github-enterprise-importer/managing-access-for-github-enterprise-importer.");
     }
 
     [Fact]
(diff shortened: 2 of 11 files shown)


src/Octoshift/Services/GithubApi.cs
```diff
@@ -206,7 +204,7 @@ public virtual async Task<string> GetEnterpriseId(string enterpriseName)
         }
     }
 
-    public virtual async Task<string> CreateAdoMigrationSource(string orgId, string adoServerUrl)
+    public virtual async Task<string> CreateAdoMigrationSource(string orgId, string orgName, string adoServerUrl)
```
timrogers, line 207: It's definitely a bit of a smell that we have to pass the `orgName` to this method in order to generate the nice error message.

I would be open to refactoring this so that `GithubApi` throws an exception with the raw error message from the API, and then the caller is responsible for catching that error and decorating it. That would work - the only disadvantage is that this class is easier to unit test that the command handler.

src/Octoshift/Services/GithubApi.cs
```diff
@@ -297,7 +297,8 @@ public virtual async Task<string> CreateGhecMigrationSource(string orgId)
         }
         catch (OctoshiftCliException ex) when (ex.Message.Contains("not have the correct permissions to execute"))
         {
-            throw new OctoshiftCliException(ex.Message + INSUFFICIENT_PERMISSIONS_HELP_MESSAGE, ex);
+            var errorSuffix = GenerateInsufficientPermissionsHelpMessage(orgName);
+            throw new OctoshiftCliException(ex.Message + errorSuffix, ex);
```
ArinGhazarian, line 301: Instead of concatenating two strings it's more efficient to use string interpolation: `$"{ex.Message}. {errorSuffix}"`

ArinGhazarian (changes requested):

src/Octoshift/Services/GithubApi.cs
timrogers: Thanks! Happy to change that.

timrogers:
@ArinGhazarian I don't think your suggestion of changing the GraphQL query to return the organization name will work here. We're handling a GraphQL error here, and given that there wa an error, the fields that we requested are not returned. All we get is the error message GitHub returns.

If we don't want to be forced to pass the org name to the `GithubApi` methods, then I think the best option is to move this error handling to the callsite.

(Adding this comment here because, for some reason, I can't comment on your review comment!)

timrogers:
@ArinGhazarian I don't think your suggestion of changing the GraphQL query to return the organization name will work here. We're handling a GraphQL error here, and given that there wa an error, the fields that we requested are not returned. All we get is the error message GitHub returns.

If we don't want to be forced to pass the org name to the `GithubApi` methods, then I think the best option is to move this error handling to the callsite. But there might be another option I'm not thinking of (cc @dylan-smith).

(Adding this comment here because, for some reason, I can't comment on your review comment!)

ArinGhazarian:
@timrogers yeah you're right, all we have is the error in this context. in that case I would say it's better to move it to the callsite instead of passing the org name to bunch of method just to be able to generate an error message.

src/Octoshift/Services/GithubApi.cs
timrogers: As discussed below, we won't be able to make GraphQL return the org name because this is an error response where we don't get to control the fields.

src/Octoshift/Services/MigrateRepoErrorHandler.cs
```diff
@@ -0,0 +1,18 @@
+using System.Diagnostics.Contracts;
+
+namespace OctoshiftCLI.Services;
+
+public static class MigrateRepoErrorHandler
+{
+    public static OctoshiftCliException DecorateInsufficientPermissionsException(OctoshiftCliException ex, string orgName)
+    {
+        Contract.Requires(ex != null);
```
timrogers, line 9: I haven't seen this before in C#, but the formatter seemed to want me to add it. Very open to alternative approaches here!

src/Octoshift/Services/MigrateRepoErrorHandler.cs
```diff
@@ -0,0 +1,18 @@
...
+        var errorSuffix = GenerateInsufficientPermissionsHelpMessage(orgName);
+        return new OctoshiftCliException($"{ex.Message}. {errorSuffix}", ex);
```
dylan-smith, line 11:
```suggestion
        var errorSuffix = GenerateInsufficientPermissionsHelpMessage(orgName);
        return new OctoshiftCliException($"{ex?.Message}. {errorSuffix}", ex);
```

note the ? in ex?.Message. that should get rid of the formatter error.

dylan-smith (changes requested):

src/bbs2gh/Handlers/MigrateRepoCommandHandler.cs
```diff
@@ -215,7 +215,18 @@ private async Task ImportArchive(MigrateRepoCommandArgs args, string archiveUrl
...
+
+        try
+        {
+            migrationSourceId = await _githubApi.CreateBbsMigrationSource(githubOrgId);
+        }
+        catch (OctoshiftCliException ex) when (ex.Message.Contains("not have the correct permissions to execute"))
+        {
+            var exception = MigrateRepoErrorHandler.DecorateInsufficientPermissionsException(ex, args.GithubOrg);
+            throw exception;
+        }
```
ArinGhazarian, line 229: To rewrite this with my suggested `MessasgeEnhancer` class:
```suggestion
        catch (OctoshiftCliException ex) when (ex.Message.HasEnhancement())
        {
            var message = ex.Message.Enhance((original, improved) => $"{original}. {string.Format(improved, args.GithubOrg)}" ;
            throw new OctoshiftCliException(message, ex);
        }
```

Note that in this particular case, we need to use the factory because the improved message is actually a template that should contain the org name. For this kind of more dynamic creation we cannot use string interpolation but we can use `string.Fromat` instead.

ArinGhazarian (changes requested):

timrogers:

src/Octoshift/MessageEnhancer.cs
```diff
@@ -0,0 +1,28 @@
...
+    public static string Enhance(string message, Func<string, string, string> messageFactory = null)
+    {
+        messageFactory ??= (originalMessage, improvedMessage) => $"{originalMessage}{improvedMessage}";
+
+        var key = GetKeyForMessage(message);
+        return key is null ? message : messageFactory(message, _map[key]);
+    }
+
+    private static string GetKeyForMessage(string message) => _map.Keys.SingleOrDefault(key => message.Contains(key, StringComparison.OrdinalIgnoreCase));
+}
```
dylan-smith, line 28: This seems like it's getting kinda convoluted. I'll give more feedback when I have a bit more time, but my gut reaction is there's gotta be a more straightforward way to do this. Maybe a custom exception that inherits from OctoshiftCliException.

dylan-smith (changes requested):

src/Octoshift/MessageEnhancer.cs
ArinGhazarian: Well, I agree that for just this one case it's a bit of an overkill, but my main goal for recommending this was to cover future use-cases too. It seems like we are going to have quite a few error message improvements like this and the problem is not only the exception message but also those `if`s and `when`s to look for those key sentences in error messages all over different places. With this approach we can aggregate everything in one place.
One other approach would be to create some sort of constants holder class to store those enhanced messages and also the key sentences that we are going to look for (in `when` clauses).

timrogers: I think this design will kinda work, but the most complex part is "variables" inside the messages. In this case, we want to interpolate the organization name, but in other messages, it might be something else that we want to interpolate.

If we're going to design a generic solution to use in future use cases, I think we need to figure that out.

ArinGhazarian (approved):
I am 👍 but let's wait for @dylan-smith's comments too before merging.

src/Octoshift/MessageEnhancer.cs
timrogers: That makes sense. But then the class calling `Enhance` needs to know ahead of time what error it's going to get back right, because it has to know what the placeholders will be? Does that make sense? I might not be explaining clearly!

ArinGhazarian: well we know what error message we are going to improve right? the improved message is going to be a value in the dictionary which will contain those placeholders. Now if an error message is enhanceable (which means if any of the predefined keys in or dictionary were found in the message text), then we are going the get the enhanced version and replace the placeholders with required info. Of course we need to carefully choose the keys in such a way that they don't exist if more than one message and then with the lambda (which is the dynamic part) we can compose the end result message.

timrogers:
@synthead Here's an update version of this PR after our discussion last week. I've reverted to a simple solution for now, and we can move to something more abstract later if needed.

src/OctoshiftCLI.Tests/ado2gh/Commands/MigrateRepo/MigrateRepoCommandHandlerTests.cs
```diff
@@ -188,6 +188,37 @@ public async Task Happy_Path_With_Wait()
...
+            AdoOrg = ADO_ORG,
+            AdoTeamProject = ADO_TEAM_PROJECT,
+            AdoRepo = ADO_REPO,
+            GithubOrg = GITHUB_ORG,
+            GithubRepo = GITHUB_REPO,
+            Wait = true,
+        }))
+            .Should()
+            .ThrowAsync<OctoshiftCliException>()
+            .WithMessage($"monalisa does not have the correct permissions to execute `CreateMigrationSource`. Please check that (a) you are a member of the `{GITHUB_ORG}` organization, (b) you are an organization owner or you have been granted the migrator role and (c) your personal access token has the correct scopes. For more information, see https://docs.github.com/en/migrations/using-github-enterprise-importer/preparing-to-migrate-with-github-enterprise-importer/managing-access-for-github-enterprise-importer.");
```
synthead, line 219: This console output is nearly 500 columns wide :sweat_smile: Could we add some newlines in the console output so that it looks something like this?

```
monalisa does not have the correct permissions to execute `CreateMigrationSource`
Please check that:
  (a) you are a member of the `{GITHUB_ORG}` organization.
  (b) you are an organization owner or you have been granted the migrator role.
  (c) your personal access token has the correct scopes.
For more information, see https://docs.github.com/en/migrations/using-github-enterprise-importer/preparing-to-migrate-with-github-enterprise-importer/managing-access-for-github-enterprise-importer.
```

synthead (commented):
Looking good!  Had a thought about console output!  Let me know what you think!

synthead (approved):

PR #961, opened 2023-04-18
Include the migration ID in the output filename by default when running `download-logs`

timrogers:
Following customer feedback, this updates the shared `download-logs` command available across our CLIs to include the migration ID in the filename by default - although this can, of course, be overridden.

This means that migration logs where you dry-run the same repo multiple times won't try to overwrite one another, and will always have a unique filename.

- [ ] Issue linked

### DIFF
--- a/src/Octoshift/Handlers/DownloadLogsCommandHandler.cs
+++ b/src/Octoshift/Handlers/DownloadLogsCommandHandler.cs
@@ -63,8 +63,6 @@ public class DownloadLogsCommandHandler : ICommandHandler<DownloadLogsCommandArg
             _log.LogInformation($"MIGRATION LOG FILE: {args.MigrationLogFile}");
         }
 
-        args.MigrationLogFile ??= $"migration-log-{args.GithubOrg}-{args.GithubRepo}.log";
-
         if (FileExists(args.MigrationLogFile))
         {
             if (!args.Overwrite)
@@ -75,20 +73,17 @@ public class DownloadLogsCommandHandler : ICommandHandler<DownloadLogsCommandArg
             _log.LogWarning($"Overwriting {args.MigrationLogFile} due to --overwrite option.");
         }
 
-        var result = await _retryPolicy.RetryOnResult(async () => await _githubApi.GetMigrationLogUrl(args.GithubOrg, args.GithubRepo), string.Empty,
+        var result = await _retryPolicy.RetryOnResult<(string MigrationLogURL, string MigrationID)>(async () => await _githubApi.GetMigrationLogUrl(args.GithubOrg, args.GithubRepo), result => string.IsNullOrEmpty(result.MigrationLogURL),
             "Waiting for migration log to populate...");
-
-        if (result.Outcome == OutcomeType.Successful && result.Result is null)
-        {
-            throw new OctoshiftCliException($"Migration for repository {args.GithubRepo} not found!");
-        }
 
         if (result.Outcome == OutcomeType.Failure)
         {
             throw new OctoshiftCliException($"Migration log for repository {args.GithubRepo} unavailable!");
         }
 
-        var logUrl = result.Result;
+        (var logUrl, var migrationId) = result.Result;
+
+        args.MigrationLogFile ??= $"migration-log-{args.GithubOrg}-{args.GithubRepo}-{migrationId}.log";
 
         _log.LogInformation($"Downloading log for repository {args.GithubRepo} to {args.MigrationLogFile}...");
         await _httpDownloadService.DownloadToFile(logUrl, args.MigrationLogFile);
--- a/src/Octoshift/Handlers/DownloadLogsCommandHandler.cs
+++ b/src/Octoshift/Handlers/DownloadLogsCommandHandler.cs
@@ -73,7 +73,7 @@ public class DownloadLogsCommandHandler : ICommandHandler<DownloadLogsCommandArg
             _log.LogWarning($"Overwriting {args.MigrationLogFile} due to --overwrite option.");
         }
 
-        var result = await _retryPolicy.RetryOnResult<(string MigrationLogUrl, string MigrationId)?>(async () => await _githubApi.GetMigrationLogUrl(args.GithubOrg, args.GithubRepo), null,
+        var result = await _retryPolicy.RetryOnResult<(string MigrationLogUrl, string MigrationId)?>(async () => await _githubApi.GetMigrationLogUrl(args.GithubOrg, args.GithubRepo), result => string.IsNullOrEmpty(result.Value.MigrationLogUrl),
             "Waiting for migration log to populate...");
 
         if (result.Outcome == OutcomeType.Failure)
--- a/src/Octoshift/Handlers/DownloadLogsCommandHandler.cs
+++ b/src/Octoshift/Handlers/DownloadLogsCommandHandler.cs
@@ -70,7 +70,7 @@ public class DownloadLogsCommandHandler : ICommandHandler<DownloadLogsCommandArg
             throw new OctoshiftCliException($"File {args.MigrationLogFile} already exists! Use --overwrite to overwrite this file.");
         }
 
-        var result = await _retryPolicy.RetryOnResult<(string MigrationLogUrl, string MigrationId)?>(async () => await _githubApi.GetMigrationLogUrl(args.GithubOrg, args.GithubRepo), r => r?.MigrationLogUrl.IsNullOrWhiteSpace() ?? false,
+        var result = await _retryPolicy.RetryOnResult(async () => await _githubApi.GetMigrationLogUrl(args.GithubOrg, args.GithubRepo), r => r?.MigrationLogUrl.IsNullOrWhiteSpace() ?? false,
             "Waiting for migration log to populate...");
 
         if (result.Outcome == OutcomeType.Successful && result.Result is null)
--- a/src/Octoshift/Handlers/DownloadLogsCommandHandler.cs
+++ b/src/Octoshift/Handlers/DownloadLogsCommandHandler.cs
@@ -63,15 +63,7 @@ public class DownloadLogsCommandHandler : ICommandHandler<DownloadLogsCommandArg
             _log.LogInformation($"MIGRATION LOG FILE: {args.MigrationLogFile}");
         }
 
-        if (FileExists(args.MigrationLogFile))
-        {
-            if (!args.Overwrite)
-            {
-                throw new OctoshiftCliException($"File {args.MigrationLogFile} already exists!  Use --overwrite to overwrite this file.");
-            }
-
-            _log.LogWarning($"Overwriting {args.MigrationLogFile} due to --overwrite option.");
-        }
+        CheckIfOutputFileAlreadyExists(args.MigrationLogFile, args.Overwrite);
 
         var result = await _retryPolicy.RetryOnResult<(string MigrationLogUrl, string MigrationId)?>(async () => await _githubApi.GetMigrationLogUrl(args.GithubOrg, args.GithubRepo), result => string.IsNullOrEmpty(result.Value.MigrationLogUrl),
             "Waiting for migration log to populate...");
@@ -91,10 +83,30 @@ public class DownloadLogsCommandHandler : ICommandHandler<DownloadLogsCommandArg
 
             args.MigrationLogFile ??= $"migration-log-{args.GithubOrg}-{args.GithubRepo}-{migrationId}.log";
 
+            // We already checked if the file exists above for the case where the user explicitly picked their own
+            // filename. This handles the case where the filename has been set to the default based on the inputs
+            // and migration ID.
+            CheckIfOutputFileAlreadyExists(args.MigrationLogFile, args.Overwrite);
+
             _log.LogInformation($"Downloading log for repository {args.GithubRepo} to {args.MigrationLogFile}...");
             await _httpDownloadService.DownloadToFile(logUrl, args.MigrationLogFile);
 
             _log.LogSuccess($"Downloaded {args.GithubRepo} log to {args.MigrationLogFile}.");
+        }
+    }
+
+    private void CheckIfOutputFileAlreadyExists(string outputPath, bool shouldOverwrite)
+    {
+        if (FileExists(outputPath))
+        {
+            if (shouldOverwrite)
+            {
+                throw new OctoshiftCliException($"File {outputPath} already exists!  Use --overwrite to overwrite this file.");
+            }
+            else
+            {
+                _log.LogWarning($"Overwriting {outputPath} due to --overwrite option.");
+            }
         }
     }
 }
--- a/src/Octoshift/Handlers/DownloadLogsCommandHandler.cs
+++ b/src/Octoshift/Handlers/DownloadLogsCommandHandler.cs
@@ -73,7 +73,7 @@ public class DownloadLogsCommandHandler : ICommandHandler<DownloadLogsCommandArg
             _log.LogWarning($"Overwriting {args.MigrationLogFile} due to --overwrite option.");
         }
 
-        var result = await _retryPolicy.RetryOnResult<(string MigrationLogURL, string MigrationID)>(async () => await _githubApi.GetMigrationLogUrl(args.GithubOrg, args.GithubRepo), result => string.IsNullOrEmpty(result.MigrationLogURL),
+        var result = await _retryPolicy.RetryOnResult<(string MigrationLogUrl, string MigrationId)?>(async () => await _githubApi.GetMigrationLogUrl(args.GithubOrg, args.GithubRepo), null,
             "Waiting for migration log to populate...");
 
         if (result.Outcome == OutcomeType.Failure)
@@ -81,13 +81,20 @@ public class DownloadLogsCommandHandler : ICommandHandler<DownloadLogsCommandArg
             throw new OctoshiftCliException($"Migration log for repository {args.GithubRepo} unavailable!");
         }
 
-        (var logUrl, var migrationId) = result.Result;
+        if (result.Outcome == OutcomeType.Successful && result.Result is null)
+        {
+            throw new OctoshiftCliException($"Migration for repository {args.GithubRepo} not found!");
+        }
+        else
+        {
+            var (logUrl, migrationId) = result.Result.Value;
 
-        args.MigrationLogFile ??= $"migration-log-{args.GithubOrg}-{args.GithubRepo}-{migrationId}.log";
+            args.MigrationLogFile ??= $"migration-log-{args.GithubOrg}-{args.GithubRepo}-{migrationId}.log";
 
-        _log.LogInformation($"Downloading log for repository {args.GithubRepo} to {args.MigrationLogFile}...");
-        await _httpDownloadService.DownloadToFile(logUrl, args.MigrationLogFile);
+            _log.LogInformation($"Downloading log for repository {args.GithubRepo} to {args.MigrationLogFile}...");
+            await _httpDownloadService.DownloadToFile(logUrl, args.MigrationLogFile);
 
-        _log.LogSuccess($"Downloaded {args.GithubRepo} log to {args.MigrationLogFile}.");
+            _log.LogSuccess($"Downloaded {args.GithubRepo} log to {args.MigrationLogFile}.");
+        }
     }
 }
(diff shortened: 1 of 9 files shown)


src/Octoshift/Handlers/DownloadLogsCommandHandler.cs
```diff
@@ -75,20 +73,17 @@ public async Task Handle(DownloadLogsCommandArgs args)
             _log.LogWarning($"Overwriting {args.MigrationLogFile} due to --overwrite option.");
         }
 
-        var result = await _retryPolicy.RetryOnResult(async () => await _githubApi.GetMigrationLogUrl(args.GithubOrg, args.GithubRepo), string.Empty,
+        var result = await _retryPolicy.RetryOnResult<(string MigrationLogURL, string MigrationID)>(async () => await _githubApi.GetMigrationLogUrl(args.GithubOrg, args.GithubRepo), result => string.IsNullOrEmpty(result.MigrationLogURL),
             "Waiting for migration log to populate...");
 
-        if (result.Outcome == OutcomeType.Successful && result.Result is null)
```
timrogers, line 81: `GetMigrationLogUrl` now throws an exception in this case, rather than returning `null`. This feels like the natural control flow for this unhappy path.

src/Octoshift/RetryPolicy.cs
```diff
@@ -29,7 +29,7 @@ public async Task<T> HttpRetry<T>(Func<Task<T>> func, Func<HttpRequestException,
             return await policy.ExecuteAsync(func);
         }
 
-        public async Task<PolicyResult<T>> RetryOnResult<T>(Func<Task<T>> func, T resultFilter, string retryLogMessage = null)
+        public async Task<PolicyResult<T>> RetryOnResult<T>(Func<Task<T>> func, Func<T, bool> resultFilter, string retryLogMessage = null)
```
timrogers, line 32: This is more flexible and allows you to define whatever filter you want, unlike the old way.

src/Octoshift/Handlers/DownloadLogsCommandHandler.cs
```diff
@@ -75,20 +73,17 @@ public async Task Handle(DownloadLogsCommandArgs args)
...
-
         if (result.Outcome == OutcomeType.Failure)
         {
             throw new OctoshiftCliException($"Migration log for repository {args.GithubRepo} unavailable!");
         }
 
-        var logUrl = result.Result;
+        (var logUrl, var migrationId) = result.Result;
+
+        args.MigrationLogFile ??= $"migration-log-{args.GithubOrg}-{args.GithubRepo}-{migrationId}.log";
```
ArinGhazarian, line 86: not sure if this is necessary because if we got to logging the log file, the `MigrationLogFile` will have a value otherwise we throw.

src/Octoshift/Handlers/DownloadLogsCommandHandler.cs
```diff
@@ -75,20 +73,17 @@ public async Task Handle(DownloadLogsCommandArgs args)
             _log.LogWarning($"Overwriting {args.MigrationLogFile} due to --overwrite option.");
         }
 
-        var result = await _retryPolicy.RetryOnResult(async () => await _githubApi.GetMigrationLogUrl(args.GithubOrg, args.GithubRepo), string.Empty,
+        var result = await _retryPolicy.RetryOnResult<(string MigrationLogURL, string MigrationID)>(async () => await _githubApi.GetMigrationLogUrl(args.GithubOrg, args.GithubRepo), result => string.IsNullOrEmpty(result.MigrationLogURL),
```
ArinGhazarian, line 76: I wouldn't change this retry logic.

ArinGhazarian (changes requested):
I think this PR should only make changes to add the migration id, later on we can create separate PRs for refactors if need be.

src/Octoshift/Services/GithubApi.cs
```diff
@@ -516,9 +517,15 @@ public virtual async Task<string> GetMigrationLogUrl(string org, string repo)
 
                 var nodes = (JArray)data["data"]["organization"]["repositoryMigrations"]["nodes"];
 
-                return nodes.Count == 0 ? null : (string)nodes[0]["migrationLogUrl"];
+                return nodes.Count == 0
+                    ? throw new OctoshiftCliException($"Migration for repository {repo} not found")
+                    : (MigrationLogURL: (string)nodes[0]["migrationLogUrl"], MigrationID: (string)nodes[0]["id"]);
```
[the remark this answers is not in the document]
timrogers: Just to confirm, do you think we should return in this case `(null, null)` rather than throwing an exception?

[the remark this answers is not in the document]
timrogers: I think this case actually means that there was no matching migration - not that there was no log URL. Am I wrong?

src/Octoshift/Handlers/DownloadLogsCommandHandler.cs
timrogers: The purpose of this line is to set the file name if the user didn't specify one when they called the command. Does that make sense?

src/Octoshift/Services/GithubApi.cs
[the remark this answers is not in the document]
ArinGhazarian: yes just like before, the return type can be a nullable value tuple so we can still return `null`.

src/Octoshift/Handlers/DownloadLogsCommandHandler.cs
timrogers: I think we have to make this change because the return value of the API method has changed. The filter has to adapt to that - unless I'm missing something.

src/Octoshift/Services/GithubApi.cs
```diff
@@ -516,9 +517,15 @@ public virtual async Task<string> GetMigrationLogUrl(string org, string repo)
...
             });
         }
+        catch (OctoshiftCliException)
+        {
+            throw;
+        }
```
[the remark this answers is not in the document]
timrogers: This was a mistake - thanks for flagging!

src/Octoshift/RetryPolicy.cs
timrogers: I was forced the change this by the type checker. It might be best to pair so we can work on this together!

src/Octoshift/Services/GithubApi.cs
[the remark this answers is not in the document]
timrogers: Ah, interesting! I didn't realise you could do that. That's better.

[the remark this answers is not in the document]
ArinGhazarian: > I think this case actually means that there was no matching migration - not that there was no log URL. Am I wrong?
I don't think so, in that case we return an `OctoshiftCliExcpetion`

src/Octoshift/Handlers/DownloadLogsCommandHandler.cs
ArinGhazarian: but how come it wasn't the case before?

src/Octoshift/Handlers/DownloadLogsCommandHandler.cs
ArinGhazarian: this retry is for when the log is yet unavailable and the method should still return `null` in that case so the retry logic  still applies. If we feel like we need to do refactors, it can be done in a separate PR.

src/Octoshift/Services/GithubApi.cs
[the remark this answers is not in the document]
timrogers: Before, or after my changes? (Just want to check!)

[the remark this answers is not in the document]
ArinGhazarian: before your changes --> https://github.com/github/gh-gei/blob/ce26a9be250f942ca7bbf9e24af70370274fcb09/src/Octoshift/GithubApi.cs#L519

src/Octoshift/Handlers/DownloadLogsCommandHandler.cs
timrogers: Okay! I understand this now - we can change `GetMigrationLogUrl` to continue to return `null`, and confusingly C#'s `string.Empty` matches empty strings _and_  nulls.

src/Octoshift/Handlers/DownloadLogsCommandHandler.cs
timrogers: This line was already here, but I've moved it down in the file so that we can include the migration ID in the filename. Am I missing something?

src/Octoshift/Handlers/DownloadLogsCommandHandler.cs
```diff
@@ -63,8 +63,6 @@ public async Task Handle(DownloadLogsCommandArgs args)
             _log.LogInformation($"MIGRATION LOG FILE: {args.MigrationLogFile}");
         }
 
-        args.MigrationLogFile ??= $"migration-log-{args.GithubOrg}-{args.GithubRepo}.log";
-
         if (FileExists(args.MigrationLogFile))
```
timrogers, line 66: We need to move this down as well, since we figure out the filename in a different place.

src/Octoshift/Services/GithubApi.cs
```diff
@@ -516,7 +517,7 @@ public virtual async Task<string> GetMigrationLogUrl(string org, string repo)
 
                 var nodes = (JArray)data["data"]["organization"]["repositoryMigrations"]["nodes"];
 
-                return nodes.Count == 0 ? null : (string)nodes[0]["migrationLogUrl"];
+                return nodes.Count == 0 ? null : (MigrationLogUrl: (string)nodes[0]["migrationLogUrl"], MigrationId: (string)nodes[0]["id"]);
```
timrogers, line 520: @ArinGhazarian Please ignore the rest of my changes - but I'm stuck right now because the compiler won't let me return `null` here. What am I doing wrong? 

```
/workspaces/gh-gei/src/Octoshift/Services/GithubApi.cs(520,17): error CS8031: Async lambda expression converted to a 'Task' returning delegate cannot return a value. Did you intend to return 'Task<T>'? [/workspaces/gh-gei/src/Octoshift/Octoshift.csproj]
```

I think it should be nullable given the `Task<(string MigrationLogUrl, string MigrationId)?>` return type.

src/Octoshift/Handlers/DownloadLogsCommandHandler.cs
```diff
@@ -75,24 +73,28 @@
             _log.LogWarning($"Overwriting {args.MigrationLogFile} due to --overwrite option.");
         }
 
-        var result = await _retryPolicy.RetryOnResult(async () => await _githubApi.GetMigrationLogUrl(args.GithubOrg, args.GithubRepo), string.Empty,
+        var result = await _retryPolicy.RetryOnResult<(string MigrationLogUrl, string MigrationId)?>(async () => await _githubApi.GetMigrationLogUrl(args.GithubOrg, args.GithubRepo), result => string.IsNullOrEmpty(result.Value.MigrationLogUrl),
```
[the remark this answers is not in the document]
timrogers: @ArinGhazarian What, if anything, should I do here?

timrogers:

src/Octoshift/Handlers/DownloadLogsCommandHandler.cs
```diff
@@ -88,11 +78,33 @@ public async Task Handle(DownloadLogsCommandArgs args)
...
+        // and migration ID.
+        CheckIfOutputFileAlreadyExists(args.MigrationLogFile, args.Overwrite);
 
         _log.LogInformation($"Downloading log for repository {args.GithubRepo} to {args.MigrationLogFile}...");
         await _httpDownloadService.DownloadToFile(logUrl, args.MigrationLogFile);
 
         _log.LogSuccess($"Downloaded {args.GithubRepo} log to {args.MigrationLogFile}.");
     }
+
+    private void CheckIfOutputFileAlreadyExists(string outputPath, bool shouldOverwrite)
```
ArinGhazarian, line 96: I wouldn't create this method (I've already added code snippets w/o using it). Also the if logic is wrong here an in case of an overwrite this is going to throw! the fact that no unit test is broken by changing the logic is a sign that we either don't have enough tests some tests are wrong. We def. need to add/fix tests for this. 
If you want to be DRY not to repeat the exception message you can create a method to throw the exception or maybe create one for building the error message, something like this:
```c#
throw new OctoshiftCliException(CreateFileAlreadyExistingErrorMessage(outputPath));
```

src/Octoshift/Services/GithubApi.cs
```diff
@@ -491,7 +491,7 @@ public virtual async Task<(string State, string RepositoryName, string FailureRe
         }
     }
 
-    public virtual async Task<string> GetMigrationLogUrl(string org, string repo)
+    public virtual async Task<(string MigrationLogUrl, string MigrationId)?> GetMigrationLogUrl(string org, string repo)
```
ArinGhazarian, line 494: As i already mentioned we can return a not nullable value tuple here since we are checking the `MigrationLogUrl` anyways.

src/Octoshift/Services/GithubApi.cs
```diff
@@ -516,7 +517,11 @@ public virtual async Task<string> GetMigrationLogUrl(string org, string repo)
 
                 var nodes = (JArray)data["data"]["organization"]["repositoryMigrations"]["nodes"];
 
-                return nodes.Count == 0 ? null : (string)nodes[0]["migrationLogUrl"];
+                return nodes.Count == 0
+                    // No matching migration was found
+                    ? ((string MigrationLogUrl, string MigrationId)?)null
+                    // A matching migration was found, which may or may not have a migration log URL. If there is no migration log, it's an empty string.
+                    : (MigrationLogUrl: (string)nodes[0]["migrationLogUrl"], MigrationId: (string)nodes[0]["id"]);
```
ArinGhazarian, line 524: When we change the return type to a not nullable value tuple this should change too:
```suggestion
                return nodes.Count == 0
                    ? default
                    : (MigrationLogUrl: (string)nodes[0]["migrationLogUrl"], MigrationId: (string)nodes[0]["id"]);
```

also the code is clear enough here so no need to add comments. Generally speaking code comments are a smell to me which means the code is not clear enough so it needs to be re-written with better variable/method naming or more clear logic.

ArinGhazarian (changes requested):

src/Octoshift/Services/GithubApi.cs
timrogers: Shouldn't it be nullable for the case so we can signal the case where no migration is found? 

I think there are three cases we need to handle here:

* Migration not found
* Migration found, but no log
* Migrations found with log

It's possible that I'm misunderstanding your comment, so apologies if so!

timrogers: We could change the logic so we return `(null, null)` in the "migration not found" case. That would be an option :)

timrogers: Ah, i think you're suggesting that in the comment below. Does the `default` keyword return `(null, null)`?

src/Octoshift/Handlers/DownloadLogsCommandHandler.cs
timrogers: This is the opposite of the intended logic 😂 Definitely needs better tests - need to check them to see what went wrong here.

src/Octoshift/Services/GithubApi.cs
ArinGhazarian: the default will return (null, null), I am also suggesting to not return a nullable value tuple.

ArinGhazarian: Oh my apologies, I totally missed the fact that we decided to return a nullable type to indicate that the migration is missing. So in that case there is no need to change the type but the `default` can def. state it will return the default value of `(string MigrationLogUrl, string MigrationId)?` which is `null` :)

src/Octoshift/Services/GithubApi.cs
timrogers: Resolving as we agreed not to make the tuple non-nullable

timrogers:
@ArinGhazarian Ready for review! In 6b2b50d82a5177e19987d8a9d3dea1289460d211, I fixed the "file exists" check logic and renamed the variable in the lambda.

src/Octoshift/Handlers/DownloadLogsCommandHandler.cs
```diff
@@ -63,19 +63,14 @@ public async Task Handle(DownloadLogsCommandArgs args)
...
-            {
-                throw new OctoshiftCliException($"File {args.MigrationLogFile} already exists!  Use --overwrite to overwrite this file.");
-            }
-
-            _log.LogWarning($"Overwriting {args.MigrationLogFile} due to --overwrite option.");
+            throw new OctoshiftCliException($"File {args.MigrationLogFile} already exists! Use --overwrite to overwrite this file.");
         }
 
-        var result = await _retryPolicy.RetryOnResult(async () => await _githubApi.GetMigrationLogUrl(args.GithubOrg, args.GithubRepo), string.Empty,
+        var result = await _retryPolicy.RetryOnResult<(string MigrationLogUrl, string MigrationId)?>(async () => await _githubApi.GetMigrationLogUrl(args.GithubOrg, args.GithubRepo), r => r?.MigrationLogUrl.IsNullOrWhiteSpace() ?? false,
```
dylan-smith, line 73:
```suggestion
        var result = await _retryPolicy.RetryOnResult(async () => await _githubApi.GetMigrationLogUrl(args.GithubOrg, args.GithubRepo), r => r?.MigrationLogUrl.IsNullOrWhiteSpace() ?? false,
```

the compiler can infer the generic type here

src/OctoshiftCLI.Tests/Octoshift/Services/GithubApiTests.cs
```diff
@@ -1436,7 +1439,56 @@ public async Task GetMigrationLogUrl_Returns_The_Migration_Log_URL()
         var expectedMigrationLog = await _githubApi.GetMigrationLogUrl(GITHUB_ORG, GITHUB_REPO);
 
         // Assert
-        expectedMigrationLog.Should().Be(migrationLogUrl);
+        expectedMigrationLog.Should().Be((migrationLogUrl, migrationId));
+    }
+
+    [Fact]
+    public async Task GetMigrationLogUrl_Returns_Empty_String_Migration_Log_URL_If_Returned_From_The_API()
```
ArinGhazarian, line 1446: This test seems unnecessary. It's just testing that it returning the empty log url value. Since an empty value is still an acceptable value it has already been covered via GetMigrationLogUrl_Returns_The_Migration_Log_URL_And_Migration_ID`

src/OctoshiftCLI.Tests/Octoshift/Handlers/DownloadLogsCommandHandlerTests.cs
ArinGhazarian, line 1: I would like to see at least two additional tests for the added logic:
1. A test to cover a scenario when --overwrite is not set and the migration log file exists.
2. when the default log filename after download is already existing.

ArinGhazarian (changes requested):

src/OctoshiftCLI.Tests/Octoshift/Handlers/DownloadLogsCommandHandlerTests.cs
timrogers: Tests added 👍🏻

src/OctoshiftCLI.Tests/Octoshift/Handlers/DownloadLogsCommandHandlerTests.cs
```diff
@@ -181,9 +264,8 @@ public async Task Throw_OctoshiftCliException_When_No_Migration()
         // Arrange
         const string githubOrg = "FooOrg";
         const string repo = "foo-repo";
-        const string logUrl = null;
 
-        _mockGithubApi.Setup(m => m.GetMigrationLogUrl(It.IsAny<string>(), It.IsAny<string>())).ReturnsAsync(logUrl);
+        //_mockGithubApi.Setup(m => m.GetMigrationLogUrl(It.IsAny<string>(), It.IsAny<string>())).ReturnsAsync(null);
```
ArinGhazarian, line 268: this line should be removed

ArinGhazarian (commented):
Looks good, just dropped a comment for deleting a commented line

src/OctoshiftCLI.Tests/Octoshift/Handlers/DownloadLogsCommandHandlerTests.cs
timrogers: Done!

ArinGhazarian (approved):

PR #963, opened 2023-04-19
Include the date with the timestamp when writing to the log

pnsk:
This PR will Include the date with the timestamp when writing to the log.

Currently, GEI CLI writes logs with only timestamp as follows: 

```
[8:42] [INFO] You are running the latest version of the gei CLI [v0.40]
[8:42] [INFO] Migrating Repo...
[8:42] [INFO] GITHUB SOURCE ORG: <source org>
[8:42] [INFO] SOURCE REPO: <source repo>
[8:42] [INFO] GITHUB TARGET ORG: <target org>
[8:42] [INFO] TARGET REPO: <target repo>
```

I see the `octoshift.log` includes the date in the filename but it would be good if the logs themselves include the date for troubleshooting.

Thanks!

- [x] Did you write/update appropriate tests -> No because it's small update. But I tested the logic worked on my local.
- [ ] Issue linked


timrogers (approved):
❤️ Thanks for your contribution! This seems like a sensible change and the code looks good to me. But I'll tag and engineer for a second look before we merge.

src/Octoshift/Services/OctoLogger.cs
```diff
@@ -11,7 +11,6 @@ internal static class LogLevel
     public const string INFO = "INFO";
     public const string WARNING = "WARNING";
     public const string ERROR = "ERROR";
-    public const string SUCCESS = "INFO";
```
ArinGhazarian, line 14: There is no need to delete this.

src/Octoshift/Services/OctoLogger.cs
```diff
@@ -149,7 +149,7 @@ public virtual void LogDebug(string msg)
     public virtual void LogSuccess(string msg)
     {
         Console.ForegroundColor = ConsoleColor.Green;
-        Log(msg, LogLevel.SUCCESS);
```
ArinGhazarian, line 152: No need to change this, it's un-relevant to this PR and an unnecessary change.

ArinGhazarian (changes requested):

src/Octoshift/Services/OctoLogger.cs
pnsk: @ArinGhazarian Thank you for reviewing this ❤️ 

It looks like that `SUCCESS` is [non-standard log level in C#](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loglevel?view=dotnet-plat-ext-7.0#fields), so I tried to standardize the log level.

The `SUCCESS` level is used only  in `LogSuccess` method, I feel like `INFO` is enough because of the method name. 

Could you please let me know why that non-standard log level was made here? 🙇 

Thanks!

dylan-smith: You're probably right that it's unnecessary. But we'd prefer to keep PR's focused on changing one thing per PR. A PR to change the timestamp format, and a separate PR to improve the log levels would be preferred.

ArinGhazarian: My main point was also not to mix arbitrary/unrelated refactors in PRs. But to answer your question, `OctoLogger` is a custom logger that we created for the cli, and those log levels are not following .NET's log levels. Actually success means something to us and we display it with a green text to indicate a success. IMO the success log level is needed, the only thing is we should probably change the constant value from `INFO` to `SUCCESS` but it shouldn't be done in this PR.

pnsk: @dylan-smith @ArinGhazarian thank you both for your kind response ❤️ 

> Actually success means something to us and we display it with a green text to indicate a success.

It does make sense 👍 Thank you so much for letting me know it 💖 

I'm going to remove the changes for `SUCCESS` level 👍

timrogers: Feel free to follow up with another PR to change the constant! ❤️ We really appreciate your contributions.

dylan-smith (approved):

ArinGhazarian (approved):
approved_1

Issue #702, opened 2022-10-26
Automatically log CLI args


ArinGhazarian:
Currently we log all of the args in the command handlers sometimes in a dedicated log method or sometimes right inside the handle method (or occasionally inside other methods too). 

There are some problems with this approach:

1. Logging (and also validation) should be part of the handler arg class itself and get called inside the command from within the `BuildHandler` method. This way we will be able to log the args early and also more importantly fail early (w/o having to create the handler) if the validations fail. 
2. Logging as it is right now is a manual process, so there's always room for human error. We can make the logging automatic with reflection. It can either be via an extension method or a base arg class. 
3. Right now we don't necessarily have a dedicated log method in all handlers, so logging is scattered here and there in the handler  which makes it easier to accidentally miss some args. 

## Todo
- [ ] Make logging automatic with reflection either via an extension method or a base arg class. 
- [ ] Remove the logging from the command handlers and do it inside the command as part of the `BuildHandler`  method.

PR #976, opened 2023-05-06
Refactor folder structure (take 2)

dylan-smith:
- [x] Issue linked

This PR does 3 things (with 3 separate commits for ease of review):
1. Restructures how we group the .cs files into folders. Instead of a Commands folder with all Command classes, and a Handlers folder with all handlers (how it used to be). Now it is one folder per Command, that contains both the Command class and the Handler class (and now the Args class also - see item 3 below). This just feels more natural, and puts the files that are usually worked on together, next to each other in the folder structure.

image

2. Updated the namespaces to match the new folder structure. The convention in .Net is to have namespaces and folders match.

3. Split out the CommandArgs classes into their own files. We used to have the Command class and the CommandArgs class in the same file, which is considered a bad practice in .Net world (each class should be in it's own file). The new folder structure makes this more reasonable to do. This will also be needed in preparation for the next refactor where we're going to move more code into the CommandArgs classes, so we'll definitely want them off in their own files.

This is in preparation for the refactors in these issues: #702 #703


synthead (approved):
I love it!  This is so much cleaner!  Thank you so much for these changes!

PR #977, opened 2023-05-06
Refactor: Move logging to base CommandArgs class

### LINKED ISSUE #702: Automatically log CLI args

dylan-smith:
- [x] Issue linked

Closes #702 and puts the foundation in place for #703 

Net deletion of 700 lines of code!! 🎉 

This PR does a few things:

1. Adds a new `[Secret]` custom attribute and decorates all Args class properties that are meant to be treated as a secret (i.e. not printed to the log).

2. Introduces a `CommandArgs` base class and updates all Args classes to inherit from it. This base class has code that will use reflection to log the values of all arguments, using the `[Secret]` attribute from above to make sure we don't print secrets.

3. Delete all code from command handlers that were logging the value of the args.

4. Update `CommandExtensions` (the place with the plumbing that actually executes the handlers) to first call `args.RegisterSecrets()`, `args.Log()`, and `args.Validate()` before invoking the handler. (Note: `args.Validate()` doesn't do anything yet, the next few PR's will change that).

5. The `CommandArgs` base class will also set the `Verbose` property on `OctoLogger` so we no longer need to remember to do that in every single handler.


src/ado2gh/Commands/GenerateScript/GenerateScriptCommandHandler.cs
```diff
@@ -414,66 +408,6 @@ private void AppendLine(StringBuilder sb, string content)
     private string Wrap(string script, string outerCommand = "") =>
         script.IsNullOrWhiteSpace() ? string.Empty : $"{outerCommand} {{ {script} }}".Trim();
 
-    private void LogOptions(GenerateScriptCommandArgs args)
-    {
-        _log.LogInformation($"GITHUB ORG: {args.GithubOrg}");
-        if (args.AdoOrg.HasValue())
-        {
-            _log.LogInformation($"ADO ORG: {args.AdoOrg}");
```
synthead, line 422: I _love_ seeing these lines getting abstracted :smile:

synthead (approved):
Looks great!  This cleanup is awesome!

I have one nit about a line starting on column 129, but this is certainly not blocking.  Feel free to :ship: if this nit doesn't bug you :smile:

PR #980, opened 2023-05-10
Display number of migration log warnings encountered after a migration completes

synthead:
Closes https://github.com/github/octoshift/issues/6948!

- [ ] Issue linked

### DIFF
--- a/src/Octoshift/Services/GithubApi.cs
+++ b/src/Octoshift/Services/GithubApi.cs
@@ -463,12 +463,26 @@ public class GithubApi
         );
     }
 
-    public virtual async Task<(string State, string RepositoryName, string FailureReason, string MigrationLogUrl)> GetMigration(string migrationId)
+    public virtual async Task<(string State, string RepositoryName, int WarningsCount, string FailureReason, string MigrationLogUrl)> GetMigration(string migrationId)
     {
         var url = $"{_apiUrl}/graphql";
 
         var query = "query($id: ID!)";
-        var gql = "node(id: $id) { ... on Migration { id, sourceUrl, migrationLogUrl, migrationSource { name }, state, failureReason, repositoryName } }";
+        var gql = @"
+                node(id: $id) {
+                    ... on Migration {
+                        id,
+                        sourceUrl,
+                        migrationLogUrl,
+                        migrationSource {
+                            name
+                        },
+                        state,
+                        warningsCount,
+                        failureReason,
+                        repositoryName
+                    }
+                }";
 
         var payload = new { query = $"{query} {{ {gql} }}", variables = new { id = migrationId } };
 
@@ -481,6 +495,7 @@ public class GithubApi
                 return (
                     State: (string)data["data"]["node"]["state"],
                     RepositoryName: (string)data["data"]["node"]["repositoryName"],
+                    WarningsCount: (int)data["data"]["node"]["warningsCount"],
                     FailureReason: (string)data["data"]["node"]["failureReason"],
                     MigrationLogUrl: (string)data["data"]["node"]["migrationLogUrl"]);
             });
--- a/src/Octoshift/Handlers/WaitForMigrationCommandHandler.cs
+++ b/src/Octoshift/Handlers/WaitForMigrationCommandHandler.cs
@@ -95,7 +95,7 @@ public class WaitForMigrationCommandHandler : ICommandHandler<WaitForMigrationCo
 
     private async Task WaitForRepositoryMigration(string migrationId, string githubPat, GithubApi githubApi)
     {
-        var (state, repositoryName, failureReason, migrationLogUrl) = await githubApi.GetMigration(migrationId);
+        var (state, repositoryName, warningsCount, failureReason, migrationLogUrl) = await githubApi.GetMigration(migrationId);
 
         _log.LogInformation($"Waiting for {repositoryName} migration (ID: {migrationId}) to finish...");
 
@@ -124,7 +124,7 @@ public class WaitForMigrationCommandHandler : ICommandHandler<WaitForMigrationCo
             _log.LogInformation($"Waiting {WaitIntervalInSeconds} seconds...");
             await Task.Delay(WaitIntervalInSeconds * 1000);
 
-            (state, repositoryName, failureReason, migrationLogUrl) = await githubApi.GetMigration(migrationId);
+            (state, repositoryName, warningsCount, failureReason, migrationLogUrl) = await githubApi.GetMigration(migrationId);
         }
     }
 }
--- a/src/OctoshiftCLI.Tests/Octoshift/Handlers/WaitForMigrationCommandHandlerTests.cs
+++ b/src/OctoshiftCLI.Tests/Octoshift/Handlers/WaitForMigrationCommandHandlerTests.cs
@@ -35,9 +35,9 @@ public class WaitForMigrationCommandHandlerTests
     {
         // Arrange
         _mockGithubApi.SetupSequence(x => x.GetMigration(REPO_MIGRATION_ID).Result)
-            .Returns((State: RepositoryMigrationStatus.InProgress, RepositoryName: TARGET_REPO, FailureReason: null, MigrationLogUrl: MIGRATION_URL))
-            .Returns((State: RepositoryMigrationStatus.InProgress, RepositoryName: TARGET_REPO, FailureReason: null, MigrationLogUrl: MIGRATION_URL))
-            .Returns((State: RepositoryMigrationStatus.Succeeded, RepositoryName: TARGET_REPO, FailureReason: null, MigrationLogUrl: MIGRATION_URL));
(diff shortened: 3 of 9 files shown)


src/Octoshift/Commands/WaitForMigration/WaitForMigrationCommandHandler.cs
```diff
@@ -84,13 +85,15 @@ private async Task WaitForRepositoryMigration(string migrationId, GithubApi gith
             if (RepositoryMigrationStatus.IsSucceeded(state))
             {
                 _log.LogSuccess($"Migration {migrationId} succeeded for {repositoryName}");
+                _warningsCountLogger.LogWarningsCount(warningsCount);
```
timrogers, line 88: I quite like this pattern - it allows us to test the logging without setting expectations on the error message itself!

src/Octoshift/Services/GithubApi.cs
```diff
@@ -463,12 +463,26 @@ public virtual async Task<string> StartBbsMigration(string migrationSourceId, st
...
     }
 
-    public virtual async Task<(string State, string RepositoryName, string FailureReason, string MigrationLogUrl)> GetMigration(string migrationId)
+    public virtual async Task<(string State, string RepositoryName, int WarningsCount, string FailureReason, string MigrationLogUrl)> GetMigration(string migrationId)
     {
         var url = $"{_apiUrl}/graphql";
 
         var query = "query($id: ID!)";
-        var gql = "node(id: $id) { ... on Migration { id, sourceUrl, migrationLogUrl, migrationSource { name }, state, failureReason, repositoryName } }";
+        var gql = @"
```
timrogers, line 471: Yey for formatting 🧹

src/Octoshift/Services/WarningsCountLogger.cs
```diff
@@ -0,0 +1,27 @@
...
+
+    public void LogWarningsCount(int warningsCount)
+    {
+        switch (warningsCount)
+        {
+            case 0:
+                _log.LogInformation("No warnings encountered during this migration");
+                break;
+            case 1:
+                _log.LogWarning("1 warning encountered during this migration");
```
timrogers, line 20: I wonder if we should point to the Migration Log explicitly here. Maybe `LogWarningsCount` could also be passed the log URL, and we could print that in the same log message?

src/Octoshift/Services/WarningsCountLogger.cs
timrogers, line 1: Should we have a unit test for this?

timrogers (commented):
I haven't reviewed all of the code in detail - so don't take this as a code review - but here's some PM comments!

src/Octoshift/Commands/WaitForMigration/WaitForMigrationCommandHandler.cs
dylan-smith, line 88: If you have code you want to reuse across multiple command handlers, then putting it into a service and injecting it like this is the right way to do it.

My main piece of feedback is this feels like a really tiny/trivial thing to build a whole service around. If it was me I'd probably ditch the service and just do this in the handlers:

```suggestion
                _log.LogWarning($"{warningsCount} warning(s) encountered during this migration");
```

dylan-smith (approved):

src/Octoshift/Commands/WaitForMigration/WaitForMigrationCommandHandler.cs
dylan-smith: but if you do choose to leave it as a service, there should probably be a unit test or two for the new service.

src/OctoshiftCLI.Tests/gei/Commands/MigrateRepo/MigrateRepoCommandHandlerTests.cs
```diff
@@ -48,6 +49,8 @@ public class MigrateRepoCommandHandlerTests
         public MigrateRepoCommandHandlerTests()
         {
             _retryPolicy = new RetryPolicy(_mockOctoLogger.Object) { _httpRetryInterval = 1, _retryInterval = 0 };
+            _warningsCountLogger = new WarningsCountLogger(_mockOctoLogger.Object);
```
dylan-smith, line 52: Usually we would mock the WarningsCountLogger itself in the Handler tests.

src/Octoshift/Commands/WaitForMigration/WaitForMigrationCommandHandler.cs
synthead: When I started working on this, I started with a private method in a handler class, and then discovered that there are four places it needed to be implemented.  I'm also using `LogInformation` when there are no warnings, and `LogWarning` when warnings are present.

It still feels like a :pinching_hand: change, and it does feel like kind of a lot for a service class.  We could move this to `OctoLogger` instead.  I decided not to put it in this class because I liked the decoupling of the logging class and the logging content, but this isn't a bad place to put it, either.

This gives us three reasonable places this could be placed:

- In four private methods as identical code
- A service class that contains common code
- A public method in OctoLogger

Any strong feelings about this?

src/Octoshift/Services/WarningsCountLogger.cs
synthead: I thought about this, and decided to use some higher-level tests instead.  I'm happy to add them, though!

Let's see where https://github.com/github/gh-gei/pull/980#discussion_r1200898571 goes, and if we keep the service class, I'll add tests for this class :+1:

src/Octoshift/Services/WarningsCountLogger.cs
synthead: The CLI writes this line after writing to the console about the warnings:

```
[2023-05-16 14:26:25] [INFO] Migration log available at  or by running `gh bbs2gh download-logs --github-target-org import-testing --target-repo ***2gh-ccf221ea-64e7-432c-8bfc-03bc195251aa`
```

It looks like there was a regression in the log output at some point, because the log URL is supposed to be present here:

https://github.com/github/gh-gei/blob/f99944f9caab330a493da25af99f2ffe3e3cf3c1/src/gei/Commands/MigrateRepo/MigrateRepoCommandHandler.cs#L166-L166

If this :point_up: was working and the URL is being written to the output, will this satisfy what you're looking for?

timrogers: If that worked, it would most definitely meet my needs and expectations 🙏🏻

src/Octoshift/Commands/WaitForMigration/WaitForMigrationCommandHandler.cs
timrogers: I don't have strong feelings on this.

dylan-smith: IMO adding something like this into OctoLogger is the worst of the three options. OctoLogger code is all at a pretty consistent level of "abstraction". It's all general-purpose logging code, with nothing in that class specific to octoshift (you could copy-paste OctoLogger into any .Net CLI project).

Another possible option that I'm not sure how I feel about could be adding it as an extension method to OctoLogger (extension methods are kind of .Net version of mixins). This way it's separate from the OctoLogger class to avoid the above concern. But adding extension methods to a class we wrote is a little weird, but maybe this is a good justification of it.

gohana: > Another possible option that I'm not sure how I feel about could be adding it as an extension method to OctoLogger (extension methods are kind of .Net version of mixins)

I'll be wrapping up this PR. To come back to this suggestion.. any updated thoughts on this? @dylan-smith @synthead 

Otherwise will stick with the current approach.

timrogers: Coming back to this, I do personally like the current approach - but unit tests would be great.

src/Octoshift/Services/WarningsCountLogger.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        _log = logger;
+    }
+
+    public void LogWarningsCount(int warningsCount)
+    {
+        switch (warningsCount)
+        {
+            case 0:
+                _log.LogInformation("No warnings encountered during this migration");
```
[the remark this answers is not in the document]
timrogers: That's a really good call-out. I do think we could remove this.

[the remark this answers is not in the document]
gohana: So if no warnings, don't log anything?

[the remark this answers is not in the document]
gohana: @hfishback01 @timrogers done in [bba1c1a](https://github.com/github/gh-gei/pull/980/commits/bba1c1a8046ee7a7fd581ae3122a2c9a65cff4dd)

PR #1012, opened 2023-06-01
Remove "not used" params in BBS migrations

synthead:
Bitbucket Server migrations are archive-only and BBS instances are assumed to be unreachable from the internet.  Therefore, Octoshift will never reach out to source instances, and a source PAT is not required.  Additionally, BBS migration archives are only single archives that we move using the `gitArchiveUrl` GraphQL field for `startMigration` mutations.  Therefore, we never use `metadataArchiveUrl` field.

We have updated the public GraphQL interface to make these two parameters optional.  Prior to this, we added filler "not used" values for `accessToken` and `metadataArchiveUrl` to satisfy the GraphQL request requirements.  These ignored "filler" values are not necessary anymore, so this PR removes them! :sparkles: 

- [ ] Issue linked


begonaguereca:
Im going to close this down since its been a year and a half since there's been movement 👍🏽 Please feel free to reopen otherwise!

PR #1024, opened 2023-06-14
When starting a repository migration, output the migration's GUID to the logs, as well as its GraphQL ID (`RM_`)

timrogers:
Right now, when starting a migration, we only print the GraphQL ID (beginning `RM_` to the logs). 

This also outputs the GUID (aka database ID), which is referred to in certain places (e.g. the Migration Log issue and text file).

- [ ] Issue linked


dylan-smith:
FYI, depending on how visible you want the UUID to be when using the CLI there is a more lightweight way to accomplish this. If you just included the change the the GQL query, but didn't change `GithubApi` return value, or any of the `LogInformation` lines, it would still accomplish the goal of getting the UUID in the log. Because all HTTP request/responses are logged in the verbose log. So as long as the GQL endpoint is returning the UUID, it will be in the verbose log file even if we do nothing with it in the code.

timrogers:
👍🏻 I think this approach is probably better, just because it doesn't rely on your being in verbose mode.

timrogers:
Closing in favour of #1032.

This solution was perhaps in some ways prettier - it leads to nicer log output - but it's probably overkill, and it was breaking our `generate-script` scripts which rely on parsing the output to get the migration ID.

PR #1029, opened 2023-06-18
Log a warning when `bbs2gh`'s `--ssh-port` is set to the port usually used for Git operations

timrogers:
This updates `bbs2gh migrate-repo` and `bbs2gh generate-script` to log a warning when `--ssh port` is set to `7999`.

By default, Bitbucket Server and Bitbucket Data Center [use][1] this port for Git operations over SSH, so it's likely that users will think this is the correct port to use. But in fact, you need to use the SSH port used to administer the server where Bitbucket is running, not this special port which is only used for Git-over- SSH.

Adding a warning will help to point out that users have probably made a mistake, and point them in the right direction!

[1]:
https://confluence.atlassian.com/bitbucketserver/setting-up-ssh-port-forwarding-776640364.html#:~:text=Bitbucket%C2%A0listens%20for%20SSH%20connections%20on%20port%207999%20by%20default.

- [ ] Issue linked

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -1,1 +1,1 @@
-
+- Warn when the `--ssh-port` argument for `bbs2gh migrate-repo` and `bbs2gh generate-script` is set to the default port for Git operations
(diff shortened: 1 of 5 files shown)


hfishback01 (approved):
LGTM! 🎉

Issue #1033, opened 2023-06-21
After GHES to GHEC migration the outputted text shows an incorrect extension name to download log


gitstua:
## Description

I completed a migration from GHES 3.7.4 to GHEC EMU and then tried to download the logs with the suggested command. This fails with the error `unknown command "gh-gei" for "gh"`

The correct command is `gh gei download-logs ...`

Mac OS 13.4 (22F66)
gei CLI [v0.45]

## Reproduction Steps
- complete single repo migration from GHES to GHEC EMU (I used GHEC 3.7.4)
- once completed the migration suggests running  downloading logs ```by running `gh gh-gei download-logs --github-target-org FOO --target-repo repo1```
-  an invalid command error is shown `unknown command "gh-gei" for "gh"`

timrogers:
![image](https://github.com/github/gh-gei/assets/116134/69f4f876-2672-4be4-96e7-5bf9e9535dce)

PR #1034, opened 2023-06-21
Fix recommended follow-up commands outputted to the logs

timrogers:
In a number of places in the CLI, we recommend follow-up `gh [gei|bbs2gh|ado2gh]` commands to run.

When the CLI is run as a `gh` extension rather than standalone, the examples commands we output in the logs are incorrect, repeating `gh` twice (e.g. `gh gh-gei`).

This fixes it so we produce accurate, usable commands that you can actually run.

I chose to fix this in `CliContext.RootCommand` (as [suggested][1] by @dylan-smith). It would also have been possible to fix this in the `Program` classes that set `CliContext.RootCommand`, but that would have required duplicating logic in 3 places.

Fixes https://github.com/github/gh-gei/issues/1033.

- [x] Issue linked

### DIFF
--- a/src/OctoshiftCLI.Tests/Octoshift/CliContextTests.cs
+++ b/src/OctoshiftCLI.Tests/Octoshift/CliContextTests.cs
@@ -6,7 +6,7 @@ namespace OctoshiftCLI.Tests.Octoshift;
 public class CliContextTests
 {
     [Fact]
-    public void SetRootCommandShouldRemoveGhPrefix()
+    public void Set_RootCommand_Should_Remove_Gh_Prefix()
     {
         CliContext.Clear();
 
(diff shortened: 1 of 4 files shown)


timrogers:
I couldn't find a natural place to test this and it doesn't break any current tests - but I'd be open to suggestions.

dylan-smith:
I would add a unit test for CliContext

timrogers:
@dylan-smith Done!

src/OctoshiftCLI.Tests/Octoshift/CliContextTests.cs
```diff
@@ -0,0 +1,29 @@
+using FluentAssertions;
+using Xunit;
+
+namespace OctoshiftCLI.Tests.Octoshift;
+
+public class CliContextTests
+{
+    [Fact]
+    public void SetRootCommandShouldRemoveGhPrefix()
```
dylan-smith, line 9: Looks good, only suggestion is to use snake casing to be consistent with other tests
```suggestion
    public void Set_RootCommand_Should_Remove_Gh_Prefix()
```

timrogers:
@dylan-smith I'd welcome your final review if you have a sec ❤️

timrogers:
We're going to rework this so that we just set the `RootCommand` statically from the `Program` classes. Closing.

PR #940, opened 2023-04-04
Rollback setup-dotnet action

dylan-smith:
- [x] Did you write/update appropriate tests
- [x] Release notes updated (if appropriate)
- [x] Appropriate logging output
- [ ] Issue linked
- [x] Docs updated (or issue created)
- [x] New package licenses are added to `ThirdPartyNotices.txt` (if applicable)

When we upgraded the setup-dotnet action from v1 to v3 yesterday it caused our builds to get significantly longer (3.5min -> 8.5 min for one job).

I narrowed it down to something specifically in v3.0.0 of that action. I will open an issue for the team that owns that action to hopefully help us shed some more light on what's going on, in the meantime this PR rolls us back to v2.

PR #1043, opened 2023-06-26
Revert "Bump actions/setup-dotnet from 1 to 3"

timrogers:
Reverts github/gh-gei#937

We're reverting this because we've seen `dotnet format` sporadically failing since the upgrade.


dylan-smith:
I think you rolled back the wrong PR.

There was PR #937 , then that was rolled back in PR #940 

Then we did PR #949 which is the one we should be rolling back now.

dylan-smith:
The main difference being that `v2` was working. Also not sure if that globals.json file introduced in #949 is part of the problem or not.

PR #1048, opened 2023-06-26
Fix recommended follow-up commands outputted to the logs by statically setting the root command for each CLI

timrogers:
In a number of places in the CLI, we recommend follow-up `gh [gei|bbs2gh|ado2gh]` commands to run.

When the CLI is run as a `gh` extension rather than standalone, the examples commands we output in the logs are incorrect, repeating `gh` twice (e.g.  `gh gh-gei`).

This fixes it so we produce accurate, usable commands that you can actually run.

We do this by hard-coding the "root command" as a static string in each `Program` class, rather than trying to deduce it.

Fixes https://github.com/github/gh-gei/issues/1033.

- [ ] Issue linked

### DIFF
--- a/src/ado2gh/Commands/MigrateRepo/MigrateRepoCommandHandler.cs
+++ b/src/ado2gh/Commands/MigrateRepo/MigrateRepoCommandHandler.cs
@@ -83,12 +83,12 @@ public class MigrateRepoCommandHandler : ICommandHandler<MigrateRepoCommandArgs>
         if (RepositoryMigrationStatus.IsFailed(migrationState))
         {
             _log.LogError($"Migration Failed. Migration ID: {migrationId}");
-            _log.LogInformation($"Migration log available at {migrationLogUrl} or by running `gh {CliContext.RootCommand} download-logs --github-target-org {args.GithubOrg} --target-repo {args.GithubRepo}`");
+            _log.LogInformation($"Migration log available at {migrationLogUrl} or by running `gh ado2gh download-logs --github-target-org {args.GithubOrg} --target-repo {args.GithubRepo}`");
             throw new OctoshiftCliException(failureReason);
         }
 
         _log.LogSuccess($"Migration completed (ID: {migrationId})! State: {migrationState}");
-        _log.LogInformation($"Migration log available at {migrationLogUrl} or by running `gh {CliContext.RootCommand} download-logs --github-target-org {args.GithubOrg} --target-repo {args.GithubRepo}`");
+        _log.LogInformation($"Migration log available at {migrationLogUrl} or by running `gh ado2gh download-logs --github-target-org {args.GithubOrg} --target-repo {args.GithubRepo}`");
     }
 
     private string GetAdoRepoUrl(string org, string project, string repo, string serverUrl)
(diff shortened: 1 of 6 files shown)


src/ado2gh/Commands/MigrateRepo/MigrateRepoCommandHandler.cs
```diff
@@ -83,12 +83,12 @@ public async Task Handle(MigrateRepoCommandArgs args)
         if (RepositoryMigrationStatus.IsFailed(migrationState))
         {
             _log.LogError($"Migration Failed. Migration ID: {migrationId}");
-            _log.LogInformation($"Migration log available at {migrationLogUrl} or by running `gh {CliContext.RootCommand} download-logs --github-target-org {args.GithubOrg} --target-repo {args.GithubRepo}`");
+            _log.LogInformation($"Migration log available at {migrationLogUrl} or by running `gh ado2gh download-logs --github-target-org {args.GithubOrg} --target-repo {args.GithubRepo}`");
```
synthead, line 86: Perhaps we could continue to use `CliContext.RootCommand` like this?

```suggestion
            _log.LogInformation($"Migration log available at {migrationLogUrl} or by running `{CliContext.RootCommand} download-logs --github-target-org {args.GithubOrg} --target-repo {args.GithubRepo}`");
```

If you dig, let's do this for bbs2gh and gei :+1:

synthead (commented):
This is so much nicer!  Thanks for opening this PR!

synthead:
It also looks like there are some tests that are still passing, even though `CliContext.RootCommand` should be similar to `"gh gei"`.  In the tests, this seems to be expected to be `"gh gh gei"` and should be failing here, but it isn't:

https://github.com/github/gh-gei/blob/3c327f8e6c2e143d5d78ccc537c639df37b65479/src/OctoshiftCLI.Tests/Octoshift/Commands/WaitForMigration/WaitForMigrationCommandHandlerTests.cs#L46-L56

Could you look into this, also?

src/ado2gh/Commands/MigrateRepo/MigrateRepoCommandHandler.cs
timrogers: I didn't mean to change this 😓 I meant to just update tests - serves me right for working on this just before going to bed!

timrogers: Reverted in 92eab21

timrogers:
> It also looks like there are some tests that are still passing, even though `CliContext.RootCommand` should be similar to `"gh gei"`. In the tests, this seems to be expected to be `"gh gh gei"` and should be failing here, but it isn't:
> 
> https://github.com/github/gh-gei/blob/3c327f8e6c2e143d5d78ccc537c639df37b65479/src/OctoshiftCLI.Tests/Octoshift/Commands/WaitForMigration/WaitForMigrationCommandHandlerTests.cs#L46-L56
> 
> Could you look into this, also?

We should definitely fix this! ❤️ I'll take a look.

src/bbs2gh/Program.cs
```diff
@@ -96,7 +96,7 @@ public static async Task Main(string[] args)
 
         private static void SetContext(InvocationContext context)
         {
-            CliContext.RootCommand = context.ParseResult.RootCommandResult.Command.Name;
+            CliContext.RootCommand = "gh bbs2gh";
```
timrogers, line 99:
```suggestion
            CliContext.RootCommand = "bbs2gh";
```

src/gei/Program.cs
```diff
@@ -89,7 +89,7 @@ public static async Task Main(string[] args)
 
         private static void SetContext(ParseResult parseResult)
         {
-            CliContext.RootCommand = parseResult.RootCommandResult.Command.Name;
+            CliContext.RootCommand = "gh gei";
```
timrogers, line 92:
```suggestion
            CliContext.RootCommand = "gei";
```

hfishback01 (approved):

Issue #954, opened 2023-04-13
`TaskCanceledException` when `bbs2gh migrate-repo` uploads migration archive to Amazon S3


timrogers:
A customer has reported getting an intermittent `TaskCanceledException` exception when running a Bitbucket Server migration with `gh bbs2gh migrate-repo`:

```
12:39 PM] [ERROR] System.Threading.Tasks.TaskCanceledException: A task was canceled.
   at Amazon.Runtime.Internal.RetryHandler.InvokeAsync[T](IExecutionContext executionContext)
   at Amazon.Runtime.Internal.CallbackHandler.InvokeAsync[T](IExecutionContext executionContext)
   at Amazon.Runtime.Internal.CallbackHandler.InvokeAsync[T](IExecutionContext executionContext)
   at Amazon.S3.Internal.AmazonS3ExceptionHandler.InvokeAsync[T](IExecutionContext executionContext)
   at Amazon.Runtime.Internal.ErrorCallbackHandler.InvokeAsync[T](IExecutionContext executionContext)
   at Amazon.Runtime.Internal.MetricsHandler.InvokeAsync[T](IExecutionContext executionContext)
   at Amazon.S3.Transfer.Internal.MultipartUploadCommand.UploadPartAsync(UploadPartRequest uploadRequest, CancellationTokenSource internalCts, SemaphoreSlim asyncThrottler)
   at Amazon.S3.Transfer.Internal.BaseCommand.WhenAllOrFirstExceptionAsync[T](List`1 pendingTasks, CancellationToken cancellationToken)
   at Amazon.S3.Transfer.Internal.MultipartUploadCommand.ExecuteAsync(CancellationToken cancellationToken)
   at OctoshiftCLI.Services.AwsApi.UploadToBucket(String bucketName, String fileName, String keyName)
   at OctoshiftCLI.BbsToGithub.Handlers.MigrateRepoCommandHandler.UploadArchiveToAws(String bucketName, String archivePath)
   at OctoshiftCLI.BbsToGithub.Handlers.MigrateRepoCommandHandler.Handle(MigrateRepoCommandArgs args)
   at OctoshiftCLI.Extensions.CommandExtensions.<>c__DisplayClass1_0`3.<<ConfigureCommand>b__0>d.MoveNext()
--- End of stack trace from previous location ---
   at System.CommandLine.Invocation.AnonymousCommandHandler.InvokeAsync(InvocationContext )
[...]

PR #1050, opened 2023-06-27
Configure timeouts on archive file upload to AWS

gohana:
Closes https://github.com/github/gh-gei/issues/954

On AWS uploads, [the Timeout value is 100 seconds by default](https://docs.aws.amazon.com/sdk-for-net/v3/developer-guide/retries-timeouts.html). This PR increases the timeout to 300 seconds, which should help with multi-part uploads of large archives.

The AWS SDK’s default retry mode is Legacy, which has a default maximum of 4 retries. Not changing any retry behavior at this time, but this is easily adjustable in the AmazonS3Config.

- [ ] Issue linked


gohana:
# Testing

## Timeout

I temporarily set the Timeout to 1ms:
https://github.com/github/gh-gei/blob/a2fe9e561d9bf5ff68cfb4d2f64183558ffff24f/src/Octoshift/Services/AwsApi.cs#L43-L47

Started a BBS migration:
```
src/bbs2gh/bin/Debug/net6.0/bbs2gh migrate-repo \
  --archive-path largefile.tar \
  --aws-bucket-name github-dev \
  --github-org import-testing \
  --github-repo aws-timeout-test
```

The upload timed out with a `TimeoutException`:

timeout

## ReadWriteTimeout

I temporarily set the cancellation token timeout to 1ms:
https://github.com/github/gh-gei/blob/a2fe9e561d9bf5ff68cfb4d2f64183558ffff24f/src/Octoshift/Services/AwsApi.cs#L60-L62

Started another migration, and the upload was cancelled with a `TaskCanceledException`:

cancelled

src/Octoshift/Services/AwsApi.cs
```diff
@@ -34,9 +36,17 @@ private static AmazonS3Client BuildAmazonS3Client(string awsAccessKeyId, string
...
-            ? new AmazonS3Client(awsAccessKeyId, awsSecretAccessKey, awsSessionToken, regionEndpoint)
-            : new AmazonS3Client(awsAccessKeyId, awsSecretAccessKey, regionEndpoint);
+        var creds = awsSessionToken.HasValue()
+            ? (AWSCredentials)new SessionAWSCredentials(awsAccessKeyId, awsSecretAccessKey, awsSessionToken)
+            : new BasicAWSCredentials(awsAccessKeyId, awsSecretAccessKey);
+
+        var config = new AmazonS3Config
+        {
+            RegionEndpoint = regionEndpoint,
+            Timeout = TimeSpan.FromMinutes(5)
```
dylan-smith, line 46: How does this timeout relate to the CancellationToken timeout? is the timeout going to be the 5 mins from here or the 10 mins from the CTS?

src/Octoshift/Services/AwsApi.cs
```diff
@@ -45,7 +55,23 @@ private static AmazonS3Client BuildAmazonS3Client(string awsAccessKeyId, string
...
+
+            await _transferUtility.UploadAsync(fileName, bucketName, keyName, cts.Token);
+        }
+        catch (TaskCanceledException ex)
+        {
+            var message = ex.CancellationToken.IsCancellationRequested ? "cancelled" : "timed out";
+
+            throw new OctoshiftCliException($"Upload of archive \"{fileName}\" to AWS {message}", ex);
+        }
+        catch (TimeoutException ex)
```
dylan-smith, line 70: If we hit the 10 (?) min timeout, do we hit a TaskCanceled Exception or a TimeoutException?

gohana: The 10 min timeout is on the Cancellation Token and would be a TaskCanceledException.

dylan-smith: So do we need to handle both exceptions if it's always going to be a TaskCanceledException. I'm struggling to understand when it would be one vs the other (also struggling to understand why we need to set the timeout in 2 places, and are using different values for each).

src/Octoshift/Services/AwsApi.cs
gohana: I believe the Cancellation Token timeout will apply to the entire upload operation, and in the case of a large file, the Timeout on the AmazonS3Config will apply to the sub-requests of the multi-part upload.

src/Octoshift/Services/AwsApi.cs
gohana: When I tested by setting the Timeout on the S3 client to 1 ms, UploadAsync threw a TimeoutException wrapping a TaskCanceledException:

cancelled

That's why I'm also catching the TimeoutException.

dylan-smith:
I suggest you add a bullet point to RELEASENOTES.md for this change

gohana:
@dylan-smith Release note added in [b03f8c4](https://github.com/github/gh-gei/pull/1050/commits/b03f8c4601d49c00f77065e052cdb35a55a4455e)

src/Octoshift/Services/AwsApi.cs
```diff
@@ -34,9 +35,17 @@ private static AmazonS3Client BuildAmazonS3Client(string awsAccessKeyId, string
...
-            ? new AmazonS3Client(awsAccessKeyId, awsSecretAccessKey, awsSessionToken, regionEndpoint)
-            : new AmazonS3Client(awsAccessKeyId, awsSecretAccessKey, regionEndpoint);
+        var creds = awsSessionToken.HasValue()
+            ? (AWSCredentials)new SessionAWSCredentials(awsAccessKeyId, awsSecretAccessKey, awsSessionToken)
+            : new BasicAWSCredentials(awsAccessKeyId, awsSecretAccessKey);
+
+        var config = new AmazonS3Config
+        {
+            RegionEndpoint = regionEndpoint,
+            Timeout = TimeSpan.FromMinutes(5)
```
timrogers, line 45: Is there any specific thinking behind the 5 minutes number? Should we maybe go for something longer so we can support really large archives? An archive could be up to 10GB, and it's likely that 10GB will take longer than 5 minutes to upload.

```suggestion
            Timeout = TimeSpan.FromMinutes(30)
```

src/Octoshift/Services/AwsApi.cs
```diff
@@ -45,7 +54,15 @@ private static AmazonS3Client BuildAmazonS3Client(string awsAccessKeyId, string
...
     public virtual async Task<string> UploadToBucket(string bucketName, string fileName, string keyName)
     {
-        await _transferUtility.UploadAsync(fileName, bucketName, keyName);
+        try
+        {
+            await _transferUtility.UploadAsync(fileName, bucketName, keyName);
+        }
+        catch (Exception ex) when (ex is TaskCanceledException or TimeoutException)
+        {
+            throw new OctoshiftCliException($"Upload of archive \"{fileName}\" to AWS timed out", ex);
```
timrogers, line 63: I like that we catch this and return a specific clear error message ❤️ 

Could we add the timeout into this error message so it's clear exactly what happened, e.g. "Upload...to AWS timed out after 5 minutes"?

src/Octoshift/Services/AwsApi.cs
timrogers: We could consider having a configurable timeout, but this seems like a good approach for now.

timrogers (changes requested):

src/Octoshift/Services/AwsApi.cs
timrogers: ...or is this 5 minutes per part of a multipart upload? If that's the case, then this might be fine?

gohana: > Is there any specific thinking behind the 5 minutes number?

The default is 100 and I just tripled it.

> ...or is this 5 minutes per part of a multipart upload?

Yes it's for the multi-part uploads 👍

timrogers: The timeout is per part, and the default size of each part is 16MB ([docs](https://docs.aws.amazon.com/mobile/sdkforxamarin/developerguide/s3-integration-transferutility.html#:~:text=value%20is%2010.-,MinSizeBeforePartUpload,-%2D%20Gets%20or%20sets)), so 5 minutes seems like a good number to me. Let's stick with this.

timrogers (approved):

src/Octoshift/Services/AwsApi.cs
gohana: Not sure about the "after 5 minutes" wording. I think it could timeout for reasons unrelated to the 5 min timeout set in this PR, like user network connection.

timrogers (approved):

PR #1054, opened 2023-07-05
Drop support for non-standard, deprecated `AWS_ACCESS_KEY` and `AWS_SECRET_KEY` environment variables

### LINKED ISSUE #856: Remove the obsolete AWS_ACCESS_KEY and AWS_SECRET_KEY environment variables
As part of #855 we introduced new AWS environment variables that are more aligned with the AWS CLI ones but we kept the support for old variables for backward compatibility. Once we make sure that no one is using the old ones we can drop the old variables and remove them from the code. 

## Todo
- [ ] Remove `AWS_ACCESS_KEY` and `AWS_SECRET_KEY` environment variables.

timrogers:
…ECRET_KEY` environment variables

Up until #855, we used non-standard environment variables for configuring AWS S3 credentials. With #855, we *added* support for the standard `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` variables, but maintained support for the old names, printing a warning when they're used.

We've had that behaviour for 4 months now, and it's time to cut the old behaviour.

With this change, only the new environment variable names will be accepted, and trying to use the old names will result in your argument not being recognised.

Closes #856.

- [ ] Issue linked


src/OctoshiftCLI.Tests/bbs2gh/Commands/MigrateRepo/MigrateRepoCommandHandlerTests.cs
```diff
@@ -740,42 +738,6 @@ public async Task It_Throws_When_Aws_Bucket_Name_Is_Provided_But_But_No_Aws_Acce
                 .WithMessage("*--aws-access-key*AWS_ACCESS_KEY_ID*");
         }
 
-        [Fact]
-        public async Task It_Does_Not_Throw_When_Aws_Bucket_Name_Is_Provided_And_Can_Fallback_To_Aws_Access_Key_Environment_Variable()
```
timrogers, line 744: This fallback behaviour is no longer here, so it doesn't need a test

timrogers, line 744: This fallback behaviour is no longer here, so it doesn't need a test

timrogers:
I just tested this locally, and I'm satisfied that it does what we're expecting 🏁

robandpdx:
@timrogers Thank you for the graceful breaking change process here! I really appreciate that you took my feedback seriously. This saves us from angry and confused customers!!!

src/bbs2gh/Factories/AwsApiFactory.cs
```diff
@@ -13,14 +13,8 @@ public AwsApiFactory(EnvironmentVariableProvider environmentVariableProvider)
...
-#pragma warning disable CS0618
-        awsAccessKeyId ??= _environmentVariableProvider.AwsAccessKeyId(false)
-                           ?? _environmentVariableProvider.AwsAccessKey(false)
-                           ?? _environmentVariableProvider.AwsAccessKeyId();
-        awsSecretAccessKey ??= _environmentVariableProvider.AwsSecretAccessKey(false)
-                               ?? _environmentVariableProvider.AwsSecretKey(false)
-                               ?? _environmentVariableProvider.AwsSecretAccessKey();
-#pragma warning restore CS0618
+        awsAccessKeyId ??= _environmentVariableProvider.AwsAccessKeyId(false);
+        awsSecretAccessKey ??= _environmentVariableProvider.AwsSecretAccessKey(false);
```
dylan-smith, line 17:
```suggestion
        awsAccessKeyId ??= _environmentVariableProvider.AwsAccessKeyId();
        awsSecretAccessKey ??= _environmentVariableProvider.AwsSecretAccessKey();
```

The true/false determines whether it throws an exception if  the env var is not set. The previous behavior was:
- Check if the new env var is set but don't throw if it isn't
- Check if the old env var is set but don't throw if it isn't
- Check if the new env var is set (again), but this time throw if it isn't

src/gei/Factories/AwsApiFactory.cs
```diff
@@ -13,14 +13,8 @@ public AwsApiFactory(EnvironmentVariableProvider environmentVariableProvider)
...
-#pragma warning disable CS0618
-        awsAccessKeyId ??= _environmentVariableProvider.AwsAccessKeyId(false)
-                           ?? _environmentVariableProvider.AwsAccessKey(false)
-                           ?? _environmentVariableProvider.AwsAccessKeyId();
-        awsSecretAccessKey ??= _environmentVariableProvider.AwsSecretAccessKey(false)
-                               ?? _environmentVariableProvider.AwsSecretKey(false)
-                               ?? _environmentVariableProvider.AwsSecretAccessKey();
-#pragma warning restore CS0618
+        awsAccessKeyId ??= _environmentVariableProvider.AwsAccessKeyId(false);
+        awsSecretAccessKey ??= _environmentVariableProvider.AwsSecretAccessKey(false);
```
dylan-smith, line 17:
```suggestion
        awsAccessKeyId ??= _environmentVariableProvider.AwsAccessKeyId();
        awsSecretAccessKey ??= _environmentVariableProvider.AwsSecretAccessKey();
```

RELEASENOTES.md
```diff
@@ -1 +1,2 @@
 
+- Drop support for deprecated `AWS_ACCESS_KEY` and `AWS_SECRET_KEY` environment variables in `gh gei` and `gh bbs2gh`. The AWS S3 credentials can now be configured the `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` variables or command line arguments.
```
dylan-smith, line 2:
```suggestion
- Drop support for deprecated `AWS_ACCESS_KEY` and `AWS_SECRET_KEY` environment variables in `gh gei` and `gh bbs2gh`. The AWS S3 credentials can now be configured using the `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` variables or command line arguments.
```

src/bbs2gh/Factories/AwsApiFactory.cs
timrogers: Thanks for flagging! Would this mean that we don't need the validation  code in the `MigrateRepoCommandHandler`? I'm a bit confused about how the two fit together.

timrogers: It looks like this code will run before the validation in the command handler, since the `AwsApi` is passed in when the handler is instantiated.

timrogers: I did a manual test so I'm surprised this is wrong - I guess either my test was totally wrong, or there's a nuanced behaviour difference.

dylan-smith: This code does run before the handler code, but I think the validation is still needed in the handler for some scenarios.

dylan-smith: if the secrets are provided via CLI args instead of env vars, the validation in the factory doesn't kick in at all, and validation happens in the handler (and/or args class).

timrogers: Makes sense - I can see how that works in this class. 

I'm happy to make this change, but do you have any idea what the scenario is where we need the validation in the handler?

I'm wondering if we need to keep it at all or if we can simplify.

dylan-smith: Which specific bit of validation do you think is unnecessary?

timrogers: I'm specifically thinking of this:

https://github.com/github/gh-gei/blob/main/src/bbs2gh/Commands/MigrateRepo/MigrateRepoCommandHandler.cs#L318-L344

I don't think this will be necessary, because `AwsApi` will error if the access key ID and secret access key aren't configured through either the CLI arguments or an environment variable, and this method is called by `ValidateOptions` which is called by `Handle` - and that happens after the handler is instantiated with its own `AwsApi` instance.

dylan-smith: Yup I think you're right, those validations can never be hit.

timrogers: I think I'll choose to tidy that up in a separate PR, just to avoid polluting this changeset.

brianaj (approved):

Issue #857, opened 2023-03-04
Make the `--aws-region` arg required


ArinGhazarian:
As part of adding support for AWS session token auth, we also added the `--aws-region` arg that is currently required to set if `--aws-session-token` is set. Right now for backward compatibility it's not required on its own but if it's not set, we spit out a warning to let the users know that they need to pass it as it will become required in future. Once we make sure that it's safe to make it required, we should do so and add the necessary changes in our code.  

## Todo
- [ ] Make the `--aws-region` arg required.

timrogers:
Fixed in https://github.com/github/gh-gei/pull/1055

PR #1055, opened 2023-07-06
Require the AWS region to be specified if using AWS S3 for blob storage

timrogers:
A few months ago, we changed the CLI to allow an AWS region to be specified when using AWS S3 for blob storage in `gh gei` and `gh bbs2gh`.

However, at the moment, the region is optional - unless you're using an AWS session token specified with `--aws-session-token` or `AWS_SESSION_TOKEN`, in which case the region is mandatory. If you don't specify it, we allow you to continue, but print a warning saying that providing a region is recommended and will soon be mandatory.

This updates the behaviour, completing our breaking change, making it mandatory to specify the session token.

- [ ] Issue linked


src/Octoshift/Services/AwsApi.cs
```diff
@@ -27,12 +26,8 @@ public AwsApi(string awsAccessKeyId, string awsSecretAccessKey, string awsRegion
 
     private static AmazonS3Client BuildAmazonS3Client(string awsAccessKeyId, string awsSecretAccessKey, string awsRegion, string awsSessionToken)
     {
-        var regionEndpoint = DefaultRegionEndpoint;
-        if (awsRegion.HasValue())
-        {
-            regionEndpoint = GetRegionEndpoint(awsRegion);
-            AWSConfigsS3.UseSignatureVersion4 = true;
-        }
+        var regionEndpoint = awsRegion.IsNullOrWhiteSpace() ? null : GetRegionEndpoint(awsRegion);
```
timrogers, line 29: I'd love to cut this out and set the expectation that we always have a region here. During runtime, we always do, because we validate that in `AwsApiFactory`, but it seems that there are some cases in our tests where the `AwsApi` is initialized without going through the factory.

src/bbs2gh/Commands/MigrateRepo/MigrateRepoCommandHandler.cs
```diff
@@ -325,16 +325,9 @@ private void ValidateUploadOptions(MigrateRepoCommandArgs args)
                 throw new OctoshiftCliException("Either --aws-secret-key or AWS_SECRET_ACCESS_KEY environment variable must be set.");
             }
 
-            if (GetAwsSessionToken(args).HasValue() && GetAwsRegion(args).IsNullOrWhiteSpace())
+            if (GetAwsRegion(args).IsNullOrWhiteSpace())
```
timrogers, line 328: It doesn't seem like we should have to have this validation here because `AwsApiFactory` will validate that the region is set when it is initialized. But in the tests, there are some cases which skip that path and this validation gets run.

timrogers:
Fixes https://github.com/github/gh-gei/issues/857.

timrogers:
The tests pass for me locally, but we get failures like this in CI:

```
[xUnit.net 00:00:03.13]     OctoshiftCLI.Tests.GithubEnterpriseImporter.Commands.MigrateRepo.MigrateRepoCommandHandlerTests.It_Uses_Aws_If_Arguments_Are_Included [FAIL]
  Failed OctoshiftCLI.Tests.GithubEnterpriseImporter.Commands.MigrateRepo.MigrateRepoCommandHandlerTests.It_Uses_Aws_If_Arguments_Are_Included [970 ms]
  Error Message:
   Amazon.Runtime.AmazonClientException : No RegionEndpoint or ServiceURL configured
  Stack Trace:
     at Amazon.Runtime.ClientConfig.Validate()
   at Amazon.S3.AmazonS3Config.Validate()
   at Amazon.Runtime.AmazonServiceClient..ctor(AWSCredentials credentials, ClientConfig config)
   at Amazon.S3.AmazonS3Client..ctor(AWSCredentials credentials, AmazonS3Config clientConfig)
   at OctoshiftCLI.Services.AwsApi.BuildAmazonS3Client(String awsAccessKeyId, String awsSecretAccessKey, String awsRegion, String awsSessionToken) in D:\a\gh-gei\gh-gei\src\Octoshift\Services\AwsApi.cs:line 43
   at OctoshiftCLI.Services.AwsApi..ctor(String awsAccessKeyId, String awsSecretAccessKey, String awsRegion, String awsSessionToken) in D:\a\gh-gei\gh-gei\src\Octoshift\Services\AwsApi.cs:line 21
   at Castle.Proxies.AwsApiProxy..ctor(IInterceptor[] , String awsAccessKeyId, String awsSecretAccessKey, String awsRegion, String awsSessionToken)
```

I'm not sure how to debug that.

dylan-smith:
> The tests pass for me locally, but we get failures like this in CI:
> 
> ```
> [xUnit.net 00:00:03.13]     OctoshiftCLI.Tests.GithubEnterpriseImporter.Commands.MigrateRepo.MigrateRepoCommandHandlerTests.It_Uses_Aws_If_Arguments_Are_Included [FAIL]
>   Failed OctoshiftCLI.Tests.GithubEnterpriseImporter.Commands.MigrateRepo.MigrateRepoCommandHandlerTests.It_Uses_Aws_If_Arguments_Are_Included [970 ms]
>   Error Message:
>    Amazon.Runtime.AmazonClientException : No RegionEndpoint or ServiceURL configured
>   Stack Trace:
>      at Amazon.Runtime.ClientConfig.Validate()
>    at Amazon.S3.AmazonS3Config.Validate()
>    at Amazon.Runtime.AmazonServiceClient..ctor(AWSCredentials credentials, ClientConfig config)
>    at Amazon.S3.AmazonS3Client..ctor(AWSCredentials credentials, AmazonS3Config clientConfig)
>    at OctoshiftCLI.Services.AwsApi.BuildAmazonS3Client(String awsAccessKeyId, String awsSecretAccessKey, String awsRegion, String awsSessionToken) in D:\a\gh-gei\gh-gei\src\Octoshift\Services\AwsApi.cs:line 43
>    at OctoshiftCLI.Services.AwsApi..ctor(String awsAccessKeyId, String awsSecretAccessKey, String awsRegion, String awsSessionToken) in D:\a\gh-gei\gh-gei\src\Octoshift\Services\AwsApi.cs:line 21
>    at Castle.Proxies.AwsApiProxy..ctor(IInterceptor[] , String awsAccessKeyId, String awsSecretAccessKey, String awsRegion, String awsSessionToken)
> ```
> 
> I'm not sure how to debug that.

When I run it in a codespace I see the same failing tests. You might not be seeing it locally because you have some AWS env vars set that the CI doesn't?

image

timrogers:
That's a great point. I definitely do. I'll give it a go on a Codespace on Monday. Another reason why clean, separated environments are a good idea...!

src/Octoshift/Services/AwsApi.cs
dylan-smith: the tests (in the process of mocking AwsAPI) will call the constructor with all null args. This is a consequence of the mocking pattern we chose to use, where the mock is technically a subclass of the thing it's mocking. So the constructors need to support being called with all null values (and ideally do a noop in that scenario).

dylan-smith: I have a workaround but it's a bit sketchy. If you make the 2nd AwsApi ctor public (it's currently internal), then move it so it appears first in the file (because our `TestHelper.CreateMock` uses the first ctor it finds - which we should probably change to use the ctor with the least args) all the tests will pass. This is because the other ctor works fine if you pass it null args (it basically does a noop).

timrogers:
That's a great point. I definitely do. I'll give it a go on a Codespace on Monday. Another reason why clean, separated environments are a good idea...!

src/Octoshift/Services/AwsApi.cs
timrogers: Thanks for the heads-up! I've made that change to get this unblocked.

synthead (approved):
LGTM!

Issue #1056, opened 2023-07-07
`bbs2gh migrate-repo` migrates wrong repo in fork hierarchy when using the "split" flow where generation and import is separate


timrogers:
## Description

When running a Bitbucket Server migration, we send the Bitbucket Server URL of the repo being migrated in the `startRepositoryMigration` mutation.

We need this because the Bitbucket Server API often generates archives containing multiple repos (e.g. for fork hierarchies), and we need a way to figure out which repo the user actually wanted to migrate.

This works well for the "end to end" flow where a single `migrate-repo` command run is used to generate, download, upload and import the archive.

At the moment, if you "split" the flow - calling `migrate-repo` once to generate your archive and then calling it a second time to upload and import it - this flow doesn't work.

There is no way of calling the CLI to allow the repo URL to be sent to the server, which means that the wrong repo may be selected and migrated from your archive. 

In order for the proper URL to be generated (rather than using `http://not-used`), you have to specify a `--bbs-server-url`:

https://github.com/github/gh-gei/blob/main/src/bbs2gh/Commands/MigrateRepo/MigrateRepoCommandHandler.cs#L260-L265

...and specifying a `--bbs-server-url` with an `--archive-path` is not permitted:

```bash
$ gh bbs2gh migrate-repo --bbs-server-url https://REDACTED \
    --bbs-project OS \
    --bbs-repo litra \
    --archive-path Bitbucket_export_57.tar
[2023-07-07 12:13:30] [WARNING] You are running an older version of the bbs2gh extension [v0.46.0.0]. The latest version is v0.47.
[2023-07-07 12:13:30] [WARNING] Please update by running: gh extension upgrade bbs2gh
[2023-07-07 12:13:30] [INFO] BBS SERVER URL: https://REDACTED.expert-services.io/
[2023-07-07 12:13:30] [INFO] BBS PROJECT: OS
[2023-07-07 12:13:30] [INFO] BBS REPO: litra
[2023-07-07 12:13:30] [INFO] ARCHIVE PATH: Bitbucket_export_57.tar
[2023-07-07 12:13:30] [INFO] SSH PORT: 22
[2023-07-07 12:13:30] [ERROR] Only one of --bbs-server-url or --archive-path can be specified.
```

### Migrating the wrong repo

1. Create a Bitbucket Server repo, and then fork it
2. Generate a BBS archive for the forked repo using the CLI:

```
gh bbs2gh migrate-repo --bbs-server-url https://REDACTED \
[...]

PR #1057, opened 2023-07-08
Require the Bitbucket Server URL, project and repo name to always be passed for `bbs2gh migrate-repo`

timrogers:
This makes a **breaking change** to `bbs2gh migrate-repo` requiring the `--bbs-server-url`, `--bbs-project` and `--bbs-repo` to always be passed, even in the flows where the CLI doesn't talk to the Bitbucket Server API.

This change allows us to always construct the URL of the repository and then send it to the GitHub API when starting the migration. 

That's important, because the URL allows us to intelligently pick the right repo to migrate from the archive, rather than picking the "first" repo in the fork hierarchy (which may be wrong!).

I did consider introducing a warning first and then making a breaking change, but I thought going straight to a breaking change was the right decision because:

* This is currently in private beta, so I can easily contact affected users
* We're getting ready to move to public beta very soon, and I want this to be "right" before the public beta
* The impact of this change is limited, given that most users use the default end-to-end flow

Fixes https://github.com/github/gh-gei/issues/1056.

- [x] Issue linked

### DIFF
--- a/src/OctoshiftCLI.Tests/bbs2gh/Commands/MigrateRepo/MigrateRepoCommandArgsTests.cs
+++ b/src/OctoshiftCLI.Tests/bbs2gh/Commands/MigrateRepo/MigrateRepoCommandArgsTests.cs
@@ -284,44 +284,6 @@ namespace OctoshiftCLI.Tests.BbsToGithub.Commands.MigrateRepo
         }
 
         [Fact]
-        public void Errors_If_BbsServer_Url_Not_Provided_But_Bbs_Project_Is_Provided()
-        {
-            // Act
-            var args = new MigrateRepoCommandArgs
-            {
-                ArchivePath = ARCHIVE_PATH,
-                GithubOrg = GITHUB_ORG,
-                GithubRepo = GITHUB_REPO,
-                BbsProject = BBS_PROJECT
-            };
-
-            // Assert
-            args.Invoking(x => x.Validate(_mockOctoLogger.Object))
-                .Should()
-                .ThrowExactly<OctoshiftCliException>()
-                .WithMessage("*--bbs-project*--bbs-server-url*");
-        }
-
-        [Fact]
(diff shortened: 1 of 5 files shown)


src/OctoshiftCLI.Tests/bbs2gh/Commands/MigrateRepo/MigrateRepoCommandArgsTests.cs
```diff
@@ -283,44 +283,6 @@ public void Errors_If_BbsServer_Url_Not_Provided_But_Smb_User_Is_Provided()
                 .WithMessage("*SSH*SMB*--bbs-server-url*");
         }
 
-        [Fact]
```
timrogers, line 286: We now validate in the `Args` class that all 3 arguments are always provided

timrogers:
I've tested all three flows (end-to-end + upload-and-import + import-only) after this change, and they all behave as expected.

timrogers:
Docs changes in https://github.com/github/docs-internal/pull/38838

dylan-smith:
Did you consider doing a 2-phase implementation (like we have done for other breaking changes). Where we first make those arguments optional when using `--archive-path` or `--archive-url` and spit out a warning if users don't provide the optional arguments.

Then at some point in the future come back and make them required.

timrogers:
> Did you consider doing a 2-phase implementation (like we have done for other breaking changes). Where we first make those arguments optional when using `--archive-path` or `--archive-url` and spit out a warning if users don't provide the optional arguments.
> 
> 
> 
> Then at some point in the future come back and make them required.

I did consider it - but I feel like a more aggressive approach is fair game during a limited private beta that is heading to public beta in the near future. 

It's easy for me to contact affected users, so I thought a more immediate change made sense to fix what is a pretty nasty bug. 

But I will think more on this over the weekend! Thanks for raising it.

dpmex4527 (approved):
LGTM!

dpmex4527 (approved):

PR #1067, opened 2023-07-17
Improve performance of single mannequin reclaims by looking up the mannequin by its `login`

timrogers:
This PR improves the performance of the `reclaim-mannequin` command when used in "single reclaim" (non-CSV) mode.

Right now, to find the right mannequin(s), we paginate through all of the organization's mannequins, 100 at a time. For organizations with many mannequins, this can take a long time and use up a lot of rate limit points.

We recently introduced the ability to filter mannequins by `login` in the GraphQL API. This means that we can immediately look up the mannequins we actually care about, rather than going through all of them, searching for a needle in a haystack.

In a pathological case (e.g. 10000 mannequins, and the one you want happens to be on the last page), this can make the operation 100x faster.

- [ ] Issue linked

### DIFF
--- a/src/Octoshift/Services/GithubApi.cs
+++ b/src/Octoshift/Services/GithubApi.cs
@@ -722,6 +722,25 @@ public class GithubApi
         }
     }
 
+    public virtual async Task<IEnumerable<Mannequin>> GetMannequinsByLogin(string orgId, string login)
+    {
+        var url = $"{_apiUrl}/graphql";
+
+        var payload = GetMannequinsByLoginPayload(orgId, login);
+
+        return await _retryPolicy.Retry(async () =>
+        {
+            return await _client.PostGraphQLWithPaginationAsync(
+                url,
+                payload,
+                data => (JArray)data["data"]["node"]["mannequins"]["nodes"],
+                data => (JObject)data["data"]["node"]["mannequins"]["pageInfo"])
+            .Select(mannequin => BuildMannequin(mannequin))
+            .ToListAsync();
+        });
+    }
+
+
     public virtual async Task<string> GetUserId(string login)
     {
         var url = $"{_apiUrl}/graphql";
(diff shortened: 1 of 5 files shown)


timrogers:
Manual testing complete ✅

src/Octoshift/Services/GithubApi.cs
```diff
@@ -722,6 +722,25 @@ public virtual async Task<IEnumerable<Mannequin>> GetMannequins(string orgId)
         }
     }
 
+    public virtual async Task<IEnumerable<Mannequin>> GetMannequinsByLogin(string orgId, string login)
```
hfishback01, line 725: Nitpick: if we're just getting one mannequin, should we change the name to `GetMannequinByLogin`?

src/Octoshift/Services/GithubApi.cs
```diff
@@ -1003,6 +1022,36 @@ private static object GetMannequinsPayload(string orgId)
         };
     }
 
+    private static object GetMannequinsByLoginPayload(string orgId, string login)
```
hfishback01, line 1025: Same nitpick as above, should this be `GetMannequinByLoginPayload`?

src/Octoshift/Services/GithubApi.cs
```diff
@@ -722,6 +722,25 @@ public virtual async Task<IEnumerable<Mannequin>> GetMannequins(string orgId)
...
+
+        return await _retryPolicy.Retry(async () =>
+        {
+            return await _client.PostGraphQLWithPaginationAsync(
+                url,
+                payload,
+                data => (JArray)data["data"]["node"]["mannequins"]["nodes"],
+                data => (JObject)data["data"]["node"]["mannequins"]["pageInfo"])
+            .Select(mannequin => BuildMannequin(mannequin))
+            .ToListAsync();
```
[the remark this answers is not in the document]
timrogers: There can be multiple mannequins with the same login, but different mannequin IDs. That's something that we already handle in the code - if multiple are returned, [we loop through them](https://github.com/github/gh-gei/blob/main/src/Octoshift/Services/ReclaimService.cs#L106) and reclaim each one.

src/Octoshift/Services/GithubApi.cs
timrogers: I think the plural is right - see my [other comment](https://github.com/github/gh-gei/pull/1067#discussion_r1268309196).

src/Octoshift/Services/GithubApi.cs
timrogers: I think the plural is right - see my [other comment](https://github.com/github/gh-gei/pull/1067#discussion_r1268309196).

Issue #1074, opened 2023-07-20
Support migrating to repository already renamed


kmaehashi:
## Description

The migration cannot be started even if the repository does not exist, if there was a repository with that name.

## Reproduction Steps

1. Create a repository `MYORG/MYREPO`
2. Rename the repository, e.g., `MYORG/MYREPO_OLD`
3. Start migration with destination `MYORG/MYREPO`

```
[2023-07-20 09:11:59] [DEBUG] HTTP GET: https://api.github.com/repos/MYORG/MYREPO
[2023-07-20 09:11:59] [DEBUG] GITHUB REQUEST ID: <snip>
[2023-07-20 09:11:59] [DEBUG] RESPONSE (Moved): {"message":"Moved Permanently","URL":"<snip>","documentation_url":"https://docs.github.com/v3/#http-redirects"}
[2023-07-20 09:11:59] [DEBUG] [HTTP ERROR 301] System.Net.Http.HttpRequestException: Expected status code NotFound but got Moved
```

timrogers:
@kmaehashi Thanks for flagging this! It definitely sounds like a bug 😅 You might not know this - which is totally okay - but do you know if this is a new bug or an old one? (Maybe you tried to migrate this repo and it used to work!)

kmaehashi:
Not sure if this is a new one. One of our users tried to migrate twice (with the previous one renamed) and I've got this reported 😅

dylan-smith:
This is a bug with how we do the validation check if the target repo already exists. This validation check only happens when migrating from GHES.

The way we do the check is we try to retrieve the target repo name, if we get a HTTP 200 then the repo exists and we fail the migration. We expect a HTTP 404 if the repo does not exist. If we get anything other than a 200/404 the CLI will error out like above.

For whoever picks this up the relevant code is here: https://github.com/github/gh-gei/blob/main/src/Octoshift/Services/GithubApi.cs#L149

timrogers:
Fix in progress in https://github.com/github/gh-gei/pull/1084

PR #1084, opened 2023-07-25
Fix target repo existence check in GHES migrations so it doesn't explode if the target repo used to exist, but has been renamed

timrogers:
When migrating from GHES with `migrate-repo`, we perform a quick validation check before starting the export to ensure that the target repo doesn't already exist. If it does already exist, we can fail fast and not run the export at all.

If the repo used to exist but has been renamed, then this causes the CLI to error, because the REST API returns `301 Moved Permanently` and that's not something we handle or expect.

This updates our code to handle `301 Moved Permanently` and treat it correctly as "the repo doesn't exist".

Fixes https://github.com/github/gh-gei/issues/1074.

- [x] Issue linked

### DIFF
--- a/src/Octoshift/Services/GithubApi.cs
+++ b/src/Octoshift/Services/GithubApi.cs
@@ -158,6 +158,10 @@ public class GithubApi
         {
             return true;
         }
+        catch (HttpRequestException ex) when (ex.StatusCode == HttpStatusCode.MovedPermanently)
+        {
+            return false;
+        }
     }
 
     public virtual async Task<bool> DoesOrgExist(string org)
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -1,1 +1,1 @@
-
+- Fix target repo existence check in GitHub Enterprise Server migrations so it doesn't error if the target repo used to exist, but has been renamed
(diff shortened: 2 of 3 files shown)


src/Octoshift/Services/GithubApi.cs
```diff
@@ -158,6 +158,10 @@ public virtual async Task<bool> DoesRepoExist(string org, string repo)
         {
             return true;
         }
+        catch (HttpRequestException ex) when (ex.StatusCode == HttpStatusCode.MovedPermanently)
```
timrogers, line 161: An alternative approach would have been to update `GetNonSuccessAsync` (see lines above) to allow multiple "allowed HTTP statuses". But that turned out to be reasonably complex, and I'm not sure how likely it is that we'll need this again, so I took the easy way out.

timrogers:
I considered covering this in the tests for `OctoshiftCLI.GithubEnterpriseImporter.Commands.MigrateRepo.MigrateRepoCommandHandler` too, but I decided not to as this is a very small change and I trust the unit tests. But I'm open to changing my approach here.

hfishback01 (approved):
LGTM! :tada:

I'm fine looking specifically for the MovedPermanently as opposed to getting a NonSuccess.

PR #1137, opened 2023-10-17
Update version check logic

pnsk:
[`VersionChecker`](https://github.com/github/gh-gei/blob/main/src/Octoshift/Services/VersionChecker.cs#L22-L30) only checks if the latest version equals to the version of GEI CLI a user is using .i.e. `curVersion == latestVersion`:

https://github.com/github/gh-gei/blob/baf18dc45e7264ca644bc2fd659372b09ac744a8/src/Octoshift/Services/VersionChecker.cs#L22-L30

This PR will change the logic to check if the current version is older than the latest version.

- [ ] Issue linked


timrogers (approved):
This changes makes sense to me - but I'll ask one of our engineers to take a look too ✅ 

I would expect to see tests updated for this change, but from what I can see, this behaviour isn't tested right now, so that wouldn't be a super easy thing to do.

src/Octoshift/Services/VersionChecker.cs
```diff
@@ -26,7 +26,7 @@ public async Task<bool> IsLatest()
 
         curVersion = curVersion[..latestVersion.Length];
 
-        return curVersion == latestVersion;
+        return latestVersion.CompareTo(curVersion) <= 0;
```
timrogers, line 29: I do wonder if we should use a semantic versioning library here to properly compare the versions - that would be the safest option - but in practice, I don't think it's necessary.

timrogers (changes requested):
@pnsk Can you add an entry to `RELEASENOTES.md` like this?:

> Fix version check so that a version later than the latest version published on GitHub isn't considered old

pnsk:
@timrogers I've updated `RELEASENOTES.md` 😄 

Could you please review this PR again? 🙇 

Thanks!

timrogers (approved):

PR #1139, opened 2023-10-17
Revert "Update LATEST-VERSION.txt to reflect current version"

timrogers:
Reverts github/gh-gei#1138

I'm reverting this so that we can instead run a normal release to fix the `LATEST-VERSION.txt` file.


timrogers:
Merging this before the build finishes, since it only updates the `LATEST-VERSION.txt` file, and not any of the code used for the build process

PR #1149, opened 2023-10-20
Add debug section to Contributing.md

begonaguereca:
- [ ] Issue linked

Very rough, this has been my experience so far! Please fell free to fill in alternative methods @timrogers @dylan-smith


dylan-smith:
I resisted writing docs like this in the past because it feels like generic "how to debug a dotnet project" guidance that I hope exists elsewhere at a more authoritative source that we could link to.

begonaguereca:
> I resisted writing docs like this in the past because it feels like generic "how to debug a dotnet project" guidance that I hope exists elsewhere at a more authoritative source that we could link to.

If there is a more authoritative document please let me know! We can pop them in. I think facilitating this for devs would make it one step easier to start contributing IMO

ArinGhazarian (changes requested):
I also don't think the extra docs are necessary since it's basic knowledge, here are the official docs regarding running and debugging .NET apps:
1. Run a .NET app (if the app is executable): https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-run
2. Debugging using vscode: https://code.visualstudio.com/docs/editor/debugging

Also note that for different IDEs it's a bit different to pass the command line args but it's generally easier than vscode.

begonaguereca:
> 2\. https://code.visualstudio.com/docs/editor/debugging

Cool! How would folks feel about linking these docs then?

ArinGhazarian:
IMO the `dotnet` CLI is the bare minimum ppl should be familiar with and probably is the very first thing. For example `dotnet run`, `dotnet test`, `dotnet build`, etc. So I don't think that we should document them. 
Regarding how to pass command args, it may be useful to add a link.

Those being said, maybe we can add a `Useful links` section to the end of the doc and add link for both [.NET CLI docs](https://learn.microsoft.com/en-us/dotnet/core/tools/) and the [vscode debugging doc](https://code.visualstudio.com/docs/editor/debugging)

begonaguereca:
@ArinGhazarian I really like your suggestion of adding a helpful links section! I also think what I wrote out is something that could be helpful (perhaps written in a more robust/elegant way?). 

My vision for writing this is enabling newer engineers, even level 1 engineers who have little to no C# experience to get a command going and be able to set some breakpoints to explore the codebase 😄

brianaj:
> @ArinGhazarian I really like your suggestion of adding a helpful links section! I also think what I wrote out is something that could be helpful (perhaps written in a more robust/elegant way?).
> 
> My vision for writing this is enabling newer engineers, even level 1 engineers who have little to no C# experience to get a command going and be able to set some breakpoints to explore the codebase 😄

I see value in both having a quick command and how to run examples and also should have useful links reference. I wonder why we are so anti-docs in this repo? It shouldn't be so much pushback for a change that will be useful to some (and the others like yourselves who are familiar can just ignore). I definitely recall having to explicitly ask how to run things when I did some dev work here and would have found these quick references useful.

ArinGhazarian (approved):

CONTRIBUTING.md
```diff
@@ -20,6 +20,36 @@ Use the [Discussions](https://github.com/github/gh-gei/discussions) tab in this
...
+
+## Debugging code 
+
+There are two ways to debug code within VS code. 
+
+### Run GH-GHEI command locally
+Run the following commands in your terminal depending on the provider you are looking to run the commands for. 
+
+Azure DevOps: 
+`dotnet run --project src/ado2gh/ado2gh.csproj -- <<INSERT COMMAND HERE>>`
```
synthead, line 32: In most CLI documentation (i.e. man pages), straight brackets explicitly document optional parameters.

For example, here is `man ls`:

```
SYNOPSIS
       ls [OPTION]... [FILE]...
```

And here's `man top`:

```
SYNOPSIS
       top [options]
```

It would be more conformist to also use straight brackets here:

```suggestion
`dotnet run --project src/ado2gh/ado2gh.csproj -- [command]`
```

If you dig, let's also use this with the BBS and GitHub commands, too :+1:

CONTRIBUTING.md
```diff
@@ -20,6 +20,36 @@ Use the [Discussions](https://github.com/github/gh-gei/discussions) tab in this
...
+`dotnet run --project src/gei/gei.csproj -- <<INSERT COMMAND HERE>>`
+
+### Use debugger
+
+If you use the built in debugger you are able to set breakpoints and inspect the code within VS Code. 
+
+1. Navigate to `.vs_code/launch.json`.
+2. Find the command you are looking to run, for example: `Launch ado2gh`.
+3. Update the args property to have the arguments you're looking to test.
+    - "args": ["migrate-repo", "--ado-org", "<<INSERT_ORG>>", "--ado-team-project", "<<INSERT_PROJECT>>" ...]
```
synthead, line 47: How about:

```suggestion
3. Update the args property to have the arguments you're looking to test (using real org and project names):
    - "args": ["migrate-repo", "--ado-org", "example-org", "--ado-team-project", "example-project" ...]
```

synthead (approved):
Looking great!  I dig the quick tips docs!

Got some slightly opinionated suggestions you can consider :ok_hand: Certainly not blocking, so if you like it as-is, feel free to merge!

CONTRIBUTING.md
```diff
@@ -20,6 +20,41 @@ Use the [Discussions](https://github.com/github/gh-gei/discussions) tab in this
...
+
+## Debugging code 
+
+There are two ways to debug code within VS code. 
+
+### Run GH-GHEI command locally
+Run the following commands in your terminal depending on the provider you are looking to run the commands for. 
+
+Azure DevOps: 
+`dotnet run --project src/ado2gh/ado2gh.csproj -- [command]`
```
begonaguereca, line 32:
```suggestion
Azure DevOps: 
- Generic: `dotnet run --project src/ado2gh/ado2gh.csproj -- [command]`
- Example: `dotnet run --project src/ado2gh/ado2gh.csproj -- migrate-repo --ado-org "test-org" --ado-team-project "test-project" --ado-repo "test-ado-repo" --github-org "test-github-org" --github-repo "test-github-repo" `
```
@synthead I liked your idea about adding a real example here! How do you feel about this? Happy to reformat

CONTRIBUTING.md
```diff
@@ -20,6 +20,41 @@ Use the [Discussions](https://github.com/github/gh-gei/discussions) tab in this
...
+    - "args": ["migrate-repo", "--ado-org", "<<INSERT_ORG>>", "--ado-team-project", "<<INSERT_PROJECT>>" ...]
+4. Set a breakpoint within your code as needed.
+5. Navigate to the `Run and debug` side panel option.
+6. Navigate to the drop down menu and select the command you would like to run, for example: `Launch ado2gh`.
+7. Press the play button
+
+### Useful links 
+
+1. [Run a .NET app](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-run)
+2. [Debugging using vscode](https://code.visualstudio.com/docs/editor/debugging)
```
begonaguereca, line 56: @ArinGhazarian  thank you for pointing out helpful links!

CONTRIBUTING.md
synthead: Ooh yeah, I like it!  I would also remove the quotes in the example, since there aren't any special characters in your example.  Or alternatively, you could use a simple arg, like `--help` :slightly_smiling_face: Feel free to format it however it makes sense to you :+1:

CONTRIBUTING.md
```diff
@@ -20,6 +20,41 @@ Use the [Discussions](https://github.com/github/gh-gei/discussions) tab in this
...
+
+Azure DevOps: 
+`dotnet run --project src/ado2gh/ado2gh.csproj -- [command]`
+
+Bitbucket Server: 
+`dotnet run --project src/bbs2gh/bbs2gh.csproj -- [command]`
+
+GitHub:
+`dotnet run --project src/gei/gei.csproj -- [command]`
+
```
gohana, line 39: A few days ago Arin helped get set up in a C# repl to test file uploads with our [AwsApi](https://github.com/github/gh-gei/blob/main/src/Octoshift/Services/AwsApi.cs). Not sure if this is the right place for it, but it might be useful to others debugging so thought I'd suggest.
````suggestion

### Run code in a C# REPL
In your terminal:
`dotnet tool install --global CSharpRepl --version 0.4.0`
(v0.4.0 is the latest version compatible with .NET 6.0)

Run it:
`csharprepl -r src/bbs2gh/bin/Debug/net6.0/Octoshift.dll`

Then load up assemblies:
```csharp
#r "Octoshift.dll"

// You might need others, for example the AWS SDK:
#r "AWSSDK.Core.dll"
#r "AWSSDK.S3.dll"

using OctoshiftCLI.Services;
```

````

Comment (ArinGhazarian) [reply CONTRIBUTING.md:39]: This is going to be a great addition, just a few notes:
1 . Add some introduction, e.g. 

You can use a C# REPL to execute any C# code. There are many C# REPLs available, [CSharpRepl](https://fuqua.io/CSharpRepl/) is one of them and can be installed globally with the following:

`dotnet tool install --global CSharpRepl`

2. Not sure if the version is required.
3. I would add a comment before `using OctoshfitCLI.Services;`, something like this:
```csharp
// Add necessary usings
using OctoshiftCLI.Services

// Instantiate your classes
var aws = new AwsApi("access-key-id", "secret-access-key");
```

timrogers (approved):
Thanks for taking the time to put this together @begonaguereca, and for working to make this project more approachable and welcoming 💟  

As a total novice to C#, I didn't find it easy to take my first steps with this project, and I'm sure that'll be the case for others too. These changes look like a great way to reduce the barriers to entry, and I would have personally found them helpful 🙏🏻

CONTRIBUTING.md
begonaguereca: Love this addition! Could you give a little more context of the situation it wold be used in?

CONTRIBUTING.md
begonaguereca: Updated!

CONTRIBUTING.md
ArinGhazarian: @begonaguereca, it's a general purpose tool just like `irb` that gives you an interactive console to do whatever you like, for example instantiating `AwsApi` to test file i/o, etc.

begonaguereca: Thank you! Ok I just merged both of your suggestions together <3

brianaj (approved):

PR #1152, opened 2023-10-24
Redact `?token=` querystring parameter from GHES <3.8 archive URLs in logs

### LINKED ISSUE #1151: Redact tokens from logs when migrating from GHES
## Description

Currently, GEI prints logs which include raw token (`AAAAAAAAAAAAAAAAAAAAAAA` in the following example) to download the exported archive from GHES. It would be great if they could be redacted as `***`.

```
2023-10-24T00:02:44.8959849Z [2023-10-24 00:02:44] [INFO] Waiting for archive with id 333 generation to finish. Current status: exported
2023-10-24T00:02:44.8961100Z [2023-10-24 00:02:44] [DEBUG] HTTP GET: https://github-enterprise.example.com/api/v3/orgs/data/migrations/333/archive
2023-10-24T00:02:45.0377341Z [2023-10-24 00:02:45] [DEBUG] GITHUB REQUEST ID: uuid-uuid-uuid
2023-10-24T00:02:45.0379998Z [2023-10-24 00:02:45] [DEBUG] RESPONSE (Redirect): https://media.github-enterprise.example.com/migrations/333/archive/uuid-uuid-uuid?token=AAAAAAAAAAAAAAAAAAAAAAA
2023-10-24T00:02:45.0383154Z [2023-10-24 00:02:45] [INFO] Archive (metadata) download url: https://media.github-enterprise.example.com/migrations/333/archive/uuid-uuid-uuid?token=AAAAAAAAAAAAAAAAAAAAAAA
...
2023-10-24T00:03:30.7471800Z [2023-10-24 00:03:30] [INFO] Archives uploaded to blob storage, now starting migration...
```

timrogers:
Existing code allows us to redact secrets in the logs to stop them being outputted to the shell and to log files. Secret strings are registered during execution, and then replaced in the output.

This PR also allows classes using `OctoLogger` to register regular expressions to replace, and begins redacting GHES download tokens for GHES <3.8 using a simple regular expression.

Fixes #1151.

- [x] Issue linked

### DIFF
--- a/src/Octoshift/Services/OctoLogger.cs
+++ b/src/Octoshift/Services/OctoLogger.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Net.Http;
 using System.Text.RegularExpressions;
 
@@ -19,6 +20,7 @@ public class OctoLogger
 {
     public virtual bool Verbose { get; set; }
     private readonly HashSet<Regex> _redactionPatterns = new();
+    private readonly HashSet<string> _secrets = new();
     private readonly string _logFilePath;
     private readonly string _verboseFilePath;
     private readonly bool _debugMode;
@@ -81,6 +83,12 @@ public class OctoLogger
     private string Redact(string msg)
     {
         var result = msg;
+
+        foreach (var secret in _secrets.Where(x => x is not null))
+        {
+            result = result.Replace(secret, "***")
+                .Replace(Uri.EscapeDataString(secret), "***");
+        }
 
         foreach (var redactionPattern in _redactionPatterns)
         {
@@ -155,14 +163,7 @@ public class OctoLogger
         Console.ResetColor();
     }
 
-    public virtual void RegisterSecret(string secret)
-    {
-        if (secret is not null)
-        {
-            _redactionPatterns.Add(new Regex(secret, RegexOptions.IgnoreCase));
-            _redactionPatterns.Add(new Regex(Uri.EscapeDataString(secret), RegexOptions.IgnoreCase));
-        }
-    }
+    public virtual void RegisterSecret(string secret) => _secrets.Add(secret);
 
     public virtual void AddRedactionPattern(Regex pattern)
     {
--- a/src/Octoshift/Services/OctoLogger.cs
+++ b/src/Octoshift/Services/OctoLogger.cs
@@ -1,8 +1,8 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.IO;
-using System.Linq;
 using System.Net.Http;
+using System.Text.RegularExpressions;
 
 namespace OctoshiftCLI.Services;
 
@@ -18,7 +18,7 @@ internal static class LogLevel
 public class OctoLogger
 {
     public virtual bool Verbose { get; set; }
-    private readonly HashSet<string> _secrets = new();
+    private readonly HashSet<Regex> _redactionPatterns = new();
     private readonly string _logFilePath;
     private readonly string _verboseFilePath;
     private readonly bool _debugMode;
@@ -59,7 +59,7 @@ public class OctoLogger
     private void Log(string msg, string level)
     {
         var output = FormatMessage(msg, level);
-        output = MaskSecrets(output);
+        output = Redact(output);
         if (level == LogLevel.ERROR)
         {
             _writeToConsoleError(output);
@@ -78,14 +78,13 @@ public class OctoLogger
         return $"[{timeFormat}] [{level}] {msg}\n";
     }
 
-    private string MaskSecrets(string msg)
+    private string Redact(string msg)
     {
         var result = msg;
 
-        foreach (var secret in _secrets.Where(x => x is not null))
+        foreach (var redactionPattern in _redactionPatterns)
         {
-            result = result.Replace(secret, "***")
-                .Replace(Uri.EscapeDataString(secret), "***");
+            result = redactionPattern.Replace(result, "***");
         }
 
         return result;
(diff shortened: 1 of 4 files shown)


src/Octoshift/Services/OctoLogger.cs
```diff
@@ -156,5 +155,17 @@ public virtual void LogSuccess(string msg)
         Console.ResetColor();
     }
 
-    public virtual void RegisterSecret(string secret) => _secrets.Add(secret);
```
timrogers, line 159: We could also avoid this refactor, and keep a separate list of `_secrets` which we replace as well as the "redaction patterns". That's probably a safer approach to avoid surprises - but I'd love to hear what y'all think.

src/gei/Commands/MigrateRepo/MigrateRepoCommand.cs
```diff
@@ -188,6 +189,8 @@ public override MigrateRepoCommandHandler BuildHandler(MigrateRepoCommandArgs ar
             var ghesVersionChecker = ghesVersionCheckerFactory.Create(ghesApi);
             var warningsCountLogger = sp.GetRequiredService<WarningsCountLogger>();
 
+            log.AddRedactionPattern(new Regex("\\?token=[A-Z0-9]{29}"));
```
timrogers, line 192: I have manually tested that this works as expected on GHES 3.7, and the URL is not logged.

src/Octoshift/Services/OctoLogger.cs
```diff
@@ -78,7 +80,7 @@ private string FormatMessage(string msg, string level)
         return $"[{timeFormat}] [{level}] {msg}\n";
     }
 
-    private string MaskSecrets(string msg)
+    private string Redact(string msg)
```
ArinGhazarian, line 83: nit: I like the old name better. To me redaction is deletion of something but masking is replacing the sensitive data with some mask characters.

src/Octoshift/Services/OctoLogger.cs
```diff
@@ -157,4 +164,9 @@ public virtual void LogSuccess(string msg)
     }
 
     public virtual void RegisterSecret(string secret) => _secrets.Add(secret);
+
+    public virtual void AddRedactionPattern(Regex pattern)
+    {
+        _redactionPatterns.Add(pattern);
+    }
```
ArinGhazarian, line 171: Instead of exposing a public method to register regex patters, I would prefer to keep a list of all patterns internally.

src/Octoshift/Services/OctoLogger.cs
```diff
@@ -18,6 +19,7 @@ internal static class LogLevel
 public class OctoLogger
 {
     public virtual bool Verbose { get; set; }
+    private readonly HashSet<Regex> _redactionPatterns = new();
```
ArinGhazarian, line 22: I would keep all patterns as strings right here in the `OctoLogger` class instead of registering them via a public method, so in this case we will have:
```c#
private readonly List<string> _recationPatterns = new()
{
    "your regex pattern here"
};
```

src/gei/Commands/MigrateRepo/MigrateRepoCommand.cs
ArinGhazarian, line 192: Two things:
1. I would move this into the`OctoLogger` itself.
2. this will also replace the `token=`, what we preferably want is to just mask the secret, for that the regex pattern needs to be a bit more complex, like this --> `\b(?<=token=)(.+?)\b`, I used look behind (`?<=`) in order to not replace the `token=`.
3. Even if you want to keep this logic (instead of moving the pattern into `OctoLogger`), I would do the registration inside the handler and not in here. The `BuildHandler`'s purpose is to just create the handler and nothing else.

src/OctoshiftCLI.Tests/Octoshift/Services/OctoLoggerTests.cs
```diff
@@ -60,6 +61,31 @@ public void Secrets_Should_Be_Masked_From_Logs_And_Console()
         _consoleError.Should().NotContain(urlEncodedSecret);
     }
 
+    [Fact]
+    public void Redaction_Patterns_Should_Be_Replaced_In_Logs_And_Console()
```
ArinGhazarian, line 65: I would add a test to exactly test the pattern too.

src/Octoshift/Services/OctoLogger.cs
timrogers: I prefer `Redact` now that it masks secrets and generic, hard-coded patterns. But I am happy to leave the name as it is rather than change things, since you don't feel the same 😊

ArinGhazarian: I don't feel strongly about it, will leave it up to you :)

src/Octoshift/Services/OctoLogger.cs
```diff
@@ -30,6 +31,11 @@ public class OctoLogger
 
     private const string GENERIC_ERROR_MESSAGE = "An unexpected error happened. Please see the logs for details.";
 
+    private readonly HashSet<Regex> _redactionPatterns = new()
```
ArinGhazarian, line 34: two things here:
1. The `HashSet` will make no difference here. It's better to use lighter types like an `Array` or `List`.
2. Even though it's trivial, I think it's easier to just keep a list (or array) of strings rather than actual regex objects. It's less allocation and easier to maintain.

src/Octoshift/Services/OctoLogger.cs
```diff
@@ -88,6 +94,11 @@ private string MaskSecrets(string msg)
                 .Replace(Uri.EscapeDataString(secret), "***");
         }
 
+        foreach (var redactionPattern in _redactionPatterns)
+        {
+            result = redactionPattern.Replace(result, "***");
```
ArinGhazarian, line 99: if you keep a list of strings, this line will change to the following which uses the static `Replace` method instead of allocating new Regex objects or using many pre-allocated ones.
```suggestion
            result = Regex.Replace(result, redactionPattern, "***");
```

timrogers: Addressed in 342be5b

src/Octoshift/Services/OctoLogger.cs
timrogers: Addressed in 342be5b

src/Octoshift/Services/OctoLogger.cs
timrogers: I'm going to go for `Redact` as I do think it's clearer - done in 062d0c2

ArinGhazarian (approved):

PR #1165, opened 2023-11-15
Fix flaky tests

ArinGhazarian:
This PR fixes [flaky tests](https://github.com/github/gh-gei/actions/runs/6870726372/job/18686177642?pr=1162) that was caused by trying to access the `CliContext` without setting it first. The `CliContext` values are only going to be set during the execution of a command but in unit tests if we want to use the context values (`CliContext.RootCommand` and `CliConext.ExecutingCommand`) we need to set them first. 

Interestingly the flakiness was happening because of a race condition! The flaky tests were in `OctoshiftCLI.Tests.Octoshift.Commands.WaitForMigration.WaitForMigrationCommandHandlerTests` and specifically the ones that were relying on the value of `CliContext.RootCommand` (for example `With_Migration_ID_That_Succeeds_With_1_Warning`). We used to set the context values only in the `VersionCheckerTests.cs` and because of a race condition the expected and actual values of the following string might have different values depending on whether `WaitForMigrationCommandHandlerTests.With_Migration_ID_That_Succeeds_With_1_Warning` or `VersionCheckerTests.GetVersionComments_Returns_Root_And_Executing_Commands` runs first:

```c#
$"Migration log available at {MIGRATION_URL} or by running `gh {CliContext.RootCommand} download-logs`"
```

To fix the problem, a new static helper methods (`SetCliContext`) was added to `TestHelpers` to set the CLI context values in a thread safe manner to make sure they are going to be set just once. Any test that needs the context values need to call `TestHelpers.SetCliContext` first to make sure the values are set. 

- [x] Issue linked

### DIFF
--- a/src/OctoshiftCLI.Tests/Octoshift/Commands/WaitForMigration/WaitForMigrationCommandHandlerTests.cs
+++ b/src/OctoshiftCLI.Tests/Octoshift/Commands/WaitForMigration/WaitForMigrationCommandHandlerTests.cs
@@ -29,6 +29,8 @@ public class WaitForMigrationCommandHandlerTests
         {
             WaitIntervalInSeconds = WAIT_INTERVAL
         };
+
+        TestHelpers.SetCliContext();
     }
 
     [Fact]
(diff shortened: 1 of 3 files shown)


begonaguereca (approved):
🚀

PR #1169, opened 2023-11-21
Redact `X-Amz-Credential` querystring parameters from URLs in logs

### LINKED ISSUE #1159: Redact Sensitive S3 URL Parameters
## Description
Raising as a separate ticket on request of @timrogers. This was originally noted as a comment here by @kmaehashi: https://github.com/github/gh-gei/issues/1151#issuecomment-1779001488

On GHES 3.8+, git/metadata archives are uploaded to S3 by GHES, and GHES returns S3 signed URLs for these archives. I think parameters in these URLs (e.g. X-Amz-Credential) should also be masked as well as anyone with access to the GEI log can then access the S3 link
```
[2023-10-25 10:44:25] [INFO] Waiting for archive with id 347 generation to finish. Current status: exported
[2023-10-25 10:44:25] [DEBUG] HTTP GET: https://github-enterprise.example.com/api/v3/orgs/myorgname/migrations/347/archive
[2023-10-25 10:44:25] [DEBUG] GITHUB REQUEST ID: uuid-uuid-uuid
[2023-10-25 10:44:25] [DEBUG] RESPONSE (Redirect): https://example-s3-bucket-name.s3.amazonaws.com/uuid-uuid-uuid.tar.gz?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=AAAAAAAAAAAAAAAAAAAAAAA&X-Amz-Date=20231025T104425Z&X-Amz-Expires=172800&X-Amz-Signature=AAAAAAAAAAAAAAAAAAAAAAA&X-Amz-SignedHeaders=host&actor_id=1&key_id=0&repo_id=0&response-content-disposition=filename%3Duuid-uuid-uuid.tar.gz&response-content-type=application%2Fx-gzip
[2023-10-25 10:44:25] [INFO] Archive (metadata) download url: https://example-s3-bucket-name.s3.amazonaws.com/uuid-uuid-uuid.tar.gz?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=AAAAAAAAAAAAAAAAAAAAAAA&X-Amz-Date=20231025T104425Z&X-Amz-Expires=172800&X-Amz-Signature=AAAAAAAAAAAAAAAAAAAAAAA&X-Amz-SignedHeaders=host&actor_id=1&key_id=0&repo_id=0&response-content-disposition=filename%3Duuid-uuid-uuid.tar.gz&response-content-type=application%2Fx-gzip
```

timrogers:
After uploading archives to AWS, we log the URL in a few different places. 

It isn't desirable for unredacted URLs to be available in the log output as this represents a security risk, providing access to potentially sensitive files.

This redacts the AWS URLs, removing the `X-Aws-Credential` argument so they will no longer be "live" URLs.

Fixes #1159.

As part of this PR, I'm also fixing some other `OctoLogger` tests which parse but are incorrect, because variable interpolation is not enabled for relevant strings.

- [x] Issue linked

### DIFF
--- a/src/Octoshift/Services/OctoLogger.cs
+++ b/src/Octoshift/Services/OctoLogger.cs
@@ -33,7 +33,8 @@ public class OctoLogger
 
     private readonly List<string> _redactionPatterns = new()
     {
-        "\\b(?<=token=)(.+?)\\b"
+        "\\b(?<=token=)(.+?)\\b",
+        "\\b(?<=X-Amz-Credential=)(.+?)\\b",
     };
 
     public OctoLogger()
--- a/src/Octoshift/Services/OctoLogger.cs
+++ b/src/Octoshift/Services/OctoLogger.cs
@@ -97,7 +97,7 @@ public class OctoLogger
 
         foreach (var redactionPattern in _redactionPatterns)
         {
-            result = Regex.Replace(result, redactionPattern, "***");
+            result = Regex.Replace(result, redactionPattern, "***", RegexOptions.IgnoreCase);
         }
 
         return result;
--- a/src/OctoshiftCLI.Tests/Octoshift/Services/OctoLoggerTests.cs
+++ b/src/OctoshiftCLI.Tests/Octoshift/Services/OctoLoggerTests.cs
@@ -63,25 +63,33 @@ public class OctoLoggerTests
     [Fact]
     public void Ghes_Archive_Url_Tokens_Should_Be_Replaced_In_Logs_And_Console()
     {
-
         var ghesArchiveUrl = "https://files.github.acmeinc.com/foo?token=foobar";
+        var variants = new[]
+        {
+            ghesArchiveUrl,
+            ghesArchiveUrl.ToUpper(),
+            ghesArchiveUrl.ToLower()
+        };
 
-        _octoLogger.Verbose = false;
-        _octoLogger.LogInformation($"Archive URL: {ghesArchiveUrl}");
-        _octoLogger.LogVerbose($"Archive URL: {ghesArchiveUrl}");
-        _octoLogger.LogWarning($"Archive URL: {ghesArchiveUrl}");
-        _octoLogger.LogSuccess($"Archive URL: {ghesArchiveUrl}");
-        _octoLogger.LogError($"Archive URL: {ghesArchiveUrl}");
-        _octoLogger.LogError(new OctoshiftCliException($"Archive URL: {ghesArchiveUrl}"));
-        _octoLogger.LogError(new InvalidOperationException($"Archive URL: {ghesArchiveUrl}"));
+        foreach (var variant in variants)
+        {
+            _octoLogger.Verbose = false;
+            _octoLogger.LogInformation($"Archive URL: {variant}");
+            _octoLogger.LogVerbose($"Archive URL: {variant}");
+            _octoLogger.LogWarning($"Archive URL: {variant}");
+            _octoLogger.LogSuccess($"Archive URL: {variant}");
+            _octoLogger.LogError($"Archive URL: {variant}");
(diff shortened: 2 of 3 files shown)


src/Octoshift/Services/OctoLogger.cs
```diff
@@ -33,7 +33,8 @@ public class OctoLogger
 
     private readonly List<string> _redactionPatterns = new()
     {
-        "\\b(?<=token=)(.+?)\\b"
+        "\\b(?<=token=)(.+?)\\b",
+        "\\b(?<=X-Amz-Credential=)(.+?)\\b",
```
ArinGhazarian, line 37: It'd be probably a good idea to also add case insensitivity to the regex match:
```C#
Regex.Replace(result, redactionPattern, "***", RegexOptions.IgnoreCase)
```

timrogers: Yes - that makes sense! I'll do that tomorrow and then I'll ping you for a fresh review.

timrogers: Addressed in b7bdc44

src/OctoshiftCLI.Tests/Octoshift/Services/OctoLoggerTests.cs
```diff
@@ -63,29 +63,72 @@ public void Secrets_Should_Be_Masked_From_Logs_And_Console()
...
+            _octoLogger.LogError(new InvalidOperationException($"Archive URL: {variant}"));
+
+            _octoLogger.Verbose = true;
+            _octoLogger.LogVerbose($"Archive URL: {variant}");
+
+            _consoleOutput.Should().NotContain(variant);
+            _logOutput.Should().NotContain(variant);
+            _verboseLogOutput.Should().NotContain(variant);
+            _consoleError.Should().NotContain(variant);
+        }
```
ArinGhazarian, line 92: Generally speaking a good unit test needs to test one thing rather than many different things, while sometimes there could be more than one assertion (e.g. asserting some fields in a payload), but still the preferred way should be acting on one thing and asserting the result. The reason is in case of a failure it would be easy to figure out what failed and why.

Now here this test has already been doing too much and was acting on different methods (`LogInformation`, `LogSuccess`, ...), and now it uses a loop to test against more inputs. This will make the test very hard to understand and debug. 

Fortunately `xUnit` has a way to deal with different input values w/o having to write more than one test or using loops and such. To do so you need to use `Theory` instead of a `Fact` and pass some input values with `InlineData`, read further [here](https://andrewlock.net/creating-parameterised-tests-in-xunit-with-inlinedata-classdata-and-memberdata/).

With a `Theory` your test will look like this:

```c#
[Theory]
[InlineData("https://files.github.acmeinc.com/foo?token=foobar")]
[InlineData("HTTPS://FILES.GITHUB.ACMEINC.COM/FOO?TOKEN=FOOBAR")]
public void Ghes_Archive_Url_Tokens_Should_Be_Replaced_In_Logs_And_Console(string url)
{
    ...
}
```

One last thing is, while there are many school of thoughts, I believe a unit test shouldn't try to be smart, yes sometimes for ease of maintenance we create base classes and private methods mostly for setup and tear down tasks but other than that anything else will make a unit test less maintainable IMO. For example using loops in a test, or using `input.ToUppoer` instead of simply using the upper cased string literal, etc.

ArinGhazarian (commented):
Looks good, I dropped a comment to simplify the unit tests.

src/OctoshiftCLI.Tests/Octoshift/Services/OctoLoggerTests.cs
timrogers: Done! 🏁

ArinGhazarian (approved):

PR #1173, opened 2023-11-29
Add abort migration generic commands

begonaguereca:
What is changing? 

How is it tested? 
- unit tests 
 (Once these generic commands are hooked up to the providers) 

- [x ] Appropriate logging output
- [x ] Issue linked

### DIFF
--- a/src/Octoshift/Commands/AbortMigration/AbortCommandBase.cs
+++ b/src/Octoshift/Commands/AbortMigration/AbortCommandBase.cs
@@ -1,0 +1,58 @@
+using System;
+using System.CommandLine;
+using System.Runtime.CompilerServices;
+using Microsoft.Extensions.DependencyInjection;
+using OctoshiftCLI.Contracts;
+using OctoshiftCLI.Services;
+
+[assembly: InternalsVisibleTo("OctoshiftCLI.Tests")]
+
+namespace OctoshiftCLI.Commands.AbortMigration;
+
+public class AbortMigrationCommandBase : CommandBase<AbortMigrationCommandArgs, AbortMigrationCommandHandler>
+{
+    public AbortMigrationCommandBase() : base(
+        name: "abort-migration",
+        description: "Aborts a migration that is in progress.")
+    {
+    }
+
+    public virtual Option<string> MigrationId { get; } = new("--migration-id")
+    {
+        IsRequired = true,
+        Description = "Waits for the specified migration to finish."
+    };
+
+    public virtual Option<string> GithubPat { get; } = new("--github-pat")
+    {
+        Description = "Personal access token of the GitHub target. Overrides GH_PAT environment variable."
+    };
+
+    public virtual Option<bool> Verbose { get; } = new("--verbose");
+
+    protected void AddOptions()
+    {
+        AddOption(MigrationId);
+        AddOption(GithubPat);
+        AddOption(Verbose);
+    }
+
+    public override AbortMigrationCommandHandler BuildHandler(AbortMigrationCommandArgs args, IServiceProvider sp)
+    {
+        if (args is null)
+        {
+            throw new ArgumentNullException(nameof(args));
+        }
+
+        if (sp is null)
+        {
+            throw new ArgumentNullException(nameof(sp));
+        }
+
+        var log = sp.GetRequiredService<OctoLogger>();
+        var githubApi = sp.GetRequiredService<ITargetGithubApiFactory>().Create(targetPersonalAccessToken: args.GithubPat);
+
+
+        return new AbortMigrationCommandHandler(log, githubApi);
+    }
+}
--- a/src/Octoshift/Commands/AbortMigration/AbortMigrationCommandArgs.cs
+++ b/src/Octoshift/Commands/AbortMigration/AbortMigrationCommandArgs.cs
@@ -1,0 +1,26 @@
+﻿using OctoshiftCLI.Extensions;
+using OctoshiftCLI.Services;
+
+namespace OctoshiftCLI.Commands.AbortMigration;
+
+public class AbortMigrationCommandArgs : CommandArgs
+{
+    public string MigrationId { get; set; }
+    [Secret]
+    public string GithubPat { get; set; }
+
+    public const string REPO_MIGRATION_ID_PREFIX = "RM_";
+
+    public override void Validate(OctoLogger log)
+    {
+        if (MigrationId.IsNullOrWhiteSpace())
+        {
+            throw new OctoshiftCliException("MigrationId must be provided");
+        }
+
+        if (!MigrationId.StartsWith(REPO_MIGRATION_ID_PREFIX))
+        {
+            throw new OctoshiftCliException($"Invalid migration id: {MigrationId}");
+        }
+    }
+}
--- a/src/Octoshift/Commands/AbortMigration/AbortMigrationCommandHandler.cs
+++ b/src/Octoshift/Commands/AbortMigration/AbortMigrationCommandHandler.cs
@@ -1,0 +1,45 @@
+using System;
+using System.Runtime.CompilerServices;
+using System.Threading.Tasks;
+using OctoshiftCLI.Services;
+
+[assembly: InternalsVisibleTo("OctoshiftCLI.Tests")]
+
+namespace OctoshiftCLI.Commands.AbortMigration;
+
+public class AbortMigrationCommandHandler : ICommandHandler<AbortMigrationCommandArgs>
+{
+    internal int WaitIntervalInSeconds = 10;
+
+    private readonly OctoLogger _log;
+    private readonly GithubApi _githubApi;
+
+    public AbortMigrationCommandHandler(OctoLogger log, GithubApi githubApi)
+    {
+        _log = log;
+        _githubApi = githubApi;
+    }
+
+    public async Task Handle(AbortMigrationCommandArgs args)
+    {
+        if (args is null)
+        {
+            throw new ArgumentNullException(nameof(args));
+        }
+
+        await AbortRepositoryMigration(args.MigrationId);
+    }
+
+    private async Task AbortRepositoryMigration(string migrationId)
+    {
+        var abortion_state = await _githubApi.AbortMigration(migrationId);
+
+        if (abortion_state == false)
+        {
+            _log.LogError($"Failed to abort migration {migrationId}");
+            return;
+        }
+
+        _log.LogInformation($"Migration {migrationId} was cancelled");
+    }
+}
--- a/src/OctoshiftCLI.Tests/Octoshift/Commands/AbortMigration/AbortMirationCommandHandlerTests.cs
+++ b/src/OctoshiftCLI.Tests/Octoshift/Commands/AbortMigration/AbortMirationCommandHandlerTests.cs
@@ -1,0 +1,69 @@
+using System.Threading.Tasks;
+using Moq;
+using OctoshiftCLI.Commands.AbortMigration;
+using OctoshiftCLI.Services;
+using Xunit;
+
+namespace OctoshiftCLI.Tests.Octoshift.Commands.AbortMigration;
+
+public class AbortMigrationCommandHandlerTests
+{
+    private readonly Mock<GithubApi> _mockGithubApi = TestHelpers.CreateMock<GithubApi>();
+    private readonly Mock<OctoLogger> _mockOctoLogger = TestHelpers.CreateMock<OctoLogger>();
+
+    private readonly WarningsCountLogger _warningsCountLogger;
+    private readonly AbortMigrationCommandHandler _handler;
+
+    private const string REPO_MIGRATION_ID = "RM_MIGRATION_ID";
+
+    public AbortMigrationCommandHandlerTests()
+    {
+        _warningsCountLogger = new WarningsCountLogger(_mockOctoLogger.Object);
+        _handler = new AbortMigrationCommandHandler(_mockOctoLogger.Object, _mockGithubApi.Object);
+
+        TestHelpers.SetCliContext();
+    }
+
+    [Fact]
+    public async Task With_Migration_ID_That_Succeeds()
+    {
+        // Arrange
+        _mockGithubApi.Setup(x => x.AbortMigration(It.IsAny<string>())).ReturnsAsync(true);
+
+        // Act
+        var args = new AbortMigrationCommandArgs
+        {
+            MigrationId = REPO_MIGRATION_ID,
+        };
+        await _handler.Handle(args);
+
+        // Assert
+        _mockOctoLogger.Verify(m => m.LogInformation(It.IsAny<string>()), Times.Exactly(7));
+        _mockOctoLogger.Verify(m => m.LogSuccess(It.IsAny<string>()), Times.Once);
+        _mockGithubApi.Verify(m => m.AbortMigration(REPO_MIGRATION_ID), Times.Exactly(1));
+
+        _mockGithubApi.VerifyNoOtherCalls();
+    }
+
+    [Fact]
+    public async Task With_Migration_ID_That_Fails()
+    {
+        // Arrange
+        _mockGithubApi.Setup(x => x.AbortMigration(It.IsAny<string>())).ReturnsAsync(false);
+
+        // Act
+        var args = new AbortMigrationCommandArgs
+        {
+            MigrationId = REPO_MIGRATION_ID,
+        };
+        await _handler.Handle(args);
+
+        // Assert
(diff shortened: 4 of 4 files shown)


src/OctoshiftCLI.Tests/Octoshift/Commands/AbortMigration/AbortMirationCommandHandlerTests.cs
```diff
@@ -0,0 +1,65 @@
...
+    private readonly AbortMigrationCommandHandler _handler;
+
+    private const string REPO_MIGRATION_ID = "RM_MIGRATION_ID";
+
+    public AbortMigrationCommandHandlerTests()
+    {
+        _warningsCountLogger = new WarningsCountLogger(_mockOctoLogger.Object);
+        _handler = new AbortMigrationCommandHandler(_mockOctoLogger.Object, _mockGithubApi.Object);
+
+        TestHelpers.SetCliContext();
```
ArinGhazarian, line 24: no need to set the context
```suggestion
        TestHelpers.SetCliContext();
```

src/OctoshiftCLI.Tests/Octoshift/Commands/AbortMigration/AbortMirationCommandHandlerTests.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        // Act
+        var args = new AbortMigrationCommandArgs
+        {
+            MigrationId = REPO_MIGRATION_ID,
+        };
+        await _handler.Handle(args);
+
+        // Assert
+        _mockOctoLogger.Verify(m => m.LogSuccess(It.IsAny<string>()), Times.Once);
```
ArinGhazarian, line 41: no need to verify any logs here, logging in only informational. 
```suggestion
```

src/OctoshiftCLI.Tests/Octoshift/Commands/AbortMigration/AbortMirationCommandHandlerTests.cs
```diff
@@ -0,0 +1,65 @@
...
+        _mockGithubApi.Verify(m => m.AbortMigration(REPO_MIGRATION_ID), Times.Exactly(1));
+        _mockGithubApi.VerifyNoOtherCalls();
+    }
```
ArinGhazarian, line 63: I would remove this test since the only difference is the log output. a similar example is [RevokeMigrationRoleCommandHandlerTests.cs](https://github.com/github/gh-gei/blob/a6d92bb445ac5ad58117cfe6cedf94d3121fc569/src/OctoshiftCLI.Tests/Octoshift/Commands/RevokeMigratorRole/RevokeMigratorRoleCommandHandlerTests.cs) that only tests for a happy path. 
```suggestion
```

src/OctoshiftCLI.Tests/Octoshift/Commands/AbortMigration/AbortMirationCommandHandlerTests.cs
```diff
@@ -0,0 +1,65 @@
...
+        var args = new AbortMigrationCommandArgs
+        {
+            MigrationId = REPO_MIGRATION_ID,
+        };
+        await _handler.Handle(args);
+
+        // Assert
+        _mockOctoLogger.Verify(m => m.LogSuccess(It.IsAny<string>()), Times.Once);
+        _mockGithubApi.Verify(m => m.AbortMigration(REPO_MIGRATION_ID), Times.Exactly(1));
+        _mockGithubApi.VerifyNoOtherCalls();
```
ArinGhazarian, line 43: This last verify is not super necessary 
```suggestion
```

src/Octoshift/Commands/AbortMigration/AbortCommandBase.cs
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        if (sp is null)
+        {
+            throw new ArgumentNullException(nameof(sp));
+        }
+
+        var log = sp.GetRequiredService<OctoLogger>();
+        var githubApi = sp.GetRequiredService<ITargetGithubApiFactory>().Create(targetPersonalAccessToken: args.GithubPat);
+
```
ArinGhazarian, line 54: nit: extra new line
```suggestion
```

src/Octoshift/Commands/AbortMigration/AbortMigrationCommandArgs.cs
```diff
@@ -0,0 +1,26 @@
...
+    public override void Validate(OctoLogger log)
+    {
+        if (MigrationId.IsNullOrWhiteSpace())
+        {
+            throw new OctoshiftCliException("MigrationId must be provided");
+        }
+
+        if (!MigrationId.StartsWith(REPO_MIGRATION_ID_PREFIX))
+        {
+            throw new OctoshiftCliException($"Invalid migration id: {MigrationId}");
```
ArinGhazarian, line 23: I would explain more why the migration id is invalid:
```suggestion
            throw new OctoshiftCliException($"Invalid migration id: {MigrationId}. Only repo migrations that start with \"RM_\" are currently supported.");
```

src/Octoshift/Commands/AbortMigration/AbortMigrationCommandArgs.cs
```diff
@@ -0,0 +1,26 @@
...
+
+namespace OctoshiftCLI.Commands.AbortMigration;
+
+public class AbortMigrationCommandArgs : CommandArgs
+{
+    public string MigrationId { get; set; }
+    [Secret]
+    public string GithubPat { get; set; }
+
+    public const string REPO_MIGRATION_ID_PREFIX = "RM_";
```
ArinGhazarian, line 12: We don't need to expose this as a public const, so I would just move this into the `Validate` method. Also since it will be a local `const` it can be a camel cased const variable:
```c#
const string repoMigrationIdPrefix = "RM_";
```

src/Octoshift/Commands/AbortMigration/AbortMigrationCommandHandler.cs
```diff
@@ -0,0 +1,45 @@
+using System;
+using System.Runtime.CompilerServices;
+using System.Threading.Tasks;
+using OctoshiftCLI.Services;
+
+[assembly: InternalsVisibleTo("OctoshiftCLI.Tests")]
```
ArinGhazarian, line 6: no need for this, it's probably a left over from a copy/paste

```suggestion
```

src/Octoshift/Commands/AbortMigration/AbortMigrationCommandHandler.cs
```diff
@@ -0,0 +1,45 @@
...
+
+namespace OctoshiftCLI.Commands.AbortMigration;
+
+public class AbortMigrationCommandHandler : ICommandHandler<AbortMigrationCommandArgs>
+{
+    internal int WaitIntervalInSeconds = 10;
```
ArinGhazarian, line 12: no need for this, it's probably a left over from a copy/paste
```suggestion
```

src/Octoshift/Commands/AbortMigration/AbortMigrationCommandHandler.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            throw new ArgumentNullException(nameof(args));
+        }
+
+        await AbortRepositoryMigration(args.MigrationId);
+    }
+
+    private async Task AbortRepositoryMigration(string migrationId)
+    {
+        var abortion_state = await _githubApi.AbortMigration(migrationId);
```
ArinGhazarian, line 35: All local variables in C# are camel cased by convention, another suggestion would be to change the name to `success`:
```suggestion
        var success = await _githubApi.AbortMigration(migrationId);
```

src/Octoshift/Commands/AbortMigration/AbortMigrationCommandHandler.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        if (abortion_state == false)
```
ArinGhazarian, line 37:
```suggestion
        if (!success)
```

src/Octoshift/Commands/AbortMigration/AbortMigrationCommandHandler.cs
```diff
@@ -0,0 +1,45 @@
+using System;
+using System.Runtime.CompilerServices;
```
ArinGhazarian, line 2:
```suggestion
```

ArinGhazarian (changes requested):

timrogers:
Excited to see this ship ❤️ 👏🏻

src/OctoshiftCLI.Tests/Octoshift/Commands/AbortMigration/AbortMirationCommandHandlerTests.cs
begonaguereca: I think this one is actually helpful! Ensuring this method is only doing one thing

src/Octoshift/Commands/AbortMigration/AbortMigrationCommandArgs.cs
begonaguereca: Why are we assuming the error?

src/Octoshift/Commands/AbortMigration/AbortMigrationCommandHandler.cs
begonaguereca: Woops forgot about the camel case, i like `abortionState` i think thats more descriptive of the actual value

src/Octoshift/Commands/AbortMigration/AbortMigrationCommandHandler.cs
```diff
@@ -0,0 +1,42 @@
...
+        await AbortRepositoryMigration(args.MigrationId);
+    }
+
+    private async Task AbortRepositoryMigration(string migrationId)
+    {
+        var abortionState = await _githubApi.AbortMigration(migrationId);
+
+        if (!abortionState)
+        {
+            _log.LogError($"Failed to abort migration {migrationId}");
```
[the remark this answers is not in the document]
ArinGhazarian: I agree but this needs to come from the backend API so I guess we need another ticket to tweak the API.

src/Octoshift/Commands/AbortMigration/AbortMigrationCommandArgs.cs
ArinGhazarian: True but it was exposed to be used outside of the class but here we don't do that.

src/Octoshift/Commands/AbortMigration/AbortMigrationCommandHandler.cs
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        await AbortRepositoryMigration(args.MigrationId);
+    }
+
+    private async Task AbortRepositoryMigration(string migrationId)
+    {
+        var abortionState = await _githubApi.AbortMigration(migrationId);
+
+        if (!abortionState)
```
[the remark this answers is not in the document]
begonaguereca: I read it as a failed abortion state. 

Ok sounds like you have a very strong opinion on this. Would it be your preference for me to change it?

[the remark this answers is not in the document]
ArinGhazarian: Usually I don't have strong opinions for variable naming but for this case the consensus is to somehow indicate that the variable is a boolean usually by prepending `is` or something meaningful, `aborttionState` doesn't indicate that it only has two states right? For further reading:
- https://github.com/ktaranov/naming-convention/blob/master/C%23%20Coding%20Standards%20and%20Naming%20Conventions.md#26-do-use-prefix-any-is-have-or-similar-keywords-for-boolean-identifier
- https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig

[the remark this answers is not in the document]
begonaguereca: Thank you for linking the literature! I am new to c# and am still learning all the ways it's opinionated as a language. 

How would you feel about is `TrueOrFalse`?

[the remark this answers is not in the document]
begonaguereca: Maybe this would also be something great to add to the `Contributing.md`?

ArinGhazarian (approved):

src/Octoshift/Commands/AbortMigration/AbortCommandBase.cs
```diff
@@ -0,0 +1,57 @@
...
+        IsRequired = true,
+        Description = "Waits for the specified migration to finish."
+    };
+
+    public virtual Option<string> GithubPat { get; } = new("--github-pat")
+    {
+        Description = "Personal access token of the GitHub target. Overrides GH_PAT environment variable."
+    };
+
+    public virtual Option<bool> Verbose { get; } = new("--verbose");
```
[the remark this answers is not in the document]
begonaguereca: https://github.com/github/gh-gei/issues/1174

✅

src/Octoshift/Commands/AbortMigration/AbortMigrationCommandHandler.cs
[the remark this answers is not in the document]
ArinGhazarian: In that case the GQL endpoint will return an error and then the CLI will show a generic error in non verbose mode and log the error anyways. And of course the error message is getting returned by the backend. I think this default behavior is good enough unless we have a strong case to tweak the api call method in the CLI.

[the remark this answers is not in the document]
ArinGhazarian: We can create a ticket to tweak the backend API to not return and error for those kind of failures (e.g. migration id is not found, etc.) and instead returns a payload that has a success and failureReason that will be set in case of a failure.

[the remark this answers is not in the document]
ArinGhazarian: Based on what we did for [revoke migrator role command](https://github.com/github/gh-gei/blob/a6d92bb445ac5ad58117cfe6cedf94d3121fc569/src/Octoshift/Commands/RevokeMigratorRole/RevokeMigratorRoleCommandHandler.cs#L36), I think this is a similar situation where the command itself was successful but it couldn't abort the migration (the returned result is false). Whereas an error would be a situation where an actual exception happens (i.e. migration id not found our invalid). 
that being said we can always change our approach and throw.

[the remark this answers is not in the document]
begonaguereca: While the error message would be surfaced as is I think its a nice idea to surface explicit error messages for each of these scenarios! Created [this](https://github.ghe.com/github/octoshift/issues/7842) WI for the refactor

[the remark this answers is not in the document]
timrogers: > We can create a ticket to tweak the backend API to not return an error for those kind of failures (e.g. migration id is not found, etc.) and instead returns a payload that has a `success` and `failureReason` that will be set in case of a failure.

We'll need to be careful about doing that, as theoretically it's a breaking change to the current behaviour (albeit for an API we only just launched). But we can discuss that in a separate follow-up issue.

src/Octoshift/Commands/AbortMigration/AbortCommandBase.cs
```diff
@@ -0,0 +1,57 @@
...
+
+[assembly: InternalsVisibleTo("OctoshiftCLI.Tests")]
+
+namespace OctoshiftCLI.Commands.AbortMigration;
+
+public class AbortMigrationCommandBase : CommandBase<AbortMigrationCommandArgs, AbortMigrationCommandHandler>
+{
+    public AbortMigrationCommandBase() : base(
+        name: "abort-migration",
+        description: "Aborts a migration that is in progress.")
```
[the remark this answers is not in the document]
timrogers: Did you deliberately not apply this suggestion? Just wanted to check, as I _think_ this description is more correct and more helpful.

timrogers (approved):
This is looking good to me - sorry that it took so long for me to clean up the blocking review 😓 

I have one question about the help text, but I'm happy for you to address that and then merge without getting another review from me.

src/Octoshift/Commands/AbortMigration/AbortCommandBase.cs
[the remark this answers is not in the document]
begonaguereca: So weird no it should have been applied. Done now

brianaj (approved):

timrogers (approved):

Issue #1077, opened 2023-07-21
Add an option to skip check for a new version.


tspascoal:
## Description

It would be nice to have an option to opt out of the version check, either with an environment variable or a `--skip-version-check` option.

This would allow skiping version checks when doing repo migrations at scale.

dylan-smith:
If we do this, we should still have the CLI output the current version to make life easy for support folks looking at customer logs.

timrogers:
Can you share a bit more information on why you want to skip the version check? Is it just to remove the overhead of the HTTP request?

PR #1178, opened 2023-12-06
Fix message when `GEI_SKIP_VERSION_CHECK` is turned on

timrogers:
Whilst testing the new feature introduced in #1077 to skip the "latest version" check, I spotted that the log message we emit when the check is skipped is incorrect. This fixes it.


tspascoal (approved):
🚀

timrogers:
I just built releases and tested this end to end locally, and this is all looking good ❤️ Once the tests pass, I'll merge and cut a new release, v1.6.0.

PR #1181, opened 2023-12-06
Created `AbortMigrationResult` class and surfacing helpful errors

begonaguereca:
This is one of the [error scenarios](https://github.ghe.com/github/octoshift/issues/7842) that was noted for the Abort migration command. We are covering when a migration id is malformed.

The other scenario, covering when a migration is trying to be aborted but it's in a completed state, will require further discussion as to where the error should be surfaced. Either here or in the backend! 

- [x] Issue linked

Closes: https://github.ghe.com/github/octoshift/issues/7867


src/Octoshift/Services/GithubApi.cs
```diff
@@ -1011,6 +1011,47 @@ public virtual async Task<string> GetEnterpriseServerVersion()
...
+        {
+            query = $"{query} {{ {gql} }}",
+            variables = new
+            {
+                migrationId = migrationSourceId,
+            },
+            operationName = "abortRepositoryMigration"
+        };
+        var response = await _client.PostGraphQLAsync(url, payload);
+        var data = JObject.Parse(response);
```
begonaguereca, line 1039: here

src/Octoshift/Services/GithubApi.cs
```diff
@@ -1011,6 +1012,41 @@ public virtual async Task<string> GetEnterpriseServerVersion()
         return (string)data["installed_version"];
     }
 
+    public virtual async Task<bool> AbortMigration(string migrationId)
```
begonaguereca, line 1015: Moving up a public class

src/OctoshiftCLI.Tests/Octoshift/Services/GithubApiTests.cs
```diff
@@ -3368,12 +3368,52 @@ public async Task AbortMigration_Returns_True_On_Success()
...
+
+        var payload = new
+        {
+            query = $"{query} {{ {gql} }}",
+            variables = new
+            {
+                migrationId,
+            },
+            operationName = "abortRepositoryMigration"
+        };
```
ArinGhazarian, line 3402: there is no actual setup code here and this test will fail because of it. You need to setup the `_githubClientMock` to throw a `OctoshiftCliException` with the desired message:

```c#
_githubClientMock
    .Setup(m => m.PostGraphQLAsync(It.IsAny<string>(), It.IsAny<object>(), null))
    .ThrowsAsync(new OctoshiftCliException("Could not resolve to a node"));
```

src/OctoshiftCLI.Tests/Octoshift/Services/GithubApiTests.cs
```diff
@@ -3368,12 +3368,52 @@ public async Task AbortMigration_Returns_True_On_Success()
...
+
+        const string actualErrorResponse = $"Invalid migration id: {migrationId}";
+
+        // Act
+        await _githubApi.AbortMigration(migrationId);
```
ArinGhazarian, line 3407: This won't work here, whenever a test is testing for an exception being thrown the correct way to test is the way you did below so this like can be safely deleted it not it will throw the actual exception and the test will fail.

src/OctoshiftCLI.Tests/Octoshift/Services/GithubApiTests.cs
```diff
@@ -3368,12 +3368,52 @@ public async Task AbortMigration_Returns_True_On_Success()
...
+        {
+            query = $"{query} {{ {gql} }}",
+            variables = new
+            {
+                migrationId,
+            },
+            operationName = "abortRepositoryMigration"
+        };
+
+        const string actualErrorResponse = $"Invalid migration id: {migrationId}";
```
ArinGhazarian, line 3404: nit: this is the expected message not actual a more appropriate name would be: `expectedErrorMessage`.

ArinGhazarian (changes requested):

src/OctoshiftCLI.Tests/Octoshift/Services/GithubApiTests.cs
begonaguereca: There are [several](https://github.com/github/gh-gei/blob/142c9f8246aef888488283d11d4bc50fe938f85a/src/OctoshiftCLI.Tests/Octoshift/Services/GithubApiTests.cs#L88) of other examples in the Tests file that do this, why does it not work in this particular scenario?

src/OctoshiftCLI.Tests/Octoshift/Services/GithubApiTests.cs
```diff
@@ -3368,12 +3368,54 @@ public async Task AbortMigration_Returns_True_On_Success()
...
+
+        var payload = new
+        {
+            query = $"{query} {{ {gql} }}",
+            variables = new
+            {
+                migrationId,
+            },
+            operationName = "abortRepositoryMigration"
+        };
```
ArinGhazarian, line 3402: These lines are redundant and need to be deleted.
```suggestion
```

src/OctoshiftCLI.Tests/Octoshift/Services/GithubApiTests.cs
```diff
@@ -3368,12 +3368,54 @@ public async Task AbortMigration_Returns_True_On_Success()
...
+
+        const string expectedErrorMessage = $"Invalid migration id: {migrationId}";
+
+        // Act
+        _githubClientMock
+            .Setup(m => m.PostGraphQLAsync(It.IsAny<string>(), It.IsAny<object>(), null))
+            .ThrowsAsync(new OctoshiftCliException("Could not resolve to a node"));
+
+        // Assert
+
```
ArinGhazarian, line 3412: the `_githubClientMock` is technically still part of the arrange, in this case act and assert are both happening in the invoke like:
```suggestion
        _githubClientMock
            .Setup(m => m.PostGraphQLAsync(It.IsAny<string>(), It.IsAny<object>(), null))
            .ThrowsAsync(new OctoshiftCliException("Could not resolve to a node"));

        // Act, Assert
```

ArinGhazarian (commented):
Just a couple of quick fixes but the rest looks good. You can also change the target branch to main since this was already merged into main. I manually ran CI against this branch though --> https://github.com/github/gh-gei/actions/runs/7173936935

ArinGhazarian (approved):

Issue #1094, opened 2023-08-07
Add support for `--target-api-url` argument across all commands


timrogers:
All CLI commands which interact with a GitHub migration target should support the `--target-api-url` argument so that you can work with a target other than GitHub.com.

This is already supported in `gh gei migrate-repo`, but there are gaps in many (honestly, most) other commands.

jessicaliu100:
.tracktocat cross-link github/octoshift

brianaj:
Closing out as per notes in https://github.ghe.com/github/octoshift/issues/7956

PR #1214, opened 2024-01-29
`--target-api-url` support for migrate repo and org

brianaj:
Part of https://github.com/github/gh-gei/issues/1094 and https://github.ghe.com/github/octoshift/issues/7956

Adds full support for `migrate-repo` (gei already implemented) and `migrate-org` commands to support `--target-api-url`.

Functional tested for all the above against EU. Details in https://github.ghe.com/github/octoshift/issues/7956#issuecomment-8112051.

- [ ] Issue linked

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -1,1 +1,1 @@
-- Add `--target-api-url` to `ado2gh migrate repo`, `bbs2gh migrate repo`, and `gei migrate org` to support newer GitHub migration paths.
+- Add `--target-api-url` to `gh ado2gh migrate-repo`, `gh bbs2gh migrate-repo`, and `gh gei migrate-org` to support newer GitHub migration paths.
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -1,1 +1,1 @@
-
+- Add `--target-api-url` to `ado2gh migrate repo`, `bbs2gh migrate repo`, and `gei migrate org` to support newer GitHub migration paths.
--- a/src/OctoshiftCLI.Tests/ado2gh/Commands/MigrateRepo/MigrateRepoCommandHandlerTests.cs
+++ b/src/OctoshiftCLI.Tests/ado2gh/Commands/MigrateRepo/MigrateRepoCommandHandlerTests.cs
@@ -23,6 +23,7 @@ public class MigrateRepoCommandHandlerTests
     private const string ADO_REPO = "foo-repo";
     private const string GITHUB_ORG = "foo-gh-org";
     private const string GITHUB_REPO = "gh-repo";
+    private const string TARGET_API_URL = "https://api.github.com";
     private readonly string ADO_REPO_URL = $"https://dev.azure.com/{ADO_ORG}/{ADO_TEAM_PROJECT}/_git/{ADO_REPO}";
     private readonly string ADO_TOKEN = Guid.NewGuid().ToString();
     private readonly string GITHUB_ORG_ID = Guid.NewGuid().ToString();
@@ -83,6 +84,7 @@ public class MigrateRepoCommandHandlerTests
             AdoRepo = ADO_REPO,
             GithubOrg = GITHUB_ORG,
             GithubRepo = GITHUB_REPO,
+            TargetApiUrl = TARGET_API_URL,
             QueueOnly = true,
         };
         await _handler.Handle(args);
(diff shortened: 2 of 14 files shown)


src/OctoshiftCLI.Tests/ado2gh/Commands/MigrateRepo/MigrateRepoCommandHandlerTests.cs
```diff
@@ -23,6 +23,7 @@ public class MigrateRepoCommandHandlerTests
     private const string ADO_REPO = "foo-repo";
     private const string GITHUB_ORG = "foo-gh-org";
     private const string GITHUB_REPO = "gh-repo";
+    private const string TARGET_API_URL = "https://api.github.com";
```
ArinGhazarian, line 26: _I will add a comment just for this line but it applies to all tests for all CLIs._ 

There is no need to add the target api url here because it's only going to be used inside the `Command` (not the handler) to create an instance of `GithubApi`. Instead a new and very simple test needs to be added to the Command tests (e.g. `MigrateOrgCommandTests.cs`) to make sure if the `--target-api-url` is provided, the factory is going to receive it.

ArinGhazarian (commented):
The code looks good. 
For the tests, I don't think you need to add the target api url to the command handler tests but rather you need to add a simple test for the handers themselves to make sure that github api factory is getting invoked with the target api url.

src/OctoshiftCLI.Tests/ado2gh/Commands/MigrateRepo/MigrateRepoCommandHandlerTests.cs
ArinGhazarian: The new tests are going to be similar to [this one](https://github.com/github/gh-gei/blob/c5c54d27fe71ec8c4b40c83fca996e1697050d73/src/OctoshiftCLI.Tests/ado2gh/Commands/MigrateRepo/MigrateRepoCommandTests.cs#L54)

src/OctoshiftCLI.Tests/gei/Commands/MigrateOrg/MigrateOrgCommandTests.cs
```diff
@@ -21,6 +21,7 @@ public void Should_Have_Options()
             TestHelpers.VerifyCommandOption(command.Options, "github-source-pat", false);
             TestHelpers.VerifyCommandOption(command.Options, "github-target-pat", false);
             TestHelpers.VerifyCommandOption(command.Options, "verbose", false);
+            TestHelpers.VerifyCommandOption(command.Options, "target-api-url", false);
```
ArinGhazarian, line 24: A similar test to `It_Uses_Target_Api_Url_When_Provided` also needs to be added for this command too.

RELEASENOTES.md
```diff
@@ -1 +1 @@
-
+- Add `--target-api-url` to `ado2gh migrate repo`, `bbs2gh migrate repo`, and `gei migrate org` to support newer GitHub migration paths.
```
ArinGhazarian, line 1:
```suggestion
- Add `--target-api-url` to `gh ado2gh migrate-repo`, `gh bbs2gh migrate-repo`, and `gh gei migrate-org` to support newer GitHub migration paths.
```

ArinGhazarian (commented):
It looks great. There is still a missing test though 😃

src/OctoshiftCLI.Tests/gei/Commands/MigrateOrg/MigrateOrgCommandTests.cs
brianaj: okay will add tomorrow, thanks for the quick feedback!

brianaj: @ArinGhazarian added in https://github.com/github/gh-gei/pull/1214/commits/833931b72b39a5e1dd253973774f603db509a397, also note I was following the gei spec which originally added this flag, hence why I added test in handler to begin with but thanks for catching it 👍🏾  https://github.com/github/gh-gei/blob/main/src/OctoshiftCLI.Tests/gei/Commands/MigrateRepo/MigrateRepoCommandTests.cs

ArinGhazarian: Yeah a lot has changed in the CLI since then mostly because the underlying `System.CommandLine` nuget package pas early beta. Later on as the package became more mature, we changed the way we created the handlers to keep the handles' constructor very simple and easy to test and also keep the concerns as separate as possible so for each command we have the arg, the command and the handler, each one doing a specific thing. IMO the most interesting thing we did was to not inject factories into handlers and delegate the creation of a handler to the command via `BuildHanlder` and that's where all the creation logic goes.

ArinGhazarian (approved):

PR #1241, opened 2024-05-03
Add support for migrating secret scanning resolution comments

### LINKED ISSUE #1240: [gei] Add support for migrating secret scanning resolution comments
## Description

This feature would allow users to migrate not only the state and resolution of their secret scanning alerts but also the associated resolution comment.
As the comment might include information very important for an organization, especially for traceability purposes, this resolution comment might be a mandatory piece of information to migrate.
Information like the date the resolution happened and who resolved it would still be lost, but this could potentially be prefixed to the migrated comment if needed, but maybe over the top.

For now this feature would be to simply support for the resolution comment as is.

dlinares-linux:
Added support for migrating secret scanning resolution comments

- [x] Issue linked

Fixes #1240


src/Octoshift/Services/GithubApi.cs
```diff
@@ -884,7 +884,14 @@ public virtual async Task UpdateSecretScanningAlert(string org, string repo, int
...
         var url = $"{_apiUrl}/repos/{org.EscapeDataString()}/{repo.EscapeDataString()}/secret-scanning/alerts/{alertNumber}";
 
-        object payload = state == SecretScanningAlert.AlertStateOpen ? new { state } : new { state, resolution };
+        var payload = state == SecretScanningAlert.AlertStateOpen
+            ? (new { state })
+            : (object)(new
+            {
+                state,
+                resolution,
+                resolution_comment = resolutionComment ?? string.Empty
```
ArinGhazarian, line 893: Why is it being set to empty string? If nothing is passed it makes sense for the default to be `null`.
```suggestion
                resolution_comment = resolutionComment
```

src/OctoshiftCLI.Tests/Octoshift/Services/GithubApiTests.cs
```diff
@@ -2782,6 +2789,30 @@ public async Task UpdateSecretScanningAlert_Calls_The_Right_Endpoint_With_Payloa
         _githubClientMock.Verify(m => m.PatchAsync(url, It.Is<object>(x => x.ToJson() == payload.ToJson()), null));
     }
 
+
+    [Fact]
+    public async Task UpdateSecretScanningAlert_Replaces_Null_Resolution_Comment_With_Empty_String()
```
ArinGhazarian, line 2794: Based on my earlier comment we probably won't need this test.

src/OctoshiftCLI.Tests/Octoshift/Services/SecretScanningAlertServiceTests.cs
```diff
@@ -358,15 +369,86 @@ public async Task Migrates_Multiple_Alerts()
...
             300,
             SecretScanningAlert.AlertStateResolved,
-            SecretScanningAlert.ResolutionFalsePositive)
+            SecretScanningAlert.ResolutionFalsePositive,
+            "False positive explanation comment")
+        );
+    }
+
+    [Fact]
+    public async Task One_Secret_Updated_With_No_Resolution_Comment()
```
ArinGhazarian, line 387: Not sure if this adds value, seems like a bit of an overkill IMO.

ArinGhazarian (commented):
Thanks for the contribution. Had a few questions but looks good overall.

src/Octoshift/Services/GithubApi.cs
dlinares-linux: Hello, thanks for the comments. 
Added it that way to align with what is done for the dismissedComment for code scanning alerts (cf. https://github.com/github/gh-gei/blob/main/src/Octoshift/Services/GithubApi.cs#L933 )
```csharp
       var payload = state == "open"
            ? (new { state })
            : (object)(new
            {
                state,
                dismissed_reason = dismissedReason,
                dismissed_comment = dismissedComment ?? string.Empty
            });
        await _client.PatchAsync(url, payload);
```

Your suggestion makes sense though, so would you prefer me to do that in a separate pull request to align both changes, or change both it in this pull request?

src/OctoshiftCLI.Tests/Octoshift/Services/GithubApiTests.cs
dlinares-linux: Agreed. 
And if we align with code scanning, we probably don't need that test either:  [UpdateCodeScanningAlert_Replaces_Null_Dismissed_Comment_With_Empty_String](https://github.com/github/gh-gei/blob/2fc08d1dcc025eac239487c38692c22657a41a27/src/OctoshiftCLI.Tests/Octoshift/Services/GithubApiTests.cs#L3118)

As per previous comment, happy to do a separate pull request to align both code scanning and secret scanning, or do the changes in this one.

src/Octoshift/Services/GithubApi.cs
ArinGhazarian: 👋 ok got it. So I guess as you said the other one needs to be changed too. I think it's better to do that one in a separate PR.

src/OctoshiftCLI.Tests/Octoshift/Services/SecretScanningAlertServiceTests.cs
dlinares-linux: What would be your preferred option:
1. remove completely that test. This would align with the current tests for code scanning where a dismissedComment is always set (so we are not testing the use case where no comments were set)
2. rework both code scanning and secret scanning tests to "squeeze" one alert with no comment (to exercise the code, but not as a separate test case).
3. keep that test separate and add a similar overkill test in code scanning 😄

ArinGhazarian: So I am leaning towards option 1, since the way I think about this is a scenario where no comments are set still means we're setting it but to a `null` value and we have test coverage that shows we indeed use that value.

src/Octoshift/Services/GithubApi.cs
dlinares-linux: Understood. I have opened #1248 and added your suggestions there. Thanks again!

src/OctoshiftCLI.Tests/Octoshift/Services/GithubApiTests.cs
dlinares-linux: I have opened #1248 and added your suggestions there. Thanks again!
Happy to work on it next.

PR #1261, opened 2024-09-05
Add Support for Archive Uploads to GitHub-Owned Storage with Multipart Upload Option

hfishback01:
### Summary:

This PR introduces the ability for the GEI CLI to upload archives to GitHub-owned storage using multipart uploads. To enable this, a new CLI option, --use-github-storage, has been added. This option allows users to explicitly specify that their archives should be uploaded to GitHub's managed storage instead of their own.

### **Key Features:**
Multipart Upload Support: Leveraging multipart upload for reliable and efficient large file uploads. This ensures the robustness of the uploads, minimizing the chances of failure during transmission. We're investigating libraries to handle multipart uploads, but for now, this feature is not rolled by hand unless necessary.

New Option: --use-github-storage:

This option is required to upload archives to GitHub-owned storage and will be hidden until GitHub-owned storage reaches general availability (GA).
By default, archives are still uploaded to user-specified storage locations unless the --use-github-storage flag is explicitly set.

### **Motivation:**
As discussed during our EDR, we want to allow users to utilize GitHub-owned storage as an option without making it the default. This approach lets us roll out GitHub-managed storage in stages and gather feedback while providing an explicit way to opt-in to this feature.

Check in BBS MigrateRepoCommandArgs removed, relating to changes in this PR: https://github.com/github/gh-gei/pull/1057

Take it out for a 🚗 :

**Migrating from GitHub Enterprise Server:**
```
dotnet run --project src/gei/gei.csproj -- migrate-repo --github-source-org valet-testing --source-repo integration-tests --github-target-org octoshift-staging --target-repo YOU_ADD_NEW_NAME--use-github-storage true --ghes-api-url https://octoshift-ghe.westus2.cloudapp.azure.com/api/v3 --verbose
```
Successful migration_id: `RM_kgHaACQxMWNjY2RlMi1hNTAzLTQwMWItOTA1OS1kM2NkZjBiYjM4Y2Q`

Closes: https://github.ghe.com/github/octoshift/issues/8969

- [ ] Issue linked

### DIFF
--- a/src/Octoshift/Services/GithubApi.cs
+++ b/src/Octoshift/Services/GithubApi.cs
@@ -1,14 +1,10 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.CommandLine;
 using System.IO;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
-using System.Net.Http.Headers;
 using System.Threading.Tasks;
-using Azure;
-using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using Octoshift.Models;
 using OctoshiftCLI.Extensions;
--- a/src/Octoshift/Services/GithubApi.cs
+++ b/src/Octoshift/Services/GithubApi.cs
@@ -1,10 +1,14 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.CommandLine;
 using System.IO;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
+using System.Net.Http.Headers;
 using System.Threading.Tasks;
+using Azure;
+using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using Octoshift.Models;
 using OctoshiftCLI.Extensions;
@@ -1049,19 +1053,28 @@ public class GithubApi
 
     public virtual async Task<string> UploadArchiveToGithubStorage(string org, bool isMultipart, string archiveName, Stream archiveContent)
     {
-        var multipartRoute = isMultipart ? "/blobs/uploads" : "";
+        using HttpContent httpContent = new StreamContent(archiveContent);
+        string response;
 
-        var url = $"{_apiUrl}/organizations/{org.EscapeDataString()}/gei/archive{multipartRoute}";
-
-        var payload = new
+        if (isMultipart)
         {
-            archive_filename = archiveName,
-            archive_Content = archiveContent
-        };
+            var url = $"{_apiUrl}/organizations/{org.EscapeDataString()}/gei/archive/blobs/uploads";
 
-        var response = await _client.PostAsync(url, payload);
+            using var content = new MultipartFormDataContent
+            {
+                { httpContent, "archive", archiveName }
+            };
+
+            response = await _client.PostAsync(url, content);
+        }
+        else
+        {
+            var url = $"{_apiUrl}/organizations/{org.EscapeDataString()}/gei/archive";
+
+            response = await _client.PostAsync(url, httpContent);
+        }
+
         var data = JObject.Parse(response);
-
         return "gei://archive/" + (string)data["archiveId"];
     }
 
--- a/src/Octoshift/Services/GithubApi.cs
+++ b/src/Octoshift/Services/GithubApi.cs
@@ -1049,12 +1049,12 @@ public class GithubApi
 
     public virtual async Task<string> UploadArchiveToGithubStorage(string org, bool isMultipart, string archiveName, Stream archiveContent)
     {
-        using HttpContent httpContent = new StreamContent(archiveContent);
+        using var httpContent = new StreamContent(archiveContent);
         string response;
 
         if (isMultipart)
         {
-            var url = $"{_apiUrl}/organizations/{org.EscapeDataString()}/gei/archive/blobs/uploads";
+            var url = $"https://uploads.github.com/organizations/{org.EscapeDataString()}/gei/archive/blobs/uploads";
 
             using var content = new MultipartFormDataContent
             {
@@ -1065,7 +1065,7 @@ public class GithubApi
         }
         else
         {
-            var url = $"{_apiUrl}/organizations/{org.EscapeDataString()}/gei/archive";
+            var url = $"https://uploads.github.com/organizations/{org.EscapeDataString()}/gei/archive?name={archiveName}";
 
             response = await _client.PostAsync(url, httpContent);
         }
--- a/src/Octoshift/Services/GithubApi.cs
+++ b/src/Octoshift/Services/GithubApi.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
@@ -759,7 +760,6 @@ public class GithubApi
         });
     }
 
-
     public virtual async Task<string> GetUserId(string login)
     {
         var url = $"{_apiUrl}/graphql";
@@ -1045,6 +1045,24 @@ public class GithubApi
         {
             throw new OctoshiftCliException($"Invalid migration id: {migrationId}", ex);
         }
+    }
+
+    public virtual async Task<string> UploadArchiveToGithubStorage(string org, bool isMultipart, string archiveName, Stream archiveContent)
+    {
+        var multipartRoute = isMultipart ? "/blobs/uploads" : "";
+
+        var url = $"{_apiUrl}/organizations/{org.EscapeDataString()}/gei/archive{multipartRoute}";
+
+        var payload = new
+        {
+            archive_filename = archiveName,
+            archive_Content = archiveContent
+        };
+
+        var response = await _client.PostAsync(url, payload);
+        var data = JObject.Parse(response);
+
+        return "gei://archive/" + (string)data["archiveId"];
     }
 
     private static object GetMannequinsPayload(string orgId)
--- a/src/Octoshift/Services/GithubApi.cs
+++ b/src/Octoshift/Services/GithubApi.cs
@@ -1065,8 +1065,8 @@ public class GithubApi
         }
         else
         {
-            var url = $"https://uploads.github.com/organizations/{org.EscapeDataString()}/gei/archive?name={archiveName}";
-
+            var url = $"https://uploads.github.com/organizations/{org.EscapeDataString()}/gei/archive\\?name\\={archiveName}";
+            // DEV: var url = $"http://uploads.github.localhost/organizations/{org.EscapeDataString()}/gei/archive?name={archiveName}"
             response = await _client.PostAsync(url, httpContent);
         }
 
--- a/src/Octoshift/Services/GithubApi.cs
+++ b/src/Octoshift/Services/GithubApi.cs
@@ -1071,14 +1071,14 @@ public class GithubApi
         }
     }
 
-    public virtual async Task<string> UploadArchiveToGithubStorage(string org, bool isMultipart, string archiveName, Stream archiveContent)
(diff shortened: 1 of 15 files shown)


src/bbs2gh/Commands/MigrateRepo/MigrateRepoCommandArgs.cs
```diff
@@ -92,11 +93,6 @@ public override void Validate(OctoLogger log)
 
     private void ValidateNoGenerateOptions()
     {
-        if (BbsUsername.HasValue() || BbsPassword.HasValue())
```
hfishback01, line 95: Removing this, I don't think it should be checked for with [this change](https://github.com/github/gh-gei/pull/1057)

src/OctoshiftCLI.Tests/bbs2gh/Commands/MigrateRepo/MigrateRepoCommandArgsTests.cs
```diff
@@ -186,44 +206,6 @@ public void It_Throws_When_Kerberos_Is_Set_And_Bbs_Username_Is_Provided()
                 .WithMessage("*--bbs-username*--kerberos*");
         }
 
-        [Fact]
-        public void Errors_If_BbsServer_Url_Not_Provided_But_Bbs_Username_Is_Provided()
```
hfishback01, line 190: Removing these tests as bbs-server-url is now required.

src/bbs2gh/Commands/MigrateRepo/MigrateRepoCommandHandler.cs
```diff
@@ -198,6 +200,22 @@ private async Task<string> UploadArchiveToAws(string bucketName, string archiveP
         return archiveBlobUrl;
     }
 
+    private async Task<string> UploadArchiveToGithub(string org, string archivePath)
+    {
+#pragma warning disable IDE0063
```
ArinGhazarian, line 205: Is there a reason the warning was disabled? I think we can use `using declaration` here.

src/OctoshiftCLI.Tests/bbs2gh/Commands/MigrateRepo/MigrateRepoCommandHandlerTests.cs
```diff
@@ -343,6 +343,59 @@ public async Task Happy_Path_Full_Flow_Bbs_Credentials_Via_Environment()
...
+                UseGithubStorage = true,
+                GithubOrg = GITHUB_ORG,
+                GithubRepo = GITHUB_REPO,
+                GithubPat = GITHUB_PAT,
+                QueueOnly = true,
+            };
+            await _handler.Handle(args);
+
+            File.Delete(archiveFilePath);
+            gitContentStream.Close();
```
ArinGhazarian, line 385: even though this is a unit test, but we should still properly dispose the stream with a `using` statement, you can use a using declaration when creating the stream, that way there will be no need to close the stream as it will be closed automatically.

src/OctoshiftCLI.Tests/gei/Commands/MigrateRepo/MigrateRepoCommandHandlerTests.cs
```diff
@@ -325,6 +325,108 @@ public async Task Happy_Path_GithubSource_Ghes()
...
+            _mockTargetGithubApi.Verify(x => x.GetMigration(migrationId));
+            _mockTargetGithubApi.Verify(x => x.UploadArchiveToGithubStorage(It.IsAny<string>(), false, It.IsAny<string>(), gitContentStream));
+            _mockTargetGithubApi.Verify(x => x.UploadArchiveToGithubStorage(It.IsAny<string>(), false, It.IsAny<string>(), metaContentStream));
+            _mockFileSystemProvider.Verify(x => x.DeleteIfExists(gitArchiveFilePath), Times.Once);
+            _mockFileSystemProvider.Verify(x => x.DeleteIfExists(metadataArchiveFilePath), Times.Once);
+
+            File.Delete(gitArchiveFilePath);
+            File.Delete(metadataArchiveFilePath);
+            gitContentStream.Close();
+            metaContentStream.Close();
```
ArinGhazarian, line 427: same concerns here too, please dispose the disposables properly.

src/bbs2gh/Commands/MigrateRepo/MigrateRepoCommandHandler.cs
```diff
@@ -198,6 +200,22 @@ private async Task<string> UploadArchiveToAws(string bucketName, string archiveP
...
+        await using (var archiveData = _fileSystemProvider.OpenRead(archivePath))
+#pragma warning restore IDE0063
+        {
+            var isMultipart = archiveData.Length > 100 * 1024 * 1024; ; // Determines if stream size is greater than 100MB
```
ArinGhazarian, line 209: two things:
1. could you use a constant for the size limit, for example `STREAM_SIZE_LIMIT`
2. there is an extra semicolon :)

src/gei/Commands/MigrateRepo/MigrateRepoCommand.cs
```diff
@@ -139,6 +140,11 @@ public MigrateRepoCommand() : base(
             Description = "Keeps the archive on this machine after uploading to the blob storage account. Only applicable for migrations from GitHub Enterprise Server versions before 3.8.0."
         };
 
+        public Option<bool> UseGithubStorage { get; } = new("--use-github-storage")
+        {
+            IsHidden = true,
+            Description = "Enables multipart upload of a user-specified filepath to Github-owned storage for use during migration",
```
ArinGhazarian, line 146:
```suggestion
            Description = "Enables multipart upload of a user-specified file path to GitHub-owned storage for use during migration",
```

src/gei/Commands/MigrateRepo/MigrateRepoCommandHandler.cs
```diff
@@ -204,12 +206,14 @@ private string ExtractGhesBaseUrl(string ghesApiUrl)
...
       string githubSourceOrg,
+      string githubTargetOrg,
       string sourceRepo,
       string awsBucketName,
       bool skipReleases,
       bool lockSourceRepo,
       bool blobCredentialsRequired,
-      bool keepArchive)
+      bool keepArchive,
+      bool UseGithubStorage)
```
ArinGhazarian, line 216: input arg should be camel cased

```suggestion
      bool useGithubStorage)
```

src/gei/Commands/MigrateRepo/MigrateRepoCommandHandler.cs
```diff
@@ -307,6 +313,21 @@ private void DeleteArchive(string path)
         return (authenticatedGitArchiveUri.ToString(), authenticatedMetadataArchiveUri.ToString());
     }
 
+    private async Task<(string, string)> UploadArchivesToGithub(string org, string gitArchiveUploadFileName, Stream gitArchiveContent, string metadataArchiveUploadFileName, Stream metadataArchiveContent)
+    {
+        var isMultipart = gitArchiveContent.Length > 100 * 1024 * 1024; ; // Determines if stream size is greater than 100MB
```
ArinGhazarian, line 318: same concerns here too, create a constant and there is an extra semicolon

src/gei/Commands/MigrateRepo/MigrateRepoCommandHandler.cs
```diff
@@ -307,6 +313,21 @@ private void DeleteArchive(string path)
...
+
+        _log.LogInformation($"Uploading git archive to GitHub Storage");
+        var uploadedGitArchiveUrl = await _targetGithubApi.UploadArchiveToGithubStorage(org, isMultipart, gitArchiveUploadFileName, gitArchiveContent);
+
+        isMultipart = metadataArchiveContent.Length > 100 * 1024 * 1024; ; // Determines if stream size is greater than 100MB
```
ArinGhazarian, line 323: same concerns here too.

src/bbs2gh/Commands/MigrateRepo/MigrateRepoCommandArgs.cs
ArinGhazarian: the check should probably still exist but the error message needs to be changed because this is validating the args for a no generate scenario where the cli doesn't generate the archive and the archive path or url is is provided.
That being said I'd recommend revising the error messages in a separate PR because there is more than just this one, for example the one on line 105 says "SSH or SMB download options can only be provided with --bbs-server-url." but based on the same logic the server url should always be provided.

src/Octoshift/Services/GithubApi.cs
```diff
@@ -1047,6 +1047,33 @@ mutation abortRepositoryMigration(
...
+        if (isMultipart)
+        {
+            var url = $"{_apiUrl}/organizations/{org.EscapeDataString()}/gei/archive/blobs/uploads";
+
+            using var content = new MultipartFormDataContent
+            {
+                { httpContent, "archive", archiveName }
+            };
+
+            response = await _client.PostAsync(url, content);
```
ArinGhazarian, line 1064: A couple of observations:
1. Even though this may work the `PostAsync` method will convert the data to JSON (basically base64 string) and will eventually create `StringContent` so I am not sure if this is the indented use here because we're now dealing with a `Stream` rather than string content. So We may want to refactor the [SendAsync](https://github.com/github/gh-gei/blob/62960fcd0f1fc00ef20e380ed52673fc8b41d47b/src/Octoshift/Services/GithubClient.cs#L140) method to also support a `MultiPartFormDataContent` and `StreamContent`. It can simply check to see if the passed in `body` is either of those and not convert it to a string content.
2. When dealing with a stream we shouldn't log the entire body as we do [here](https://github.com/github/gh-gei/blob/62960fcd0f1fc00ef20e380ed52673fc8b41d47b/src/Octoshift/Services/GithubClient.cs#L154) instead as I suggested in option 1, we can check the body and if it's a multipart form data or a stream content we can just say `BLOB` or `Binary data` instead of dumping the entire JSON encoded body!

ArinGhazarian (commented):
Dropped a couple of more comments.

src/bbs2gh/Commands/MigrateRepo/MigrateRepoCommandArgs.cs
hfishback01: Since we now [require](https://github.com/github/gh-gei/pull/1057) the `bbs-server-url`, we always check for a username and password, even in the non-generate archive case.

`--bbs-server-url` enables the create of the bbsApiFactory, which then does the check for username and password.

src/Octoshift/Services/GithubApi.cs
```diff
@@ -1047,6 +1047,33 @@ mutation abortRepositoryMigration(
...
+        }
+        else
+        {
+            var url = $"https://uploads.github.com/organizations/{org.EscapeDataString()}/gei/archive\\?name\\={archiveName}";
+            // DEV: var url = $"http://uploads.github.localhost/organizations/{org.EscapeDataString()}/gei/archive?name={archiveName}"
+            response = await _client.PostAsync(url, httpContent);
+        }
+
+        var data = JObject.Parse(response);
+        return "gei://archive/" + (string)data["archiveId"];
```
begonaguereca, line 1074: here

src/Octoshift/Services/GithubApi.cs
```diff
@@ -1047,6 +1047,33 @@ mutation abortRepositoryMigration(
...
+            using var content = new MultipartFormDataContent
+            {
+                { httpContent, "archive", archiveName }
+            };
+
+            response = await _client.PostAsync(url, content);
+        }
+        else
+        {
+            var url = $"https://uploads.github.com/organizations/{org.EscapeDataString()}/gei/archive\\?name\\={archiveName}";
```
begonaguereca, line 1068: here

PR #1276, opened 2024-10-01
Add call to get org database id

begonaguereca:
This pull request introduces new methods to the `GithubApi` service and adds corresponding tests to ensure their functionality. 

We need this method that fetches the organization's database ID, so it can be used in part to upload archives to GitHub storage. 

We will need this for the GitHub owned storage support 

- [ ] Issue linked

### DIFF
--- a/src/Octoshift/Services/GithubApi.cs
+++ b/src/Octoshift/Services/GithubApi.cs
@@ -1082,36 +1082,6 @@ public class GithubApi
         }
     }
 
-<<<<<<< Updated upstream
-=======
-    public virtual async Task<string> UploadArchiveToGithubStorage(string org, bool isMultipart, string archiveName, Stream archiveContent)
-    {
-        using var httpContent = new StreamContent(archiveContent);
-        string response;
-
-        if (isMultipart)
-        {
-            var url = $"https://uploads.github.com/organizations/{org.EscapeDataString()}/gei/archive/blobs/uploads";
-
-            using var content = new MultipartFormDataContent
-            {
-                { httpContent, "archive", archiveName }
-            };
-
-            response = await _client.PostAsync(url, content);
-        }
-        else
-        {
-            var url = $"https://uploads.github.com/organizations/93741352/gei/archive?name={archiveName}";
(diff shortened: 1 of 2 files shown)


src/Octoshift/Services/GithubApi.cs
```diff
@@ -225,6 +225,41 @@ public virtual async Task<string> GetOrganizationId(string org)
...
+            variables = new { login = org }
+        };
+
+        try
+        {
+            return await _retryPolicy.Retry(async () =>
+            {
+                var data = await _client.PostGraphQLAsync(url, payload);
+
+                var databaseId = data["data"]["organization"]["databaseId"].ToString();
```
ArinGhazarian, line 252: We need to cast to a string rather than calling `ToString()` here because if in case of a `null` value `ToString()`  will through be a cast to a `string` won't and since we are not using nullable reference types the method can also return `null`:

```suggestion
                var databaseId = (string)data["data"]["organization"]["databaseId"];
```

ArinGhazarian (changes requested):

src/Octoshift/Services/GithubApi.cs
```diff
@@ -8,6 +8,7 @@
 using Octoshift.Models;
 using OctoshiftCLI.Extensions;
 using OctoshiftCLI.Models;
+using System.Diagnostics;
```
ArinGhazarian, line 11:
```suggestion
```

src/Octoshift/Services/GithubApi.cs
```diff
@@ -241,7 +266,8 @@ public virtual async Task<string> GetEnterpriseId(string enterpriseName)
             {
                 var data = await _client.PostGraphQLAsync(url, payload);
 
-                return (string)data["data"]["enterprise"]["id"];
+                var entireprise_id = (string)data["data"]["enterprise"]["id"];
+                return entireprise_id;
```
ArinGhazarian, line 270: Could you revert this as I guess it was created for debugging.

begonaguereca: 🧹 [a842577](https://github.com/github/gh-gei/pull/1276/commits/a8425779edd122246aebe73697c415819a438f33)

ArinGhazarian (approved):

PR #1285, opened 2024-10-22
Multipart uploads updated logic 

begonaguereca:
Closes https://github.ghe.com/github/octoshift/issues/9496!

- [ ] Issue linked

-------------------------------------------------------------------------------------------------------------
## Take it out for a 🚗 :

*** Remember to export env variables***
``` 
export GH_SOURCE_PAT=<<PAT>>
export GH_PAT=<<PAT>> 
```
**Migrating from GitHub Enterprise Server- Multi part upload:**

Sample repo to export: https://octoshift-ghe.westus2.cloudapp.azure.com/github-owned-storage/1-gib

```
dotnet run --project src/gei/gei.csproj -- migrate-repo --github-source-org github-owned-storage --source-repo large-1-gib --github-target-org import-testing --target-repo YOU_ADD_NEW_NAME --use-github-storage true --ghes-api-url https://octoshift-ghe.westus2.cloudapp.azure.com/api/v3 --verbose
```

Successful migration_id: RM_kgDaACQ0MzZmN2RjMS0wNjk4LTRlYjctYmMzNy0wMTAxNGRhOTQ3YTU


src/Octoshift/Services/GithubApi.cs
```diff
@@ -1089,24 +1090,102 @@ public virtual async Task<string> UploadArchiveToGithubStorage(string orgDatabas
...
+
+    private async Task<(string Response, KeyValuePair<string, IEnumerable<string>>[] Headers)> StartUploadAsync(string uploadUrl, string archiveName, long contentSize)
+    {
+        // Create the body for the request
+        var body = new JObject
+        {
+            ["content_type"] = "application/octet-stream",
+            ["name"] = archiveName,
+            ["size"] = contentSize
+        };
```
ArinGhazarian, line 1139: you can just create a dynamic object here:

```suggestion
        var body = new
        {
            content_type = "application/octet-stream",
            name = archiveName,
            size = contentSize
        };
```

src/Octoshift/Services/GithubApi.cs
```diff
@@ -1089,24 +1090,102 @@ public virtual async Task<string> UploadArchiveToGithubStorage(string orgDatabas
...
+            nextUrl = await UploadPartAsync(buffer, bytesRead, nextUrl.ToString());
+        }
+
+        // 3. Complete the upload
+        await CompleteUploadAsync(nextUrl.ToString());
+
+        return $"gei://archive/{guid}";
+    }
+
+    private async Task<(string Response, KeyValuePair<string, IEnumerable<string>>[] Headers)> StartUploadAsync(string uploadUrl, string archiveName, long contentSize)
```
ArinGhazarian, line 1131: We didn't add the `Async` suffix to the end of async methods (the only exception is the clients where we wanted the method names to be aligned with .NET's `HttpClient`)

```suggestion
    private async Task<(string Response, KeyValuePair<string, IEnumerable<string>>[] Headers)> StartUpload(string uploadUrl, string archiveName, long contentSize)
```

src/Octoshift/Services/GithubApi.cs
```diff
@@ -1089,24 +1090,102 @@ public virtual async Task<string> UploadArchiveToGithubStorage(string orgDatabas
...
+        // Set the correct Content-Type header (adjust based on API requirements)
+        content.Headers.ContentType = new System.Net.Http.Headers.MediaTypeHeaderValue("application/octet-stream");
+
+        await _client.PutAsync(lastUrl, content);
+    }
+
+    private Uri GetNextUrl(IEnumerable<KeyValuePair<string, IEnumerable<string>>> headers)
+    {
+        // Find the Location header
+        var locationHeader = headers.FirstOrDefault(header => header.Key == "Location");
```
ArinGhazarian, line 1172: Do we expect the header to be always present here? If yes then we should probably raise an exception or just use `First` so it will raise an exception if there header is not there.

src/Octoshift/Services/GithubApi.cs
```diff
@@ -1089,24 +1090,102 @@ public virtual async Task<string> UploadArchiveToGithubStorage(string orgDatabas
...
+
+        // Check if the Location header exists and has a value
+        if (locationHeader.Key != null)
```
ArinGhazarian, line 1175: This can raise a null reference error because of the `FirstOrDefault`.  If the methods definitely needs to return `null` I would also inverse the logic:
```c#
if (locationHeader is null)
{
    return null;
}
```

src/Octoshift/Services/GithubApi.cs
```diff
@@ -1089,24 +1090,102 @@ public virtual async Task<string> UploadArchiveToGithubStorage(string orgDatabas
...
+        {
+            var locationValue = locationHeader.Value.FirstOrDefault();
+
+            if (!string.IsNullOrEmpty(locationValue))
```
ArinGhazarian, line 1179: We do have a handy extension method for this:
```suggestion
            if (locationValue.HasValue())
```

src/Octoshift/Services/GithubApi.cs
```diff
@@ -1089,24 +1090,102 @@ public virtual async Task<string> UploadArchiveToGithubStorage(string orgDatabas
...
-        return (string)data["uri"];
+    private async Task<string> UploadMultipart(Stream archiveContent, string archiveName, string uploadUrl)
+    {
+        var buffer = new byte[_streamSizeLimit];
+
+        // 1. Start the upload
+        var (startResponse, startHeaders) = await StartUploadAsync(uploadUrl, archiveName, archiveContent.Length);
+
+        var nextUrl = GetNextUrl(startHeaders);
+        var guid = System.Web.HttpUtility.ParseQueryString(nextUrl.Query)["guid"];
```
ArinGhazarian, line 1116: I would add `System.Web` to the usings list to simplify this:
```suggestion
        var guid = HttpUtility.ParseQueryString(nextUrl.Query)["guid"];
```

src/Octoshift/Services/GithubApi.cs
```diff
@@ -1089,24 +1090,102 @@ public virtual async Task<string> UploadArchiveToGithubStorage(string orgDatabas
...
+
+        // Return the tuple with response and headers
+        return (response, headers);
+    }
+
+
+    private async Task<Uri> UploadPartAsync(byte[] body, int bytesRead, string nextUrl)
+    {
+        var content = new ByteArrayContent(body, 0, bytesRead);
+        content.Headers.ContentType = new System.Net.Http.Headers.MediaTypeHeaderValue("application/octet-stream");
```
ArinGhazarian, line 1152:
```suggestion
        content.Headers.ContentType = new ("application/octet-stream");
```

src/Octoshift/Services/GithubApi.cs
```diff
@@ -1089,24 +1090,102 @@ public virtual async Task<string> UploadArchiveToGithubStorage(string orgDatabas
...
+        var (response, headers) = await _client.PostTupleAsync(uploadUrl, body);
+
+        // Return the tuple with response and headers
+        return (response, headers);
+    }
+
+
+    private async Task<Uri> UploadPartAsync(byte[] body, int bytesRead, string nextUrl)
+    {
+        var content = new ByteArrayContent(body, 0, bytesRead);
```
ArinGhazarian, line 1151: because we're uploading the entire body and are not using the offset we can use the [other overload](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.bytearraycontent.-ctor?view=net-8.0#system-net-http-bytearraycontent-ctor(system-byte())) that simply accepts the `contents`:
```suggestion
        var content = new ByteArrayContent(body);
```

src/Octoshift/Services/GithubApi.cs
```diff
@@ -1089,24 +1090,102 @@ public virtual async Task<string> UploadArchiveToGithubStorage(string orgDatabas
...
-        var data = JObject.Parse(response);
-        return (string)data["uri"];
+    private async Task<string> UploadMultipart(Stream archiveContent, string archiveName, string uploadUrl)
+    {
+        var buffer = new byte[_streamSizeLimit];
+
+        // 1. Start the upload
+        var (startResponse, startHeaders) = await StartUploadAsync(uploadUrl, archiveName, archiveContent.Length);
+
+        var nextUrl = GetNextUrl(startHeaders);
```
ArinGhazarian, line 1115: if `nextUrl` returns a `null` value the next line will throw a null reference exception. So if null values are we should take it into account and make our code more resilient

src/Octoshift/Services/GithubApi.cs
```diff
@@ -1089,24 +1090,102 @@ public virtual async Task<string> UploadArchiveToGithubStorage(string orgDatabas
...
+
+    private async Task CompleteUploadAsync(string lastUrl)
+    {
+        // Create an empty HttpContent object
+        var content = new StringContent(string.Empty);
+
+        // Set the correct Content-Type header (adjust based on API requirements)
+        content.Headers.ContentType = new System.Net.Http.Headers.MediaTypeHeaderValue("application/octet-stream");
+
+        await _client.PutAsync(lastUrl, content);
```
ArinGhazarian, line 1166: can we just do a simple put here:

```c#
await _client.PustAsync(lastUrl);
```

src/Octoshift/Services/GithubClient.cs
```diff
@@ -76,6 +76,9 @@ public virtual async IAsyncEnumerable<JToken> GetAllAsync(
     public virtual async Task<string> PostAsync(string url, object body, Dictionary<string, string> customHeaders = null) =>
         (await SendAsync(HttpMethod.Post, url, body, customHeaders: customHeaders)).Content;
 
+    public virtual async Task<(string Content, KeyValuePair<string, IEnumerable<string>>[] ResponseHeaders)> PostTupleAsync(string url, object body, Dictionary<string, string> customHeaders = null) =>
```
ArinGhazarian, line 79: renaming suggestion: `PostWithFullResponseAsync`

src/Octoshift/Services/GithubClient.cs
```diff
@@ -140,6 +143,9 @@ public virtual async Task<string> PutAsync(string url, object body, Dictionary<s
     public virtual async Task<string> PatchAsync(string url, object body, Dictionary<string, string> customHeaders = null) =>
         (await SendAsync(HttpMethod.Patch, url, body, customHeaders: customHeaders)).Content;
 
+    public virtual async Task<(string Content, KeyValuePair<string, IEnumerable<string>>[] ResponseHeaders)> PatchTupleAsync(string url, object body, Dictionary<string, string> customHeaders = null) =>
```
ArinGhazarian, line 146: Renaming suggestion: `PatchWithFullResponseAsync`

src/Octoshift/Services/GithubApi.cs
ArinGhazarian: Even in that case there is probably no need to create a string content because it's supposed to be handled by the client so you can probably just send an empty string to put.

src/Octoshift/Services/GithubApi.cs
begonaguereca: Screenshot 2024-10-23 at 12 53 37 PM

begonaguereca: Its not letting me do that

src/Octoshift/Services/GithubApi.cs
begonaguereca: Ok I made the method First 👍🏽

src/Octoshift/Services/GithubApi.cs
begonaguereca:
```
  public virtual async Task<string> PutAsync(string url, object body, Dictionary<string, string> customHeaders = null) =>
        (await SendAsync(HttpMethod.Put, url, body, customHeaders: customHeaders)).Content;
```

This is the put command, should I make a new one then?

ArinGhazarian: no you can pass `null` for `body`

src/Octoshift/Services/GithubApi.cs
ArinGhazarian: that's because you need to add the appropriate `using` to the beginning of the cs file (`using OctoshiftCLI.Extensions`). It's an extension method that originally doesn't belong to the `string` class.

src/Octoshift/Services/GithubApi.cs
begonaguereca: I think we narrowed down that we do need this logic 👍🏽

src/Octoshift/Services/GithubApi.cs
begonaguereca: Ok so when I pass in `null` it will error out and not complete the upload. Adding this logic back in.

src/Octoshift/Services/MultipartUploaderService.cs
```diff
@@ -0,0 +1,129 @@
...
+
+        if (!string.IsNullOrEmpty(locationHeader.Key))
+        {
+            var locationValue = locationHeader.Value.FirstOrDefault();
+            if (locationValue.HasValue())
+            {
+                return new Uri(new Uri(_base_url), locationValue);
+            }
+        }
+        return null;
```
ArinGhazarian, line 127: I I would rewrite this method like the following:
```suggestion
        var locationHeader = headers.First(header => header.Key.Equals("Location", StringComparison.OrdinalIgnoreCase));

        if (locationHeader.Key.IsNullOrWhiteSpace() || locationHeader.Value.FirstOrDefault().IsNullOrWhiteSpace())
        {
            throw new OctoshiftCliException("Failed to retrieve the next URL for the upload part.");
        }

        return new Uri(new Uri(BASE_URL), locationHeader.Value.First());
```

src/Octoshift/Services/MultipartUploaderService.cs
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    public async Task<string> UploadMultipart(Stream archiveContent, string archiveName, string uploadUrl)
+    {
+        var buffer = new byte[_streamSizeLimit];
+
+        try
+        {
+            // 1. Start the upload
+            var startHeaders = await StartUpload(uploadUrl, archiveName, archiveContent.Length);
```
ArinGhazarian, line 30: I would add logs to indicate progress, like for example:
`Initiating upload with Upload URL: ... and Archive Name: ...`

src/Octoshift/Services/MultipartUploaderService.cs
```diff
@@ -0,0 +1,129 @@
...
+                throw new OctoshiftCliException("Failed to retrieve the next URL for the upload.");
+            }
+
+            var guid = HttpUtility.ParseQueryString(nextUrl.Query)["guid"];
+
+            // 2. Upload parts
+            int bytesRead;
+            while ((bytesRead = await archiveContent.ReadAsync(buffer, 0, buffer.Length)) > 0)
+            {
+                nextUrl = await UploadPart(buffer, bytesRead, nextUrl.ToString());
```
ArinGhazarian, line 44: an info log here would be good too to indicate progress. It's a good UX specially for larger archives. We can also calculate the number of chunks in advance to report a better progress:

`Uploading part 1 out of 100...`

src/Octoshift/Services/MultipartUploaderService.cs
```diff
@@ -0,0 +1,129 @@
...
+            }
+
+            // 3. Complete the upload
+            await CompleteUpload(nextUrl.ToString());
```
ArinGhazarian, line 48: also a log here to indicate the upload is done

`Finished uploading the archive.`

src/Octoshift/Services/MultipartUploaderService.cs
```diff
@@ -0,0 +1,129 @@
...
+            if (nextUrl == null)
+            {
+                throw new OctoshiftCliException("Failed to retrieve the next URL for the upload.");
+            }
+
+            var guid = HttpUtility.ParseQueryString(nextUrl.Query)["guid"];
+
+            // 2. Upload parts
+            int bytesRead;
+            while ((bytesRead = await archiveContent.ReadAsync(buffer, 0, buffer.Length)) > 0)
```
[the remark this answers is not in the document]
begonaguereca: Resolved 👍🏽

src/Octoshift/Services/MultipartUploaderService.cs
begonaguereca: Is there something wrong or inefficient with my logic?

src/Octoshift/Services/MultipartUploaderService.cs
begonaguereca: I've added the parts, Im not really sure what the URL or archive name will give people

src/Octoshift/Services/GithubApi.cs
```diff
@@ -1089,24 +1092,19 @@ public virtual async Task<string> UploadArchiveToGithubStorage(string orgDatabas
...
+            return response;
         }
         else
         {
-            var url = $"https://uploads.github.com/organizations/{orgDatabaseId.EscapeDataString()}/gei/archive?name={archiveName.EscapeDataString()}";
+            var url = $"{_base_url}/{orgDatabaseId.EscapeDataString()}/gei/archive?name={archiveName.EscapeDataString()}";
 
             response = await _client.PostAsync(url, streamContent);
+            var data = JObject.Parse(response);
+            return (string)data["uri"];
```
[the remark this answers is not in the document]
begonaguereca: Is there something incorrect or inefficient about my logic?

src/Octoshift/Services/MultipartUploaderService.cs
ArinGhazarian: The only main thing was to check for the `null` values and throwing an exception in the method. Aside from that my suggestion also simplified the logic a bit.

src/Octoshift/Services/GithubApi.cs
```diff
@@ -17,13 +17,14 @@ public class GithubApi
     private readonly GithubClient _client;
     private readonly string _apiUrl;
     private readonly RetryPolicy _retryPolicy;
-    internal int _streamSizeLimit = 100 * 1024 * 1024; // 100 MiB
+    private readonly ArchiveUploader _multipartUploader;
 
-    public GithubApi(GithubClient client, string apiUrl, RetryPolicy retryPolicy)
+    public GithubApi(GithubClient client, string apiUrl, RetryPolicy retryPolicy, ArchiveUploader multipartUploader)
```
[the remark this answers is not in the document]
begonaguereca: Passing in null now 👍🏽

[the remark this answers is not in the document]
ArinGhazarian: We're ultimately using the archive uploader is our `migrate-repo` commands. For example the [MigrateRepoCommandHandler](https://github.com/github/gh-gei/blob/550f4b859110c58e174888a00f7ae57167a66cce/src/bbs2gh/Commands/MigrateRepo/MigrateRepoCommandHandler.cs) for `bbs2gh`. What I mean is we can directly inject the `ArchiveUploader` into the command that we are eventually going to use.

src/Octoshift/Services/MultipartUploaderService.cs
```diff
@@ -0,0 +1,166 @@
...
+                throw new OctoshiftCliException("Failed to retrieve the next URL for the upload.");
+            }
+
+            var guid = HttpUtility.ParseQueryString(nextUrl.Query)["guid"];
+
+            // 2. Upload parts
+            int bytesRead;
+            var partsRead = 0;
+            var totalParts = (int)Math.Ceiling((double)archiveContent.Length / _streamSizeLimit);
+            while ((bytesRead = await archiveContent.ReadAsync(buffer.AsMemory(0, buffer.Length))) > 0)
```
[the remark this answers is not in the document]
begonaguereca: I am not sure why `archiveContent.ReadAsync(buffer)` does not work, all i know is that the logic does work how it is configured 👍🏽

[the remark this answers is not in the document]
ArinGhazarian: Converting `buffer` to a `Memory<T>` object is really unnecessary and an overhead here, if we really need to use the overload we can suppress the warning but only when we understand what's wrong with the overload that simply accepts the entire buffer. My suggestion is to work on the `ArchiveUploader` unit tests to test this logic and then try to change it back to  `archiveContent.ReadAsync(buffer)` so if there is any problem the unit tests should reveal it.

src/OctoshiftCLI.Tests/Octoshift/Services/GithubApiTests.cs
```diff
@@ -3491,20 +3498,36 @@ public async Task UploadArchiveToGithubStorage_Should_Upload_Stream_Content()
...
+            .Setup(m => m.PatchWithFullResponseAsync(It.IsAny<string>(), It.IsAny<HttpContent>(), null))
+            .ReturnsAsync((jsonResponse, new[]
+            {
+            new KeyValuePair<string, IEnumerable<string>>("Location", new[] { "/organizations/93741352/gei/archive/blobs/uploads?part_number=1&guid=123456&upload_id=i63n.35ClspPC5tSNp1rPcjm3FfKsnhLEPud627KummDTa29LSoFjIRlNjfzwJdax5sekyqJB7YPHtBZUCDCVwFUznod.p8XY_xHKu9FxFOcobs6keWC_9Xx8HFMQHxx" })
+            }));
+
+        // Mocking the final PUT request to complete the multipart upload
+        _githubClientMock
+            .Setup(m => m.PutAsync(It.IsAny<string>(), It.IsAny<HttpContent>(), null))
+            .ReturnsAsync(string.Empty);  // Ensure a non-null response
```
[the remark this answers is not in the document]
begonaguereca: Ok cool so im going to keep these tests here and just added tests for ArchiveUploader 👍🏽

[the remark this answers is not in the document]
ArinGhazarian: In the `GithubApi` the `ArchiveUploader` needs to be mocked which will simplify the test. So this test should change.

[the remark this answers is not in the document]
begonaguereca: ArchiveUploader is mocked at the top of the class, but the methods that actually do the http calls are private. 

I dont understand how else to create this test. I can add:

` var actualStringResponse = await _archiveUploader.Upload(archiveContent, archiveName, orgDatabaseId);` instead? But i feel like that test belongs in the other class and not here.

Is this test incorrect as it stands?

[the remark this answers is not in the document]
begonaguereca: [0cba519](https://github.com/github/gh-gei/pull/1285/commits/0cba51941fbab1490ec8ac6ade24b34a8513c6aa)

src/Octoshift/Services/ArchiveUploader.cs
```diff
@@ -0,0 +1,167 @@
...
+    {
+        // Use FirstOrDefault to safely handle missing Location headers
+        var locationHeader = headers.First(header => header.Key.Equals("Location", StringComparison.OrdinalIgnoreCase));
+
+        if (!string.IsNullOrEmpty(locationHeader.Key))
+        {
+            var locationValue = locationHeader.Value.FirstOrDefault();
+            if (locationValue.HasValue())
+            {
+                return new Uri(new Uri(BASE_URL), locationValue);
```
ArinGhazarian, line 162: There is a bug here that I discovered when fiddling with the unit tests. The base url is `https://uploads.github.com/organizations` but it's actually not a base url but rather a full url. The base here would be `https://uploads.github.com`. So if you call `.ToString()` on the Uri object that is getting returned from the method it will actually omit the `organizations`:

image

To fix we can compose the url like below or alternatively we can change the base url and prepend `organizations` to the `locationValue`, I guess the first is just easier:

```c#
new Uri($"{BASE_URL}/{locationValue}")
```

src/OctoshiftCLI.Tests/Octoshift/Services/GithubApiTests.cs
```diff
@@ -3475,10 +3481,10 @@ public async Task UploadArchiveToGithubStorage_Should_Upload_Stream_Content()
         var expectedUri = "gei://archive/123456";
         var jsonResponse = $"{{ \"uri\": \"{expectedUri}\" }}";
 
-        _githubClientMock
-            .Setup(m => m.PostAsync(It.IsAny<string>(), It.IsAny<StreamContent>(), null))
-            .ReturnsAsync(jsonResponse);
-
+        // Mocking the Upload method on _archiveUploader to return the expected URI
+        _archiveUploader
+            .Setup(m => m.Upload(It.IsAny<Stream>(), archiveName, orgDatabaseId))
```
ArinGhazarian, line 3486: Let's be explicit about what we are expecting to receive:

```suggestion
            .Setup(m => m.Upload(archiveContent, archiveName, orgDatabaseId))
```

src/Octoshift/Services/ArchiveUploader.cs
```diff
@@ -0,0 +1,167 @@
...
+        }
+    }
+
+    private async Task<string> UploadMultipart(Stream archiveContent, string archiveName, string uploadUrl)
+    {
+        if (archiveContent == null)
+        {
+            throw new ArgumentNullException(nameof(archiveContent), "Archive content stream cannot be null.");
+        }
+
```
ArinGhazarian, line 63: This can be removed now because we're null guarding in the public method.
```suggestion
```

ArinGhazarian (commented):
Looks so much better now, my main concerns were mostly around testing this time.

src/Octoshift/Services/ArchiveUploader.cs
ArinGhazarian: fixed in c93e44df9a493fb9c0b2e91feab03e471aefa368

src/OctoshiftCLI.Tests/Octoshift/Services/ArchiveUploadersTests.cs
```diff
@@ -0,0 +1,90 @@
...
+        // Assert
+        archiveUploaderMock.Protected().Verify<Task<string>>(
+            "UploadMultipart",
+            Times.Once(),
+            ItExpr.Is<Stream>(s => s == largeStream),
+            ItExpr.Is<string>(name => name == archiveName),
+            ItExpr.Is<string>(url => url.Contains(orgDatabaseId))
+        );
+        Assert.Equal(expectedMultipartResponse, result);
+    }
```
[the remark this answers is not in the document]
ArinGhazarian: fixed in c93e44df9a493fb9c0b2e91feab03e471aefa368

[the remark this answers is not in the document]
begonaguereca: Thanks! I also updated this test to account for a bug in the code 👍🏽

src/Octoshift/Services/ArchiveUploader.cs
begonaguereca: Fixed this to account for the bug within the 🐛 [ede2cd8](https://github.com/github/gh-gei/pull/1285/commits/ede2cd8616b8c43ccc95b730f31e7f976327054e)

src/OctoshiftCLI.Tests/Octoshift/Services/GithubApiTests.cs
```diff
@@ -3491,20 +3497,17 @@ public async Task UploadArchiveToGithubStorage_Should_Upload_Stream_Content()
...
-
-        _githubApi._streamSizeLimit = 1;
-
-        _githubClientMock
-            .Setup(m => m.PostAsync(It.IsAny<string>(), It.IsAny<MultipartFormDataContent>(), null))
-            .ReturnsAsync(jsonResponse);
+        // Mocking the Upload method on _archiveUploader to return the expected URI
+        _archiveUploader
+            .Setup(m => m.Upload(It.IsAny<Stream>(), archiveName, orgDatabaseId))
+            .ReturnsAsync(expectedUri);
```
ArinGhazarian, line 3510: As I has mentioned earlier this test can be removed.

ArinGhazarian (approved):
All good 🎉 🥇

PR #1294, opened 2024-11-07
Add --git-archive-path and --metadata-archive-path options

jfine:
Adds the ability to migrate a local file utilizing the --git-archive-path and --metadata-archive-path options. This change is similar to the --git-archive-url and --metadata-archive-url options though instead of starting the migration directly it uploads the archives (only one upload if the same archive is specified) to the storage option selected and then starts the migration.

- [x] Did you write/update appropriate tests
- [x] Release notes updated (if appropriate)
- [x] Appropriate logging output
- [ ] Issue linked
- [ ] Docs updated (or issue created)
- [ ] New package licenses are added to `ThirdPartyNotices.txt` (if applicable)

PR #1308, opened 2024-12-05
Use args.GithubTargetOrg instead of args.GithubSourceOrg

jfine:
This updates the argument passed into `UploadArchive` when migrating local archives which was introduced via https://github.com/github/gh-gei/pull/1294.

When using `UploadArchiveToAzure` or `UploadArchiveToAws` the `githubTargetOrg` isn't utilized but when using `UploadArchiveToGithub` the bug is encountered since the argument is utilized there.

- [ ] Issue linked

### DIFF
--- a/src/gei/Commands/MigrateRepo/MigrateRepoCommandHandler.cs
+++ b/src/gei/Commands/MigrateRepo/MigrateRepoCommandHandler.cs
@@ -131,7 +131,7 @@ public class MigrateRepoCommandHandler : ICommandHandler<MigrateRepoCommandArgs>
             var shouldUploadOnce = args.GitArchivePath == args.MetadataArchivePath;
 
             args.GitArchiveUrl = await UploadArchive(
-                args.GithubSourceOrg,
+                args.GithubTargetOrg,
                 args.AwsBucketName,
                 args.UseGithubStorage,
                 args.GitArchivePath,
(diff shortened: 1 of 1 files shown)


ArinGhazarian (approved):

src/gei/Commands/MigrateRepo/MigrateRepoCommandHandler.cs
```diff
@@ -131,7 +131,7 @@ public async Task Handle(MigrateRepoCommandArgs args)
             var shouldUploadOnce = args.GitArchivePath == args.MetadataArchivePath;
 
             args.GitArchiveUrl = await UploadArchive(
-                args.GithubSourceOrg,
+                args.GithubTargetOrg,
```
synthead, line 134: We should probably have a test for this :smile:

jfine: Adding an integration test for using github owned storage w/ local file would be 👍

ArinGhazarian: We don't necessarily need an INT test here this as it can be done via a simple unit test. All that matters is to make sure we're using the target org [here](https://github.com/github/gh-gei/blob/main/src/gei/Commands/MigrateRepo/MigrateRepoCommandHandler.cs#L374)

PR #1311, opened 2024-12-07
Add progress report when uploading migration archives to Azure Blob or AWS S3

ArinGhazarian:
Closes https://github.ghe.com/github/octoshift/issues/9735

This PR adds the following:

1. Adds progress reports when uploading migration archives to Azure Blob or AWS S3 to the `gh [gei|bbs2gh] migrate-repo` commands.

Sample progress report when uploading to AWS S3:

image

Sample progress report when uploading to Azure blob:

image

2. Additionally I made the `OctoLogger` thread safe by using a shared `mutex` object for all public log methods.

- [x] Issue linked


iomekam (approved):
LGTM!

src/Octoshift/Services/AzureApi.cs
```diff
@@ -89,4 +101,24 @@ private Uri GetServiceSasUriForBlob(BlobClient blobClient)
...
+        {
+            if (DateTime.Now < _nextProgressReport)
+            {
+                return;
+            }
+
+            _nextProgressReport = _nextProgressReport.AddSeconds(UPLOAD_PROGRESS_REPORT_INTERVAL_IN_SECONDS);
+        }
+
+        var percentage = (int)(uploadedBytes * 100L / totalBytes);
```
jeviet, line 117: If archives are being transferred on a lossy connection, I believe that the percentage completed could exceed 100%.

jeviet: (Depending upon how/if TCP header size is removed from uploadedBytes, that could also cause a discrepancy)

ArinGhazarian: I don't think so as the `uploadedBytes` will always be reported by the Azure library. It should take care of everything including retries, etc internally.

jeviet (commented):
Approving based upon conversation

PR #1313, opened 2024-12-10
Improve integration test console output

ArinGhazarian:
## Description

This PR adds the follows two changes:

1. By default xUnit only shows the tests outputs after the test was run. This PR adds live outputs to the console during the test run (supported in xUnit Runners v2 2.8.1+), in addition to showing them after the test has completed. This helps a lot with debugging specially debugging the migration script itself.
2. We used .NET's default `Process` class for shelling out to run the generated powershell migration script. The problem was the we hadn't redirected the `stdout` and `stderr` so we could only see the output of the `migrate-repo` commands but not the script itself. Redirecting the outputs using the `Process` class is a bit of a hassle though and not very intuitive so instead we switched to using [CliWrap](https://github.com/Tyrrrz/CliWrap) which is a better alternative.

here is a sample output that we were getting before:

image

- [x] Issue linked


synthead (approved):
LGTM!

PR #1330, opened 2025-03-28
Fixes secret scanning alerts migration - locations processing and matching

theztefan:
This PR addresses a bug reported #1331 where Secret Scanning alerts with non-commit locations (e.g., pull_request_comment, issue_title, issue_body) were:
1. Locations were no being parsed correctly and `NullReferenceException ` exceptions was thrown
2. The matching between non `commit` and `wiki_commit` locations was not correct. The issue was the strict URL comparison logic, which failed when the base URL differed between source and target repository.

Changes Made:

- Updated `BuildSecretScanningAlertLocation` method to make it more robust by safely accessing properties when building the object in `src/Octoshift/Services/GithubApi.cs`
- Enhanced `AreLocationsEqual` function in `src/Octoshift/Services/SecretScanningAlertService.cs`:
  - For `commit` and `wiki_commit` locations, the comparison remains strict, checking all relevant fields 
  - For all other location types, the comparison now focuses on the unique identifier in the URL (e.g., the comment ID or issue number) by extracting and comparing the final path segment of the URLs.
- Added tests to ensure proper matching for all non-commit location types and  that alerts with matching unique identifiers in URLs are correctly updated in the target repository.

- [x] Issue linked
- [ ] ~~Docs updated (or issue created)~~
- [ ] ~~New package licenses are added to `ThirdPartyNotices.txt` (if applicable)~~


ArinGhazarian (changes requested):

src/Octoshift/Services/SecretScanningAlertService.cs
```diff
@@ -153,6 +147,30 @@ private bool AreLocationsEqual(GithubSecretScanningAlertLocation sourceLocation,
             .GroupBy(alert => (alert.Alert.SecretType, alert.Alert.Secret))
             .ToDictionary(group => group.Key, group => group.ToList());
     }
+
+    // Compares the final segment of an URL which is relevant for the comparison
+    private bool CompareUrlIds(string sourceUrl, string targetUrl)
+    {
+        return !string.IsNullOrWhiteSpace(sourceUrl) && !string.IsNullOrWhiteSpace(targetUrl) && sourceUrl.TrimEnd('/').Split('/').Last()
+            == targetUrl.TrimEnd('/').Split('/').Last();
```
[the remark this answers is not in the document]
theztefan: @ArinGhazarian the urls we are comparing here are coming from Secret Scanning Locations REST API(https://docs.github.com/en/rest/secret-scanning/secret-scanning?apiVersion=2022-11-28#list-locations-for-a-secret-scanning-alert). An alert with any type of location will have a uri to the location where a secret was matched (irrelevant of location type). Therefore, I though I think your suggestion is legit, in this cases is not necessary. 

I've switched to using `HasValue()`

src/Octoshift/Services/GithubApi.cs
```diff
@@ -1188,23 +1188,23 @@ private static GithubSecretScanningAlertLocation BuildSecretScanningAlertLocatio
...
+            IssueBodyUrl = alertLocation["details"]["issue_body_url"]?.ToString(),
+            IssueCommentUrl = alertLocation["details"]["issue_comment_url"]?.ToString(),
+            DiscussionTitleUrl = alertLocation["details"]["discussion_title_url"]?.ToString(),
+            DiscussionBodyUrl = alertLocation["details"]["discussion_body_url"]?.ToString(),
+            DiscussionCommentUrl = alertLocation["details"]["discussion_comment_url"]?.ToString(),
+            PullRequestTitleUrl = alertLocation["details"]["pull_request_title_url"]?.ToString(),
+            PullRequestBodyUrl = alertLocation["details"]["pull_request_body_url"]?.ToString(),
+            PullRequestCommentUrl = alertLocation["details"]["pull_request_comment_url"]?.ToString(),
+            PullRequestReviewUrl = alertLocation["details"]["pull_request_review_url"]?.ToString(),
+            PullRequestReviewCommentUrl = alertLocation["details"]["pull_request_review_comment_url"]?.ToString(),
```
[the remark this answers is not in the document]
theztefan: The original intention of this change was to handle the exception described in #1331 where, depending on the location type, properties might be missing in the API's JSON response. Basically, I tried to handle nulls gracefully. 

There was no real reason why I switched to `ToString()` other than my limited knowledge and reading that ToString() should not throw an exception 😅. Having read now that direct casting is more performant I have reverted back to direct cast while keeping the null safety checks and assigning default values for the properties that may vary in API responses.

src/Octoshift/Services/GithubApi.cs
```diff
@@ -1184,29 +1184,31 @@ private static GithubSecretScanningAlert BuildSecretScanningAlert(JToken secretA
...
-            IssueCommentUrl = (string)alertLocation["details"]["issue_comment_url"],
-            DiscussionTitleUrl = (string)alertLocation["details"]["discussion_title_url"],
-            DiscussionBodyUrl = (string)alertLocation["details"]["discussion_body_url"],
-            DiscussionCommentUrl = (string)alertLocation["details"]["discussion_comment_url"],
-            PullRequestTitleUrl = (string)alertLocation["details"]["pull_request_title_url"],
-            PullRequestBodyUrl = (string)alertLocation["details"]["pull_request_body_url"],
-            PullRequestCommentUrl = (string)alertLocation["details"]["pull_request_comment_url"],
-            PullRequestReviewUrl = (string)alertLocation["details"]["pull_request_review_url"],
-            PullRequestReviewCommentUrl = (string)alertLocation["details"]["pull_request_review_comment_url"],
+            Path = (string)details?["path"],
```
ArinGhazarian, line 1193: I'm not sure if we want to use a null-conditional here. To me it's an exceptional case and probably should not happen (we don't expect it to happen). Is there a scenario that the `details` would be `null` ?

test.output
ArinGhazarian, line 1: This was probably an accidental push!

ArinGhazarian (changes requested):
Looks good, the only thing is `test.output` should be removed.

test.output
theztefan: 🤦

ArinGhazarian (approved):
🎉

ArinGhazarian:
INT tests are all 🟢 merging now.

Issue #1365, opened 2025-06-20
Support GHES and GHEC Data Residency with `ado2gh rewire-pipeline` command


markphip:
After migrating a repository to GHEC Data Residency, you need to also rewire the pipelines. When you run the command it "sort of" works except it hard codes github.com URL's for the repositories when it updates the Azure DevOps build definitions.

One solution would be to support the --target-api-url option that migrate-repo supports. Even though a GitHub API is technically not used by this command, it could use that URL to determine the URL to use for the repository. The payload it sends to Azure DevOps does include the API URL so it is needed.

As example, if the API URL is `https://api.instance.ghe.com` then this value would be used directly where the API URL is needed and it could be parsed to remove `api.` when determining the URL to use for a repository URL

dylan-smith:
For anybody looking at this in the future this is where we have this hardcoded that will need to change: https://github.com/github/gh-gei/blob/main/src/Octoshift/Services/AdoApi.cs#L537

dylan-smith:
Solved with https://github.com/github/gh-gei/pull/1366

PR #1366, opened 2025-06-20
Add support for `--target-api-url` to `rewire-pipeline` command 

markphip:
Add support for `--target-api-url` to the `rewire-pipeline` command so that the command can be used when migrating repositories to GHEC with Data Residency

- [x] Issue linked

Fixes Issue #1365

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,27 @@
+This is a C# based repository that produces several CLIs that are used by customers to interact with the GitHub migration APIs. Please follow these guidelines when contributing:
+
+## Code Standards
+
+### Required Before Each Commit
+- Run `dotnet format src/OctoshiftCLI.sln` before committing any changes to ensure proper code formatting. This will run dotnet format on all C# files to maintain consistent style
+
+### Development Flow
+- Build: `dotnet build src/OctoshiftCLI.sln /p:TreatWarningsAsErrors=true`
+- Test: `dotnet test src/OctoshiftCLI.Tests/OctoshiftCLI.Tests.csproj`
+
+## Repository Structure
+- `src/`: Contains the main C# source code for the Octoshift CLI
+- `src/ado2gh/`: Contains the ADO to GH CLI commands
+- `src/bbs2gh/`: Contains the BBS to GH CLI commands
+- `src/gei/`: Contains the GitHub to GitHub CLI commands
+- `src/Octoshift/`: Contains shared logic used by multiple commands/CLIs
+- `src/OctoshiftCLI.IntegrationTests/`: Contains integration tests for the Octoshift CLI
+- `src/OctoshiftCLI.Tests/`: Contains unit tests for the Octoshift CLI
+
+## Key Guidelines
+1. Follow C# best practices and idiomatic patterns
+2. Maintain existing code structure and organization
+4. Write unit tests for new functionality.
+5. When making changes that would impact our users (e.g. new features or bug fixes), add a bullet point to `RELEASENOTES.md` with a user friendly brief description of the change
+6. Never silently swallow exceptions.
+7. If an exception is expected/understood and we can give a helpful user-friendly message, then throw an OctoshiftCliException with a user-friendly message. Otherwise let the exception bubble up and the top-level exception handler will log and handle it appropriately.
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -6,6 +6,14 @@ on:
       pr_number:
         type: number
         required: true
+      sha:
+        type: string
+        required: true
+
+permissions:
+  contents: read
+  checks: write
+  pull-requests: write
 
 jobs:
   build-for-e2e-test:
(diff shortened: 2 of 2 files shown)


markphip:
The code for this was written for me by Copilot Agent mode in VS but I was able to test and verify that it worked as exptected.

src/Octoshift/Services/AdoApi.cs
```diff
@@ -523,34 +523,59 @@ public virtual async Task ShareServiceConnection(string adoOrg, string adoTeamPr
...
+                manageUrl,
                 orgName = githubOrg,
-                refsUrl = $"https://api.github.com/repos/{githubOrg.EscapeDataString()}/{githubRepo.EscapeDataString()}/git/refs",
+                refsUrl,
                 safeRepository = $"{githubOrg.EscapeDataString()}/{githubRepo.EscapeDataString()}",
                 shortName = githubRepo,
                 reportBuildStatus = true
             },
             id = $"{githubOrg}/{githubRepo}",
             type = "GitHub",
```
markphip, line 576: Even though this worked for me, I am not 100% certain this does not need to be a different value if the repository is on GHES or GHEC with Data Residency

src/Octoshift/Services/AdoApi.cs
```diff
@@ -523,34 +523,59 @@ public virtual async Task ShareServiceConnection(string adoOrg, string adoTeamPr
...
+    public virtual async Task ChangePipelineRepo(string adoOrg, string teamProject, int pipelineId, string defaultBranch, string clean, string checkoutSubmodules, string githubOrg, string githubRepo, string connectedServiceId, string targetApiUrl = null)
     {
         var url = $"{_adoBaseUrl}/{adoOrg.EscapeDataString()}/{teamProject.EscapeDataString()}/_apis/build/definitions/{pipelineId}?api-version=6.0";
 
         var response = await _client.GetAsync(url);
         var data = JObject.Parse(response);
 
+        // Determine base URLs
+        string apiUrl, webUrl, cloneUrl, branchesUrl, refsUrl, manageUrl;
+        if (!string.IsNullOrWhiteSpace(targetApiUrl))
```
dylan-smith: Use our extension method HasValue() to do this check
```suggestion
        if (targetApiUrl.HasValue())
```

dylan-smith:
There'll need to be some unit tests for the changes in `AdoApi`

src/OctoshiftCLI.Tests/ado2gh/Commands/RewirePipeline/RewirePipelineCommandTests.cs
```diff
@@ -31,7 +31,7 @@ public void Should_Have_Options()
         {
             Assert.NotNull(_command);
             Assert.Equal("rewire-pipeline", _command.Name);
-            Assert.Equal(8, _command.Options.Count);
+            Assert.Equal(9, _command.Options.Count); // Updated from 8 to 9
```
offbyone:
```suggestion
            Assert.Equal(9, _command.Options.Count); 
```

don't put a comment about a change in the code like this

src/OctoshiftCLI.Tests/ado2gh/Commands/RewirePipeline/RewirePipelineCommandTests.cs
```diff
@@ -41,6 +41,7 @@ public void Should_Have_Options()
             TestHelpers.VerifyCommandOption(_command.Options, "service-connection-id", true);
             TestHelpers.VerifyCommandOption(_command.Options, "ado-pat", false);
             TestHelpers.VerifyCommandOption(_command.Options, "verbose", false);
+            TestHelpers.VerifyCommandOption(_command.Options, "target-api-url", false); // Add check for new option
```
offbyone:
```suggestion
            TestHelpers.VerifyCommandOption(_command.Options, "target-api-url", false);
```

Same

src/ado2gh/Commands/RewirePipeline/RewirePipelineCommandArgs.cs
```diff
@@ -12,5 +12,6 @@ public class RewirePipelineCommandArgs : CommandArgs
         public string ServiceConnectionId { get; set; }
         [Secret]
         public string AdoPat { get; set; }
+        public string TargetApiUrl { get; set; } // Added for GHEC Data Residency support
```
offbyone:
```suggestion
        public string TargetApiUrl { get; set; } 
```

This as well

offbyone (commented):
Please remove the comments that are basically inline commit messages. Otherwise LGTM

src/Octoshift/Services/AdoApi.cs
iomekam: I'm not even sure that this is doing anything. From what I'm looking at, [these](https://learn.microsoft.com/en-us/rest/api/azure/devops/build/definitions/update?view=azure-devops-rest-6.0#definitiontype) are the potential value for "type" and I don't see `GitHub` as an option

iomekam (commented):
Changes look good to me. My only thought would be to maybe check for whether the target uri is well formed. If we don't do that elsewhere in the codebase, maybe it's not worth the effort here

src/Octoshift/Services/AdoApi.cs
dylan-smith: The types in the docs you linked is the type of _build definition_. The type being used here is the type of _repository_ (https://learn.microsoft.com/en-us/rest/api/azure/devops/build/definitions/update?view=azure-devops-rest-6.0#buildrepository)

markphip: Yeah my assumption, since it all worked correctly, was that this was just the type of repository but I wanted to call it out. Talking with the ADO team working on this feature they seemed to affirm that this was doing the right things

markphip:
> Changes look good to me. My only thought would be to maybe check for whether the target uri is well formed. If we don't do that elsewhere in the codebase, maybe it's not worth the effort here

Let me know if anything more is needed of me. If this is merged, does release happen automatically?

iomekam (approved):
LGTM

PR #1462, opened 2025-12-02
[Copy of 1460] Add path validation and logging for bbs2gh archive operations 

### LINKED ISSUE #1211: Validate `--bbs-shared-home` and `--archive-path`
When running `bbs2gh` **on the Bitbucket instance** we need to make sure that if the `--bbs-shared-home` is provided it points to an existing directory. Additionally if the `--archive-path` is provided (regardless of where the CLI is running) we need to make sure it points to an existing file.

Currently we don't check those paths and if they are wrong the upload phase errors out but the error message is not user friendly (take this [support ticket](https://github.com/github/migration-friction/issues/1798) for example) and can cause some confusion. With this change we can clearly say what is wrong and fail fast.

## Todo
- [ ] Validate `--bbs-shared-home` if it's provided and the CLI is running on Bitbucket instance
- [ ] Validate `--bbs-shared-home` if it's provided regardless of where the CLI is running

brianaj:
To ship changes in https://github.com/github/gh-gei/pull/1460
Fixes #1211 

- [x] Issue linked


jfine (approved):
I didn't directly excersise the code however it looks ✨.

PR #1480, opened 2025-12-08
Potential fix for code scanning alert no. 1035: Environment variable built from user-controlled sources

ismith:
Potential fix for [https://github.com/github/gh-gei/security/code-scanning/1035](https://github.com/github/gh-gei/security/code-scanning/1035)

To address this vulnerability, we should strictly control the value of `PR_NUMBER` before it is written to `$GITHUB_ENV`. This means ensuring that:

- Only valid, single-line numeric data is used (since PR numbers are always positive integers).
- All potential for newlines, extra characters, or injection attacks is removed.

This was already being done - `tr -cd '0-9'` removes (`-d`) the complement (`-c`) of the specified characters. But CodeQL doesn't recognize that, so we'll use its suggested approach (`grep -E '^[0-9]+$'`) instead.

 ---

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/.github/workflows/publish-test-results.yml
+++ b/.github/workflows/publish-test-results.yml
@@ -36,12 +36,8 @@ jobs:
       - name: Extract PR Number
         run: |
           prnumber=$(jq -r '.pull_request.number' < 'artifacts/Event File/event.json')
-          sanitized_prnumber=$(echo "$prnumber" | grep -E '^[0-9]+$' || echo "")
-          if [ -n "$sanitized_prnumber" ]; then
-            echo "PR_NUMBER=$sanitized_prnumber" >> "$GITHUB_ENV"
-          else
-            echo "PR_NUMBER=" >> "$GITHUB_ENV"
-          fi
+          sanitized_prnumber=$(grep -E '^[0-9]+$' <<< "$prnumber")
+          echo "PR_NUMBER=$sanitized_prnumber" >> "$GITHUB_ENV"
       - name: Publish Unit Test Results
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-test-results.yml
```diff
@@ -35,10 +35,13 @@ jobs:
...
-          prnumber=`echo $(jq -r '.pull_request.number' <<< "$eventjson")`
-          echo "PR_NUMBER=$(echo $prnumber | tr -cd '0-9')" >> $GITHUB_ENV
-      
+          prnumber=$(jq -r '.pull_request.number' < 'artifacts/Event File/event.json')
+          sanitized_prnumber=$(echo "$prnumber" | grep -E '^[0-9]+$' || echo "")
+          if [ -n "$sanitized_prnumber" ]; then
+            echo "PR_NUMBER=$sanitized_prnumber" >> "$GITHUB_ENV"
+          else
+            echo "PR_NUMBER=" >> "$GITHUB_ENV"
+          fi
```
synthead: How about:

```suggestion
          sanitized_prnumber=$(grep -E '^[0-9]+$' <<< "$prnumber")
          echo "PR_NUMBER=$sanitized_prnumber" >> "$GITHUB_ENV"
```

This works because:

- `<<<` is the same as `echo ... |`, but without `echo` and a pipe.
- If `grep` does not match `^[0-9]+$`, then it will return `""` to stdout anyway.
- We don't need to check if `$sanitized_prnumber` is `""` to explicitly write `""`.

PR #1465, opened 2025-12-03
Harder pipeline rewire monitoring and error handling 

georgy-gorelko:
- Fixed branch policy check for pipelines with prefixed repository names to use repository ID directly when available, preventing 404 errors during rewiring
- Skip branch policy checks for disabled repositories with a warning instead of attempting API calls that may fail
- Skip pipeline rewiring entirely for disabled repositories, exiting early with an appropriate warning message
- Fixed misleading success message that appeared even when pipeline rewiring was skipped for disabled repositories
- Fixed monitor timeout minutes to only display when --dry-run mode is enabled, reducing confusion during regular pipeline rewiring operations
- Check if pipeline is disabled before attempting to queue a test build, preventing 400 Bad Request errors and providing clear warning messages

PR #1493, opened 2025-12-22
Improve rewire-pipeline error handling for disabled repos/pipelines

brianaj:
This PR merges external PR #1465 from @georgy-gorelko which improves error handling and monitoring for the `rewire-pipeline` command.

## Summary
Comprehensive improvements to ADO pipeline rewiring with better error handling, especially around disabled repositories and branch policy checks. Fixes issues introduced in v1.19.0.

### 1. Fixed Branch Policy Check with Prefixed Repo Names
- Now uses repository ID directly when available instead of just name
- Prevents 404 errors during rewiring for classic pipelines with prefixed names
- Smart fallback: uses ID if available, otherwise looks up by name

### 2. Disabled Repository Handling
- Checks if repository is disabled before attempting operations
- Skips pipeline rewiring entirely for disabled repos with clear warnings
- Prevents unnecessary API calls that would fail
- Returns proper status (true/false) to indicate success/skip

### 3. Pipeline Disabled Check
- New `IsPipelineEnabled()` method checks `queueStatus` field
- Prevents 400 Bad Request errors when trying to queue builds on disabled pipelines
- Clear warning messages when pipeline is disabled

### 4. Fixed Misleading Success Messages
- Changed `RewirePipelineToGitHub` to return `bool` instead of `void`
- Only logs success when rewiring actually happened (not when skipped)
- Caller checks return value before showing success message

### 5. Monitor Timeout Display
- Only shows timeout minutes in dry-run mode (where it's relevant)
- Updated description: "(Dry-run mode only)"
- Custom `Log()` override in `RewirePipelineCommandArgs` excludes `MonitorTimeoutMinutes`

### 6. Caching Improvements
- Enhanced cache to store both ID and disabled status
- 404 responses treated as disabled repos and cached to avoid repeated calls
- Reduces redundant API calls

## Test Coverage
Excellent test coverage with 11 new tests:
- `IsPipelineEnabled` tests (4 tests)
- Branch policy tests (3 tests)
- Rewiring tests (3 tests)
- Pipeline test service (1 test)

## Testing
- All 1029 tests passing
- Comprehensive coverage of disabled repo/pipeline scenarios

Closes #1465

cc: @georgy-gorelko

### DIFF
--- a/src/Octoshift/Services/AdoPipelineTriggerService.cs
+++ b/src/Octoshift/Services/AdoPipelineTriggerService.cs
@@ -22,7 +22,7 @@ public class AdoPipelineTriggerService
     private readonly string _adoBaseUrl;
 
     // Cache for repository IDs and branch policies to avoid redundant API calls
-    private readonly Dictionary<string, (string id, bool isDisabled)> _repositoryCache = [];
+    private readonly Dictionary<string, AdoRepository> _repositoryCache = [];
     private readonly Dictionary<string, AdoBranchPolicyResponse> _branchPolicyCache = [];
 
     public AdoPipelineTriggerService(AdoApi adoApi, OctoLogger log, string adoBaseUrl)
@@ -103,21 +103,21 @@ public class AdoPipelineTriggerService
 
         try
         {
-            var (repositoryId, isRepositoryDisabled) = await GetRepositoryIdAndStatus(adoOrg, teamProject, repoName, repoId, pipelineId);
+            var repoInfo = await GetRepositoryIdAndStatus(adoOrg, teamProject, repoName, repoId, pipelineId);
 
-            if (string.IsNullOrEmpty(repositoryId))
+            if (string.IsNullOrEmpty(repoInfo.Id))
             {
                 return false;
             }
 
-            if (isRepositoryDisabled)
+            if (repoInfo.IsDisabled)
             {
                 var repoIdentifier = repoName ?? repoId;
                 _log.LogInformation($"Repository {adoOrg}/{teamProject}/{repoIdentifier} is disabled. Branch policy check skipped for pipeline {pipelineId} - will use default trigger configuration.");
                 return false;
             }
 
-            return await CheckBranchPoliciesForPipeline(adoOrg, teamProject, repositoryId, repoName, repoId, pipelineId);
+            return await CheckBranchPoliciesForPipeline(adoOrg, teamProject, repoInfo.Id, repoName, repoId, pipelineId);
         }
         catch (Exception ex) when (ex is HttpRequestException or TaskCanceledException or JsonException or ArgumentException or InvalidOperationException)
         {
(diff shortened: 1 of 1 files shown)


gohana:
e2e tests are passing after updating our trial license for https://e2e-bbs-linux-1.westus2.cloudapp.azure.com/!

RELEASENOTES.md
```diff
@@ -1,2 +1,5 @@
-- Added validation to detect and return clear error messages when a URL is provided instead of a name for organization, repository, or enterprise arguments (e.g., `--github-org`, `--github-target-org`, `--source-repo`, `--github-target-enterprise`)
-- Added `--target-api-url` as an optional arg to the `add-team-to-repo` command
+- **ado2gh**: Improved `rewire-pipeline` command to gracefully handle disabled repositories and pipelines with clear warnings instead of errors
+- **ado2gh**: Fixed 404 errors when checking branch policies for classic pipelines with prefixed repository names
+- **ado2gh**: Fixed 400 errors when running `rewire-pipeline --dry-run` on disabled pipelines
```
brianaj: This can be removed 
```suggestion
```

RELEASENOTES.md
```diff
@@ -1,2 +1,5 @@
-- Added validation to detect and return clear error messages when a URL is provided instead of a name for organization, repository, or enterprise arguments (e.g., `--github-org`, `--github-target-org`, `--source-repo`, `--github-target-enterprise`)
-- Added `--target-api-url` as an optional arg to the `add-team-to-repo` command
+- **ado2gh**: Improved `rewire-pipeline` command to gracefully handle disabled repositories and pipelines with clear warnings instead of errors
```
brianaj:
```suggestion
```

RELEASENOTES.md
```diff
@@ -1,2 +1,5 @@
-- Added validation to detect and return clear error messages when a URL is provided instead of a name for organization, repository, or enterprise arguments (e.g., `--github-org`, `--github-target-org`, `--source-repo`, `--github-target-enterprise`)
```
brianaj, line 1: we should keep this

RELEASENOTES.md
```diff
@@ -1,2 +1,5 @@
...
-- Added `--target-api-url` as an optional arg to the `add-team-to-repo` command
```
brianaj, line 2: we should keep this

RELEASENOTES.md
gohana: Done [23bbedc](https://github.com/github/gh-gei/pull/1493/commits/23bbedc933f52b0738825bc73266b8270f4a7a01)

RELEASENOTES.md
gohana: Done [23bbedc](https://github.com/github/gh-gei/pull/1493/commits/23bbedc933f52b0738825bc73266b8270f4a7a01) thanks!

src/ado2gh/Commands/RewirePipeline/RewirePipelineCommandArgs.cs
```diff
@@ -16,5 +19,43 @@ public class RewirePipelineCommandArgs : CommandArgs
...
         public int MonitorTimeoutMinutes { get; set; } = 30;
+
+        public override void Log(OctoLogger log)
+        {
+            if (log is null)
+            {
+                throw new ArgumentNullException(nameof(log));
+            }
+
+            log.Verbose = Verbose;
```
offbyone, line 30: This looks like it mutates the log argument and doesn't set it back to the original verbose state. Is that intentional?

src/Octoshift/Services/AdoPipelineTriggerService.cs
```diff
@@ -22,7 +22,7 @@ public class AdoPipelineTriggerService
     private readonly string _adoBaseUrl;
 
     // Cache for repository IDs and branch policies to avoid redundant API calls
-    private readonly Dictionary<string, string> _repositoryIdCache = [];
+    private readonly Dictionary<string, (string id, bool isDisabled)> _repositoryCache = [];
```
offbyone: Please define a data object here, so that the boolean is clearly defined when used. I don't know CSharp, but I know in Python that would look like this:

```python
@dataclass
class RepositoryMetadata:
    id: str
    isDisabled: bool
    # maybe other fields later?
```

The benefit there is that later, when you are reading out of the cache, instead of code like `if (_repositoryCache[cacheKey][1]) { ... }` you have `if(_repositoryCache[cacheKey].isDisabled) { ... }`.

That's especially valuable here, because the cached value is a boolean of a "negative" state (`isDisabled` as opposed to a more conventional `enabled` flag, which is a more natural thing to expect when seeing a bare boolean)

src/ado2gh/Commands/RewirePipeline/RewirePipelineCommandArgs.cs
gohana: This appears to be intentionally mirroring the base class behavior: 

https://github.com/github/gh-gei/blob/ce47d4b618688bac1eafb37a456c106c22229fda/src/Octoshift/Commands/CommandArgs.cs#L23

This state is initialized based on the `--verbose` flag, so there isn't really a meaningful state to restore to.

PR #1529, opened 2026-03-19
fix: exclude errored code scanning analyses from migration

lindluni:
Filter out code scanning analyses with non-empty error fields when listing analyses from the source repository, preventing failed analyses from being migrated to the target repository.

When the `migrate-code-scanning-alerts` is invoked, it used the `ListAnalyses` API. The results objects in the returned array have an `error` field. If on the initial upload the SARIF failed to process, the error field is populated. When we try to `GetAnalysis` to retrieve the SARIF content, we get a `422 Unprocessable Entity` because the results retrieval system can't process the SARIF. The GEI CLI fails at that point, instead of skipping that particular result.

We've also built and tested this locally against a production migration and confirmed this works as intended.

- [ ] Issue linked


scottluskcis (approved):

offbyone:
I don't love that the error looks like it's silently dropped. Is this error message reported anywhere?

lindluni:
I updated the implementation. I was a bit overzealous on the logging. I'm trying to make it clear this was an error processing the original upload and not with the process GEI is undertaking. Trying to avoid users opening issues with support and migration-friction thinking they had errors in the tool.

jfine (approved):

